PR #3, opened 2021-02-04
Registration, Login, and Logout

lonnieezell:


MGatner (commented):
Holy cow, you're crushing it. What's `Sparks`?

lonnieezell:
Thanks! 

I originally put it in the CodeIgniter space, but that caused issues. Figuring this will live in codeigniter-projects or addons, but both of those were a longer namespace than I wanted. Since CI2 had the sparks addons for a while, I thought using that namespace might work. Up for discussion, but I needed to change it to something so I could keep moving. :)

PR #4, opened 2021-02-06
Multi authenticator flow

lonnieezell:


docs/1 - authentication.md
```diff
@@ -206,3 +236,18 @@ if ($user->tokenCant('forums:manage'))
 }
 ```
 
+## Pipeline Handler
+
+The Pipeline handler is a unique case. It does not have any of its own logic. Instead, it runs the 
+desired method called through each of the other specified handlers, until it either gets a successful 
+result or they all fail. When it fails it only returns the result from the last handler, so you may want
```
MGatner, line 243: Is there a way this could be abstracted even more such that pipeline could eventually handle dual-authentication? i.e. right now the handlers are `OR` but with 2FA they would be `AND`.

docs/1 - authentication.md
```diff
@@ -206,3 +236,18 @@ if ($user->tokenCant('forums:manage'))
...
+to re-order the handlers within the the configuration file to provide the error that is most helpful
+for your application.
+
+This can be useful when you want to share a set of API calls between your SPA, which uses the Session
+handler, and a mobile application, which might use Access Tokens. By using the pipeline on these calls
+you can ensure the user is logged in no matter where the call came from. 
+
+It is not recommended to set this handler as the default authenticator since, though it can be done
+as long as one is very careful about methods like the login method, as that would only log the user
+into the first handler, and not both as you might expect.
```
MGatner, line 253: This seems confusing but I will reserve judgment until I have had a chance to look at the actual code. :)

src/Auth.php
```diff
@@ -50,6 +50,27 @@ public function withHandler(string $handler = null)
 		return $this;
```
MGatner, line 50: I know this was a previous commit, but using a "with____" method to me implies an immutable class. I would recommend either returning a cloned version or using `setHandler` instead.

MGatner (commented):
Well kids are interrupting my review but here are a few comments. I still haven't had a chance to be through the whole library, I will get there!

docs/1 - authentication.md
lonnieezell: That's a good question. That is coming up on my list very shortly to figure out how to handle 2FA, activators, etc. I'll think on that.

docs/1 - authentication.md
lonnieezell: well - logging in with multiple providers is just plain weird. :) It's more to allow access and authentication by a one of several, but now that I'm thinking about it, maybe just making this part of the filter makes more sense...

src/Auth.php
lonnieezell: Fair point, and since I don't see the point in immutable classes for many things, I'm happy to change the method name.

MGatner: In this case I think a setter works great. I've been warming to immutability. I think in general PHP just isn't a language where it will make too much difference, but I've been poking around at Swoole and the concurrency gains are pretty impressive. It also seems to be a growing trend - whether it ends up being a fad or "here to stay" remains to be seen but I've been incorporating more immutable classes in my work when it makes sense in the name of collaboration.

lonnieezell: I know it’s a trend and I think in many of the cases I’ve seen it used there was no gain - and a number of extra copies of that data in memory. But that could be the old grumpy man talking lol

I hadn’t heard of Swoole before but I’m definitely intrigued and will be checking it out more soon. Could prove very useful for a personal project I’m in the beginning stages of....

MGatner: Oh yeah definitely look at Swoole! I think the whole thing typically with immutability in PHP is that our program execution time, start to finish, is often mere milliseconds. Swoole and other such extensions introduce long-running processes, so all of a sudden things that didn't matter before can make or break an app. That said, doing things "the right way" can also really pay off. At the end of the day my rubric is still something like "does this class *make sense conceptually* as an immutable object"... if it is going to make my code easier to grasp then that's great, if it doesn't matter then I'll try immutability for the practice at this point, if it makes things super convoluted then screw that :P

PR #5, opened 2021-02-08
Replaced PipelineHandler with ChainFilter.

lonnieezell:
The PipelineHandler was way too heavy for the real intention - which was simply to check them both for login. So I removed the PipelineHandler altogether, and replaced it with a new filter that just checks on the login itself. This required a small refactor of how the AccessTokens handler works, but mostly because I didn't handle the statelessness correctly previously. 

I am aware that a single RegisterTest test is failing. Not sure why yet, but wanted to get this all in place before hitting the sack tonight.


docs/1 - authentication.md
```diff
@@ -236,18 +236,26 @@ if ($user->tokenCant('forums:manage'))
...
+Shield provides 3 [Controller Filters](https://codeigniter.com/user_guide/incoming/filters.html) you can 
+use to protect your routes, `session`, `tokens`, and `chained`. The first two cover the two handlers, 
+`Session` and `AccessTokens`, respectively. The `chained` filter will check both handlers in sequence 
+to see if the user is logged in through either of authenticators, allowing a single API endpoint to 
+work for both an SPA using session auth, and a mobile app using access tokens.
 
-This can be useful when you want to share a set of API calls between your SPA, which uses the Session
-handler, and a mobile application, which might use Access Tokens. By using the pipeline on these calls
-you can ensure the user is logged in no matter where the call came from. 
+Before you can use the filters you must define their aliases in `app/Config/Validation.php`:
```
MGatner, line 247: Filters.php

src/Auth.php
```diff
@@ -45,7 +45,10 @@ public function __construct(Authentication $authenticate)
 	 */
 	public function setHandler(string $handler = null)
 	{
-		$this->handler = $handler;
+		if (! empty($handler))
```
MGatner, line 48: What happens if `null` is passed?

MGatner (commented):
I think this makes a lot more sense. A couple of notes.

PR #6, opened 2021-02-09
WIP: Authentication Actions

lonnieezell:
Authentication Actions are a way to standardize email activation, email/SMS-based 2FA, etc.

### DIFF
--- a/src/Authentication/Actions/ActionInterface.php
+++ b/src/Authentication/Actions/ActionInterface.php
@@ -1,0 +1,45 @@
+<?php namespace Sparks\Shield\Authentication\Actions;
+
+use CodeIgniter\HTTP\IncomingRequest;
+
+/**
+ * Interface ActionInterface
+ *
+ * Authentication Actions are steps that can happen after
+ * the main authentication steps, like registration and login.
+ * They can be email activation steps, SMS-based 2FA, etc.
+ *
+ * @package Sparks\Shield\Authentication\Actions
+ */
+interface ActionInterface
+{
+	/**
+	 * Shows the initial screen to the user to start the flow.
+	 * This might be asking for the user's email to reset a password,
+	 * or asking for a cell-number for a 2FA.
+	 *
+	 * @return mixed
+	 */
+	public function show();
+
+	/**
+	 * Processes the form that was displayed in the previous form.
+	 *
+	 * @param \CodeIgniter\HTTP\IncomingRequest $request
+	 *
+	 * @return mixed
+	 */
+	public function handle(IncomingRequest $request);
+
+	/**
+	 * This handles the response after the user takes action
+	 * in response to the show/handle flow. This might be
+	 * from clicking the 'confirm my email' action or
+	 * following entering a code sent in an SMS.
+	 *
+	 * @param \CodeIgniter\HTTP\IncomingRequest $request
+	 *
+	 * @return mixed
+	 */
+	public function verify(IncomingRequest $request);
+}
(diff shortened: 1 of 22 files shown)


src/Controllers/ActionController.php
```diff
@@ -0,0 +1,81 @@
...
+	 */
+	protected $action;
+
+	/**
+	 * Perform an initial check if we have a valid action or not.
+	 *
+	 * @param $method
+	 * @param mixed  ...$params
+	 */
+	public function _remap($method, ...$params)
```
MGatner, line 29: This feels like an odd use of _remap(). I assume you did this because it needs to come after filters so the constructor was not an option. Maybe extending initController? Or maybe moving all this into a filter itself? Actions are sort of a grey area between middleware and controller, just brainstorming.

src/Authentication/Actions/ActionInterface.php
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Interface ActionInterface
+ *
+ * Authentication Actions are steps that can happen after
+ * the main authentication steps, like registration and login.
+ * They can be email activation steps, SMS-based 2FA, etc.
+ *
+ * @package Sparks\Shield\Authentication\Actions
+ */
+interface ActionInterface
```
MGatner, line 14: I know this is still WIP so maybe you're figuring this out still, but the looseness of this interface makes me a little uneasy. It looks like we could safely return a `ResponseInterface` from these methods, which makes me wonder if this should be an extension of `RequestHandlerInterface` and part of a legit PSR-15 middleware (https://www.php-fig.org/psr/psr-15/). 

That might be putting the cart before the horse though, but with the framework moving toward more PSR compliance in the HTTP layer this could be a good chance to model interactions with it.

lonnieezell: PSR-15 is close, but only designed to handle a single response and would make this more complex I think, since my current thinking is a single Action class provides 3 responses: 

1. Display the initial notification/form
2. Handle the form response/send the email/whatever, and display another view to tell the user it's been sent
3. Handle the final action that is probably triggered by entering the code from step 2, or clicking the link in the email, or whatever. 

And middleware definitely won't work in CI since we have to support both route-based and legacy auto-routing, which is why Controller Filters exist instead. :)

src/Controllers/ActionController.php
lonnieezell: It is  a little odd, sure. You're correct about the constructor, and I have always preferred to think of `initController` as a system function and pretend it doesn't exist, but I guess it could be used here.

src/Authentication/Actions/ActionInterface.php
lonnieezell: Likely I will end up requiring a Response as the return value, I just want to make sure it all actually works the way I think it will :)

src/Controllers/ActionController.php
MGatner: I usually use `initController1 for this kind fo stuff (assuming `__construct()` doesn't work) but it's awkward to have to re-pass the parameters for what is essentially "things you would usually do in the constructor". This feels like a re-appropriation of the remap method but maybe that's better than the other options out there.

MGatner:
See this on your test failure: https://github.com/lonnieezell/myth-auth/blob/e838cb8de6ffa118caf2b9909e71776a866c8973/tests/controllers/RegisterTest.php#L19-L23

MGatner:
🥳

PR #9, opened 2021-09-04
Static Analysis

MGatner:
This PR updates PHPStan an applies it to the existing codebase, making as few actual content changes as possible. It includes a new GitHub workflow for including static analysis in continuous integration.


MGatner:
Note that I have only gone trough **src/** so far so there will be failures. I've got a good start on **tests/** but it won't be done tonight - I wanted you to have this though inc are you work on Shield later tonight.

MGatner:
@lonnieezell Ready for review!

The only thing to note (I marked these as `@todo` in the code) is that there are some discrepancies between classes and interfaces, especially `Authenticatable` and `User`; some classes rely on methods in `User` that are not represented in the interface, which means they either need to be restricted to `User` or the interface needs to be expanded. Same thing happens with `AccessTokens` because it relies on some of the trait methods form `HasAccessTokens` that don't have interface methods. I see two options...

1. Lock everything into `User` and use that typehint everywhere, then developers can extend `User` if they want to user their own classes with the methods.
2. Add more interfaces that extend `Authenticatable` yet specify the required methods, so individual handlers can specify which interfaces they need yet there will still be a fair amount of overlap.

I lean towards 2 for its sound programming practices, but it does proliferate the classes a bit more so I think 1 has some traction.

lonnieezell (approved):

lonnieezell:
I think 2 is the better option, also, just in case someone wants to go crazy and create their own User type class that doesn't extend ours for some reason.

PR #17, opened 2021-11-12
Update Toolkit

MGatner:
* Fixes some issues with the integration workflows
* Updates PHPStan to `^1.0` and applies analysis fixes

PR #18, opened 2021-11-12
Settings Dependency

MGatner:
Switch to released version of Settings dependency.


MGatner:
The static analysis failures will be fixed after #17. The PHPUnit failures are unrelated to this change and must be due to some upstream changes (CodeIgniter framework probably?).

PR #21, opened 2021-12-12
vuepress static site generater added

cijagani:
to generate site follow this steps.

1.  go to project root directory
2. yarn install (it will install dependancies)
2. npm run docs:dev (this command will compile .md and it will start node server)

image

image

image

image


cijagani:
4) to Deploying doc file run `npm docs:build`
more informations can be [found here](https://vuepress.vuejs.org/guide/deploy.html#github-pages)

cijagani:
Didn't received any reply since long time. so i am closing this PR.

kenjis:
I don't know VuePress well, but I want to ask why VuePress?

datamweb:
I'm not familiar with [VuePress GitHub](https://github.com/vuejs/vuepress) either, from a quick look at the docs I found it's a good tool, also very popular with developers.
Let @cijagani give us more details.

lonnieezell:
VuePress might be a decent option, I haven't looked into it. I do think it would be good to get small sub-sites up for these packages, but none of us have had time to do it. When we do that, though, I imagine something like Jekyll that is directly supported by GitHub Pages is probably the route we would want to take.

cijagani:
@lonnieezell yes Jekyll is also good option.

kenjis:
Why not Jekyll?
Do you have a reason?

datamweb:
@kenjis I don't disagree with Jekyll, I read a bit about it today after [lonnieezell](https://github.com/lonnieezell) introduced **Jekyll**.
I trust lonnieezell decisions and I'm sure he has logical reasons for his decisions.
Do you have the necessary mastery of **Jekyll**?

@cijagani Thank you for your reply. 
From your reply, I realized that your PR goal was to improve Shield documents and that you have no particular bias. Thank you for this.

lonnieezell:
I don’t have any experience with either of them but I want it to be as simple to integrate into GitHub actions as possible. I would have to check into any solution and how easy to integrate it is for automatic builds.

kenjis:
GitHub Pages supports Jekyll.
https://docs.github.com/en/pages/setting-up-a-github-pages-site-with-jekyll/creating-a-github-pages-site-with-jekyll

datamweb:
Well, it wasn't as hard as I thought, I think this is what you want.
Of course, it needs some changes, also we need a subdomain, for example **shield.codeigniter.com**,
I will try to work a little more to have a complete understanding of this topic.
Of course, if everyone agrees with this change.
https://datamweb.github.io/ShieldSite/

kenjis:
See https://kenjis.github.io/codeigniter-shield/

Just setting on the repository, after it, alll we need to do is to customize theme.

Screenshot 2022-09-01 9 08 55

datamweb:
> Just setting on the repository, after it, alll we need to do is to customize theme.

I wonder, it was wonderful.
After the configuration, we can edit and update only the `anyfile.md` files as before. This is great.
[just-the-docswas](https://github.com/just-the-docs/just-the-docs) also interesting for me. It also has a search box.

lonnieezell:
Wow that looks like a lot easier than I thought it would be! That's definitely something we can work with. Thanks for exploring these everyone. Hopefully we can get a theme that is similar in feel to the main docs, though I don't imagine we'll get the same layout. Guess We'll need to update the index page with some About Shield information, too.

MGatner:
I have a strong bias for white backgrounds on documentation. It's the best for quick reading and I need my docs to be incredibly utilitarian. Almost all these themes are grey background - maybe we could change that one aspect though? My specific thoughts:
* Like Lonnie I enjoy Slate, minus the background
* Cayman has an almost-white background and good typography, overall pleasant
* Time Machine to me is the most utilitarian (while still being appealing) except the preview site has some responsive issues on mobile (maybe that's just an issue  with the preview?)

kenjis:
I think we can use custom CSS.
See https://github.com/pages-themes/architect#stylesheet

lonnieezell:
Once we get a PR merged for this I can setup the sub-domains.

PR #30, opened 2022-04-06
Simplification and unification filters

nowackipawel:


src/Config/Auth.php
```diff
@@ -74,6 +74,16 @@ class Auth extends BaseConfig
...
     ];
 
+    /**
+     * --------------------------------------------------------------------
+     * Name of Authenticator Header
+     * --------------------------------------------------------------------
+     * The name of Header that
+     */
+    public $authenticatorHeader = [
+        'tokens' => 'Authorization',
```
lonnieezell, line 84: Can you provide an example when changing this header name would be desirable?  This header is part of the [HTTP authorization spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#the_general_http_authentication_framework).

nowackipawel: In general you are right - that's why for sure default name should be Authorization.
In some cases when you already have an mobile app that is sending tokens with different header name + do not have opportunity to update it it could be useful.

PR #32, opened 2022-04-06
Form UX  + named routes + $_SESSION -> session()

nowackipawel:
- token (new input name, as is expected in Email2FA)
- right keyboard
- right input mode
- pattern
- set "one-time-code" that could help sometimes

### DIFF
--- a/src/Views/email_2fa_verify.php
+++ b/src/Views/email_2fa_verify.php
@@ -15,7 +15,7 @@
 			<div class="alert alert-danger"><?= session('error') ?></div>
 			<?php endif ?>
 
-			<form action="<?= site_url('auth/a/verify') ?>" method="post">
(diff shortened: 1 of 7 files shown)


src/Views/email_2fa_verify.php
```diff
@@ -20,7 +20,8 @@
 
 				<!-- Code -->
 				<div class="mb-2">
-					<input type="text" class="form-control" name="code" autocomplete="code" placeholder="000000" required />
+					<input type="number" class="form-control" name="token" autocomplete="token" placeholder="000000" 
+                           inputmode="numeric" pattern="[0-9]*" autocomplete="one-time-code" required />
```
lonnieezell, line 24: You have autocomplete on here twice.

Issue #33, opened 2022-04-08
Is force email verification after registration and Email 2FA working ? 


jozefrebjak:
If `Force email verification after registration` is enabled it will redirect after register form to  `auth/a/show` route, but it return

```
404 - File Not Found
Sorry! Cannot seem to find the page you were looking for.
```

The same behaviour with `Force 2FA check after login`

nowackipawel:
Yes, it is working nicely.

jozefrebjak:
I'm tried to find where I have something wrong, but I can't find it. I noticed there is probably a mistake in comment within file `/src/Config/Auth.php`:

```php
 - register: Shield\Authentication\Actions\EmailActivate
```

There is no `EmailActivate` action, but there is `EmailActivator` action within `/src/Authentication/Actions`, so I changed it to correct action, but nothing changed. It seems like a some kind a problem with registering an action. Magic link is working as expected for me, but email verification and 2FA not. @lonnieezell I tried it with Bonfire2 and it's the same for me.

jozefrebjak:
I found an issue. We need to specify correct actions like:

Wrong

```php
     - login:    Shield\Authentication\Actions\Email2FA
     - register: Shield\Authentication\Actions\EmailActivate
``` 
 
Correct

```php
     - login:    Sparks\Shield\Authentication\Actions\Email2FA
     - register: Sparks\Shield\Authentication\Actions\EmailActivator
```

jozefrebjak:
`Email2F` is working for me now. But how to get to work`EmailActivator`? . How we can get

```php
$user = auth()->user();
``` 

if we are not logged in within `/src/Authentication/Actions/EmailActivator.php`? 

@lonnieezell We are getting `error: Attempt to read property "id" on null`, because there is no logged in user, what make sense because We want to check user before login with Email code. 

@nowackipawel How it's working nicely for you ?

jozefrebjak:
@lonnieezell I can see another problem with `Email2F`. We can skip the whole 2F by just changing URL, because the user is already logged in. The whole 2F flow is trigged after logging in. OK, we have typed the correct username and password, but if we want to use 2F, we need to change something in database to not allow the user to access an application without 2F code.

lonnieezell:
[...]

PR #36, opened 2022-04-11
Permission check in chain login 2fa

nowackipawel:
PR #33 

* session field added to check if user finished its chain flow
* setting() usage instead of config in some places

what was fixed:
* if user did not finish 2fa it is redirected to login page if he is trying to access page that requires login.

what was messed up:
* I am not sure :D I hope nothing.

### DIFF
--- a/src/Authentication/Actions/Email2FA.php
+++ b/src/Authentication/Actions/Email2FA.php
@@ -86,6 +86,7 @@ class Email2FA implements ActionInterface
             return view(setting('Auth.views')['action_email_2fa_verify']);
         }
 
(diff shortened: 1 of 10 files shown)


nowackipawel:
What have to be done:
* implement this in auth_remember_token - I will take care about it later; but as long as I saw this issue 2 days ago ... I've got some idea for this so ... here it is.

lonnieezell:
Handled differently and fixed in https://github.com/lonnieezell/codeigniter-shield/commit/6af0f3a4a26cac81e510d8d7c2a1f4dcaafd14b9

PR #39, opened 2022-04-12
Setting instead of hardcoded model.

nowackipawel:


jozefrebjak:
@nowackipawel Great idea. I would also like to see something similar with `LoginController`. Now we need to extend Controllers if we want to add logic for example based on groups within `getLoginRedirect($user)`function.

MGatner:
@nowackipawel Could you please update the Config `Auth::$userProvider` to be typed to the user provider interface? That way we will be sure these calls are using something we can anticipate.

nowackipawel:
```
     * You can change it as long as they adhere to the
     * Sparks\Shield\Interfaces\UserProvider.```

**Comment by MGatner** [comment]
Yes, but also provide the type in the docblock. Since this is a string we can't really enforce it via PHP but static analysis will catch failures. Basically change the bottom of **src/Config/Auth.php** to be this:
```php

    /**
     * User Provider
     * The name of the class that handles user persistence.
     * By default, this is the included UserModel, which
     * works with any of the database engines supported by CodeIgniter.
     * 
     * @var class-string<\Sparks\Shield\Interfaces\UserProvider>
     */
    public $userProvider = 'Sparks\Shield\Models\UserModel';
```

(Or include a `use` statement at top and use the class basename).

nowackipawel:
@MGatner , I've never been an doc expert. Thx.

src/Controllers/RegisterController.php
```diff
@@ -90,7 +90,7 @@ public function registerAction()
      */
     protected function getUserProvider()
     {
-        return model('UserModel'); // @phpstan-ignore-line
+        return model(setting('Auth.userProvider')); // @phpstan-ignore-line
```
lonnie-vault, line 93:
```suggestion
        return model(setting('Auth.userProvider')); 
```

lonnieezell, line 93:
```suggestion
        return model(setting('Auth.userProvider')); 
```

lonnieezell:
please run `composer style` on the code to fix the style issues and commit those changes.

lonnieezell (approved):

MGatner (approved):

PR #46, opened 2022-04-19
test: replace  deprecated CIDatabaseTestCase

kenjis:
- add `DatabaseTestCase`
- remove deprecated `CIDatabaseTestCase`
- fix `TestCase` to use Array Settings Handler

### DIFF
--- a/tests/Authentication/SessionFilterTest.php
+++ b/tests/Authentication/SessionFilterTest.php
@@ -3,7 +3,8 @@
 namespace Tests\Authentication;
 
 use CodeIgniter\Config\Factories;
-use CodeIgniter\Test\CIDatabaseTestCase;
+use CodeIgniter\Test\CIUnitTestCase;
+use CodeIgniter\Test\DatabaseTestTrait;
 use CodeIgniter\Test\FeatureTestTrait;
 use Config\Services;
 use Sparks\Shield\Filters\SessionAuth;
@@ -12,9 +13,10 @@ use Sparks\Shield\Models\UserModel;
 /**
  * @internal
  */
-final class SessionFilterTest extends CIDatabaseTestCase
+final class SessionFilterTest extends CIUnitTestCase
 {
     use FeatureTestTrait;
+    use DatabaseTestTrait;
 
     protected $namespace;
 
--- a/tests/Authentication/SessionFilterTest.php
+++ b/tests/Authentication/SessionFilterTest.php
@@ -3,20 +3,18 @@
 namespace Tests\Authentication;
 
 use CodeIgniter\Config\Factories;
-use CodeIgniter\Test\CIUnitTestCase;
-use CodeIgniter\Test\DatabaseTestTrait;
 use CodeIgniter\Test\FeatureTestTrait;
 use Config\Services;
 use Sparks\Shield\Filters\SessionAuth;
 use Sparks\Shield\Models\UserModel;
+use Tests\Support\DatabaseTestCase;
 
 /**
  * @internal
  */
-final class SessionFilterTest extends CIUnitTestCase
+final class SessionFilterTest extends DatabaseTestCase
 {
     use FeatureTestTrait;
-    use DatabaseTestTrait;
 
     protected $namespace;
 
--- a/tests/Authentication/TokenFilterTest.php
+++ b/tests/Authentication/TokenFilterTest.php
@@ -3,7 +3,8 @@
 namespace Tests\Authentication;
 
 use CodeIgniter\Config\Factories;
-use CodeIgniter\Test\CIDatabaseTestCase;
+use CodeIgniter\Test\CIUnitTestCase;
+use CodeIgniter\Test\DatabaseTestTrait;
 use CodeIgniter\Test\FeatureTestTrait;
 use Config\Services;
 use Sparks\Shield\Entities\AccessToken;
@@ -14,9 +15,10 @@ use Sparks\Shield\Models\UserModel;
 /**
  * @internal
  */
-final class TokenFilterTest extends CIDatabaseTestCase
+final class TokenFilterTest extends CIUnitTestCase
 {
     use FeatureTestTrait;
+    use DatabaseTestTrait;
 
     protected $namespace;
 
--- a/tests/Authentication/TokenFilterTest.php
+++ b/tests/Authentication/TokenFilterTest.php
@@ -3,22 +3,20 @@
 namespace Tests\Authentication;
 
 use CodeIgniter\Config\Factories;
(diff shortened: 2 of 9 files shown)


MGatner (changes requested):
How about a new class with the trait and namespace that these extend instead? `Tests\Support\DatabaseTestCase`

It would make it easier to add methods or work with anything database-related in the tests.

kenjis:
Good suggestion! It would be more maintainable.

tests/_support/DatabaseTestCase.php
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Tests\Support;
+
+use CodeIgniter\Test\CIUnitTestCase;
+use CodeIgniter\Test\DatabaseTestTrait;
+
+/**
+ * @internal
+ */
+abstract class DatabaseTestCase extends CIUnitTestCase
```
lonnieezell, line 11: This should extend `Tests\Support\TestCase`.

tests/Authentication/SessionFilterTest.php
```diff
@@ -12,9 +13,10 @@
 /**
  * @internal
  */
-final class SessionFilterTest extends CIDatabaseTestCase
+final class SessionFilterTest extends CIUnitTestCase
```
lonnieezell, line 16: Shouldn't this extend the new DatabaseTestCase?

tests/Authentication/TokenFilterTest.php
```diff
@@ -14,9 +15,10 @@
 /**
  * @internal
  */
-final class TokenFilterTest extends CIDatabaseTestCase
+final class TokenFilterTest extends CIUnitTestCase
```
lonnieezell, line 18: Shouldn't this extend the new DatabaseTestCase?

tests/Controllers/RegisterTest.php
```diff
@@ -12,9 +13,10 @@
 /**
  * @internal
  */
-final class RegisterTest extends CIDatabaseTestCase
+final class RegisterTest extends CIUnitTestCase
```
lonnieezell, line 16: Shouldn't this extend the new DatabaseTestCase?

lonnieezell (changes requested):

kenjis:
@MGatner @lonnieezell 
Because `TestCase` depends on `codeigniter/settings`, it means it depends on database.
https://github.com/lonnieezell/codeigniter-shield/blob/4801828e0e343652403ae0dd87f26a722f72841c/tests/_support/TestCase.php#L21

`TestCase` is the base test class for this project,
so all test cases would depends on database.

Then, since the `TestCase` does not run migrations, the table for settings is not created.
So all test cases that extend `TestCase` will cause the error:
> ErrorException : SQLite3::query(): Unable to prepare statement: 1, no such table: db_settings

If previous test cases did not create the table.

tests/_support/DatabaseTestCase.php
kenjis: Done.

tests/Authentication/SessionFilterTest.php
kenjis: Done.

tests/Authentication/TokenFilterTest.php
kenjis: Done.

tests/Controllers/RegisterTest.php
kenjis: Done.

kenjis:
Changed TestCase to use the Array Settings Handler. This removed the dependency on the database.

MGatner (approved):
Array handler is the correct solution! We should probably provide a trait to handle that automatically from the Settings repo. Looks good here!

kenjis:
> We should probably provide a trait to handle that automatically from the Settings repo. 

Yes, it is needed definitely!

kenjis:
I did not know that assigning null to `$namespace` means `--all`!
It is not intuitive, and it is not documented.

I wondered why the settings table had been created, but I finally figured out why.

kenjis:
>  most of the tests end up hitting the database anyway.

Generally it is not good practice. We need more pure unit tests.
Integration tests like database tests should be less than unit tests.

Testing Pyramid : How to jumpstart Test Automation | BrowserStack
https://www.browserstack.com/guide/testing-pyramid-for-test-automation

kenjis:
@lonnieezell 
If there is some problem and you have not merged this PR, please point it out to me.

lonnieezell:
> Generally it is not good practice. We need more pure unit tests.
> Integration tests like database tests should be less than unit tests.

I used to believe this, but don't anymore. Now I think you should unit test your business logic classes, absolutely. But integration testing at the highest level possible ensures that you show that all functionality is working, but doesn't require you to test the implementation any more than required. This makes it simpler to refactor code and still be sure the end user's (or developers) stuff still works, without refactoring a hundred tests (I've been there and it's horrible). Less mocking of the database layer also means a) less time setting up mocks and b) provides something much closer to real-world usage. All of those sounded better to me the more I tested. 

I've found that on most projects simply using an in-memory SQLite database works for the majority of use-cases and is quite fast. Granted, if your code base gets too large it can still be an issue, and you might want to make different decisions. This library is not that big, though.  

I'll look over this PR again tonight.

lonnieezell:
> I did not know that assigning null to $namespace means --all!
> It is not intuitive, and it is not documented.

I agree. MGatner had to point that out to me a while back. I guess I forgot to update the docs.

kenjis:
> I think you should unit test your business logic classes, absolutely.

Yes.

Understood your opinion. I have not yet been able to see the specific test code for this library.

At least TestCase is no longer dependent on the database, so probably there are not any major problems.
And yes, this library is certainly still relatively small.

kenjis:
> MGatner had to point that out to me a while back. I guess I forgot to update the docs.

It was documented in several places.

But it is not documented in Testing `$namespace`, I sent a PR:
https://github.com/codeigniter4/CodeIgniter4/pull/5915

lonnieezell (approved):

MGatner:
The `null` namespace was a workaround/hack because we didn't have a way to specify "all namespaces". If I could go back I would do it differently, but here we are. 😓

MGatner:
Re: testing, Rails has a really slick feature that I want to bring to CodeIgniter. They roll up all the migrations into a database schema, and tests use the schema to "flash" the database. Because all you're doing is dropping the database and recreating it entirely in one step it is very fast, even though Ruby is generally slower than PHP. Our team has some (IMO) rather messy tests that normally would be very costly but the whole suite still runs very quickly because of this.

PR #52, opened 2022-04-20
chore: add Psalm

kenjis:
See https://psalm.dev/

The current type coverage is 81%.

> Psalm was able to infer types for 81.4696% of the codebase

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,8 @@
         "nexusphp/tachycardia": "^1.0",
         "php-coveralls/php-coveralls": "^2.4",
         "phpstan/phpstan": "^1.0",
-        "phpunit/phpunit": "^9.1"
+        "phpunit/phpunit": "^9.1",
+        "vimeo/psalm": "^4.22"
     },
     "config": {
         "allow-plugins": {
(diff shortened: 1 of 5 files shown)


lonnieezell:
Why do we need a second static analysis tool?

kenjis:
Psalm could fix types in some places automatically.

MGatner (commented):
What does Psalm bring that isn't covered by PHPStan and Rector? If there is a significant contribution, should we incorporate this into the DevKit as well?

MGatner:
Ah sorry I missed the conversation above. CS Fixer can convert docblock types to actual types (`phpdoc_to_property_types => true`). Rector also has a rule to do the same. Is Psalm somehow deriving typed from their context, at a deeper level than these two?

kenjis:
@MGatner I think we should use Psalm, too.

Checkout this PR, and try.

```
$ gh pr checkout #52
$ composer update
$ vendor/bin/psalm
$ vendor/bin/psalm --show-info=true
```

lonnieezell (approved):

PR #54, opened 2022-04-20
docs: use "remember-me" everywhere

kenjis:


lonnieezell:
I'm not sure I see it hyphenated like that very often. What prompted this change?

kenjis:
Since *remember-me* seemed to be the most commonly used in the source code comments, I adapted it.
It is easy to search.

MGatner:
I like the hyphen or quotes:
* remember-me token
* "remember me" token

We could also use a noun adjunct:
* remembrance token
* persistence token

lonnieezell:
I guess I didn't remember I had hyphenated it anywhere lol

We can hyphenate, that's fine.

PR #56, opened 2022-04-21
refactor: remove unused Response class

kenjis:
I believe the class is not used.


kenjis:
Where is the Response used? I was not able to find.
Can you show me one exact line?

lonnieezell:
I'm sorry. You're correct. That class was moved and renamed to `Sparks\Shield\Result`.

PR #59, opened 2022-04-21
refactor: use Interface methods

kenjis:
- use Interface methods instead of methods in concrete classes

Fixes 12 errors of 35: https://github.com/lonnieezell/codeigniter-shield/pull/44#issuecomment-1105787493

### DIFF
--- a/src/Authentication/Actions/Email2FA.php
+++ b/src/Authentication/Actions/Email2FA.php
@@ -25,7 +25,7 @@ class Email2FA implements ActionInterface
 
         // Delete any previous activation identities
         $identities = new UserIdentityModel();
-        $identities->where('user_id', $user->id)
+        $identities->where('user_id', $user->getAuthId())
             ->where('type', 'email_2fa')
             ->delete();
 
@@ -34,7 +34,7 @@ class Email2FA implements ActionInterface
         $code = random_string('nozero', 6);
 
         $identities->insert([
-            'user_id' => $user->id,
+            'user_id' => $user->getAuthId(),
             'type'    => 'email_2fa',
             'secret'  => $code,
             'name'    => 'login',
@@ -56,12 +56,12 @@ class Email2FA implements ActionInterface
         $email = $request->getPost('email');
         $user  = auth()->user();
 
-        if (empty($email) || $email !== $user->email) {
+        if (empty($email) || $email !== $user->getAuthEmail()) {
             return redirect()->route('auth-action-show')->with('error', lang('Auth.invalidEmail'));
         }
 
         $identities = new UserIdentityModel();
-        $identity   = $identities->where('user_id', $user->id)
+        $identity   = $identities->where('user_id', $user->getAuthId())
             ->where('type', 'email_2fa')
             ->first();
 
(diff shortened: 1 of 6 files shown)


lonnieezell:
I am not a fan of this one. Some of things, like `id` are a class property on the entity. I don't see a need to pull it from an interface method. Using `getAuthX` methods instead of simple properties makes it less readable and harder to remember, and doesn't always apply. If I'm getting the email address, auth shouldn't have anything to do with it. 

There's got to be a different way to make Stan happy, because this doesn't work for me.

kenjis:
# Interfaces
An interface is a contract between the code and our developers:
1. A class that implements an interface must have all the defined methods.
2. An interface cannot be changed without a major release.
3. **Anything that uses an interface in a typehint or docblock must use only that interface's methods.**

https://github.com/codeigniter4/CodeIgniter4/issues/4356

---

Now `Email2FA` depends on `Auth`:
https://github.com/lonnieezell/codeigniter-shield/blob/270ec98c6d26ee96b9f5bc56e25c25d6ffa38872/src/Authentication/Actions/Email2FA.php#L24

And depends on the return value type `Authenticatable`:
https://github.com/lonnieezell/codeigniter-shield/blob/270ec98c6d26ee96b9f5bc56e25c25d6ffa38872/src/Auth.php#L66

That is `Email2FA` depends on `Authenticatable` (Interface), not `User` (concrete class).
So you can't use any method or property on `User` except methods defined in `Authenticatable`.
This is not a matter of someone liking or disliking. If you break the rule, the interface will be meaningless.
Because if someone implements a class that follows the interface, it is not guaranteed to work.
If the interface methods are hard to remember, then the method names need to be changed.

The true issue is `Email2FA` should depend on `Authenticatable` or not.
If it depends on `User`, of course you can use all properties and methods in it.
But if users want to use their own `User` class, they can't use it with `Email2FA`.

MGatner:
An alternative would be to ditch the `Authenticatable` interface altogether and use `User` everywhere. Developers who want to use their own entity just need it to extend `User`. 

I am a fan of the interoperable Auth interfaces (see `Tatter\Users`) but I don't know that trying to provide that from within a package adds any benefit. If this seems important then maybe a better approach would be to have separate packages - one could be `CodeIgniter\Auth` with just the interfaces and the basic services, and then `Shield` could be an implementation of that.

lonnieezell:
My apologies. I jumped the gun last night. I've been sick and tired the last few days and apologize. But, yes, I do understand how interfaces work. I must have been thinking that was being used somewhere other than where it was and forcing that on users for some silly reason. Again - sick lol. 

Reviewing it again it looks fine. Thanks for your good eye, @kenjis

lonnieezell:
Now if only we could get these actions to authenticate again. I need to look into that more, because it's starting to drive me batty.

kenjis:
@lonnieezell No problem. You did not do anything that you need to apology.
Everybody has own preference, and you just stated your preference.
It is legal in my country and should not be condemned.

I am considering interface design for this library.
There is `src/Interfaces` directory
https://github.com/lonnieezell/codeigniter-shield/tree/develop/src/Interfaces
and it has `Authenticatable` and `UserProvider`.
So it seems to me that it is assumed that a user can replace them
(without extending existing concrete classes).

lonnieezell:
@kenjis Yes, I agree with that.

lonnieezell (approved):

MGatner:
> Now if only we could get these actions to authenticate again

My hunch is that it is some sort of caching, since the actual Action call now is un-authenticated. It looks like the last batch of workflows all succeeded, so maybe we'r in the clear? I'm keeping an eye on it.

PR #60, opened 2022-04-22
refactor: remove unused param $user

kenjis:


src/Controllers/LoginController.php
```diff
@@ -62,16 +62,14 @@ public function logoutAction(): RedirectResponse
...
+        return redirect()->to($this->getLogoutRedirect());
     }
 
     /**
      * Returns the URL that a user should be redirected
      * to after a successful login.
-     *
-     * @param mixed $user
      */
-    public function getLoginRedirect($user): string
```
lonnieezell, line 74: These need to stay in the `getLoginRedirect()` and `getLogoutRedirect()` method definitions. They are there so that when a dev overrides it they can have the user to check role of, etc. Actually - nevermind, they should be logged in at that point and should be able to use `auth()->user()` to get the user it needs.

lonnieezell (approved):

PR #63, opened 2022-04-23
refactor: move SQL logic into UserIdentityModel

kenjis:
Refactor to revisit the interfaces. This is the first PR.

- Move SQL related to `auth_identities` table to `UserIdentityModel`.

### DIFF
--- a/src/Authentication/Actions/Email2FA.php
+++ b/src/Authentication/Actions/Email2FA.php
@@ -23,17 +23,17 @@ class Email2FA implements ActionInterface
     {
         $user = auth()->user();
 
+        /** @var UserIdentityModel $identityModel */
+        $identityModel = model(UserIdentityModel::class);
+
         // Delete any previous activation identities
-        $identities = new UserIdentityModel();
-        $identities->where('user_id', $user->getAuthId())
-            ->where('type', 'email_2fa')
-            ->delete();
+        $identityModel->deleteIdentitiesByType($user->getAuthId(), 'email_2fa');
 
         // Create an identity for our 2fa hash
         helper('text');
         $code = random_string('nozero', 6);
 
-        $identities->insert([
+        $identityModel->insert([
             'user_id' => $user->getAuthId(),
             'type'    => 'email_2fa',
             'secret'  => $code,
@@ -60,10 +60,10 @@ class Email2FA implements ActionInterface
             return redirect()->route('auth-action-show')->with('error', lang('Auth.invalidEmail'));
         }
 
-        $identities = new UserIdentityModel();
-        $identity   = $identities->where('user_id', $user->getAuthId())
-            ->where('type', 'email_2fa')
-            ->first();
+        /** @var UserIdentityModel $identityModel */
+        $identityModel = model(UserIdentityModel::class);
+
+        $identity = $identityModel->getIdentityByType($user->getAuthId(), 'email_2fa');
 
         if (empty($identity)) {
             return redirect()->route('auth-action-show')->with('error', lang('Auth.need2FA'));
@@ -99,11 +99,11 @@ class Email2FA implements ActionInterface
             return view(setting('Auth.views')['action_email_2fa_verify']);
         }
 
+        /** @var UserIdentityModel $identityModel */
+        $identityModel = model(UserIdentityModel::class);
+
         // On success - remove the identity and clean up session
-        model(UserIdentityModel::class)
-            ->where('user_id', $user->getAuthId())
-            ->where('type', 'email_2fa')
-            ->delete();
+        $identityModel->deleteIdentitiesByType($user->getAuthId(), 'email_2fa');
 
         // Clean up our session
         session()->remove('auth_action');
--- a/src/Entities/User.php
+++ b/src/Entities/User.php
@@ -46,10 +46,14 @@ class User extends Entity implements \Sparks\Shield\Interfaces\Authenticatable
      */
     public function getIdentities(): array
     {
-        if (! array_key_exists('identities', $this->attributes) || ! is_array($this->attributes['identities'])) {
-            $this->attributes['identities'] = model(UserIdentityModel::class)
-                ->where('user_id', $this->id)
-                ->findAll();
+        if (
+            ! array_key_exists('identities', $this->attributes)
+            || ! is_array($this->attributes['identities'])
+        ) {
+            /** @var UserIdentityModel $identityModel */
+            $identityModel = model(UserIdentityModel::class);
+
+            $this->attributes['identities'] = $identityModel->getIdentities($this->id);
         }
 
         return $this->attributes['identities'];
@@ -92,10 +96,10 @@ class User extends Entity implements \Sparks\Shield\Interfaces\Authenticatable
      */
     public function getEmailIdentity()
     {
-        return model(UserIdentityModel::class)
-            ->where('user_id', $this->id)
-            ->where('type', 'email_password')
-            ->first();
+        /** @var UserIdentityModel $identityModel */
+        $identityModel = model(UserIdentityModel::class);
+
+        return $identityModel->getIdentityByType($this->id, 'email_password');
     }
 
     /**
--- a/src/Entities/User.php
+++ b/src/Entities/User.php
@@ -81,7 +81,10 @@ class User extends Entity implements \Sparks\Shield\Interfaces\Authenticatable
      */
     public function createEmailIdentity(array $credentials)
     {
-        model(UserIdentityModel::class)->insert([
+        /** @var UserIdentityModel $identityModel */
+        $identityModel = model(UserIdentityModel::class);
+
+        $identityModel->insert([
             'user_id' => $this->id,
             'type'    => 'email_password',
             'secret'  => $credentials['email'],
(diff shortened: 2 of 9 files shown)


lonnieezell (approved):

src/Authentication/Actions/Email2FA.php
```diff
@@ -23,17 +23,17 @@ public function show()
...
 
+        /** @var UserIdentityModel $identityModel */
+        $identityModel = model(UserIdentityModel::class);
+
         // Delete any previous activation identities
-        $identities = new UserIdentityModel();
-        $identities->where('user_id', $user->getAuthId())
-            ->where('type', 'email_2fa')
-            ->delete();
+        $identityModel->deleteIdentitiesByType($user->getAuthId(), 'email_2fa');
```
MGatner, line 30: In general I like the plural noun for the model variable, like `$identities` - it makes the code read smoothly. Not opposed to the more utilitarian version including "Model", but a style consideration we should probably be consistent with.

src/Entities/User.php
```diff
@@ -46,10 +46,14 @@ public function getIdentity(string $type)
...
-                ->where('user_id', $this->id)
-                ->findAll();
+        if (
+            ! array_key_exists('identities', $this->attributes)
+            || ! is_array($this->attributes['identities'])
+        ) {
+            /** @var UserIdentityModel $identityModel */
+            $identityModel = model(UserIdentityModel::class);
+
+            $this->attributes['identities'] = $identityModel->getIdentities($this->id);
```
MGatner, line 56: I realize this preexists you but... I feel fairly certain that relations should not go in `$attributes`. I've been down this path and it can cause a lot of confusion around what is in an Entity. 

My usually solution is a private/protected property with getter/setter methods; this also helps with static analysis.

src/Models/UserIdentityModel.php
```diff
@@ -25,6 +26,146 @@ class UserIdentityModel extends Model
     ];
     protected $useTimestamps = true;
 
+    public function getAccessTokenByToken(string $token): ?AccessToken
```
MGatner, line 29: Can we differentiate these any better? Get AccessToken by _____... String? Hash? Input? Credential?

src/Models/UserIdentityModel.php
```diff
@@ -25,6 +26,146 @@ class UserIdentityModel extends Model
...
+            ->where('type', 'access_token')
+            ->where('secret', hash('sha256', $token))
+            ->asObject(AccessToken::class)
+            ->first();
+    }
+
+    /**
+     * @param int|string $userId
+     */
+    public function getAccessToken($userId, string $token): ?AccessToken
```
MGatner, line 41: I ascribe to the "pass large" design philosophy, where you use the largest logical parameter for method input (here: `User $user`, or Authenticatable or whatever). This lets your validation logic remain outside the current method (I.e. you know that User exists, has a certain structure) and helps static analysis know your true intent (much less likely to pass a random string than a whole User). 

Probably you are being consistent with other methods that already exist, but if we're doing a refactor it is something I would like to consider. One "CodeIgniter-specific" issue this solves is our inconsistency around Entity primary keys - sometimes `int`, sometimes autoincrement (natural), sometimes string (UUID or other).

src/Models/UserIdentityModel.php
```diff
@@ -25,6 +26,146 @@ class UserIdentityModel extends Model
...
+     *
+     * @return AccessToken[]
+     */
+    public function getAllAccessTokens($userId): array
+    {
+        return $this
+            ->where('user_id', $userId)
+            ->where('type', 'access_token')
+            ->asObject(AccessToken::class)
+            ->find();
```
MGatner, line 76:
```suggestion
            ->findAll();
```
I realize both work but I want to move away from `find()` as a return type of single/multiple results.

src/Models/UserIdentityModel.php
```diff
@@ -25,6 +26,146 @@ class UserIdentityModel extends Model
...
+    }
+
+    /**
+     * @param int[]|string[] $userIds
+     *
+     * @return UserIdentity[]|null
+     */
+    public function getIdentitiesByUserIds(array $userIds)
+    {
+        return $this->whereIn('user_id', $userIds)->find();
```
MGatner, line 117:
```suggestion
        return $this->whereIn('user_id', $userIds)->findAll();
```

src/Models/UserIdentityModel.php
```diff
@@ -25,6 +26,146 @@ class UserIdentityModel extends Model
...
+    }
+
+    /**
+     * @param int|string $userId
+     *
+     * @return UserIdentity[]
+     */
+    public function getIdentities($userId): array
+    {
+        return $this->where('user_id', $userId)->findAll();
```
MGatner, line 107: `findAll()` Nailed it!

src/Models/UserIdentityModel.php
```diff
@@ -25,6 +26,146 @@ class UserIdentityModel extends Model
...
+    }
+
+    /**
+     * @param int|string $userId
+     */
+    public function getIdentityByType($userId, string $type): ?UserIdentity
+    {
+        return $this->where('user_id', $userId)
+            ->where('type', $type)
+            ->first();
```
MGatner, line 86: Why not consolidate these?
```suggestion
        return reset($this->getIdentitiesByTypes($userId, [$type]));
```

src/Models/UserIdentityModel.php
```diff
@@ -25,6 +26,146 @@ class UserIdentityModel extends Model
...
+    public function deleteIdentitiesByType($userId, string $type): void
+    {
+        $this->where('user_id', $userId)
+            ->where('type', $type)
+            ->delete();
+    }
+
+    /**
+     * Given the token, will retrieve the token to
+     * verify it exists, then delete it.
```
MGatner, line 145: More confusion over "token". But also, this appears to delete blindly without the "verify it exists" part.

MGatner (changes requested):
Good move towards centralization and consistency. A number of opinionated comments, open to debate!

src/Authentication/Actions/Email2FA.php
kenjis: `$identities` implies multiple identities, and there was the code to get multiple identities from the Model.
So I used `$identityModel` in the all places.

```php
        $identities = $identityModel->getIdentitiesByTypes(
            auth('session')->id(),
            ['email_2fa', 'email_activate']
        );
```

src/Entities/User.php
kenjis: Personally, I dislike magic properties. Preventing static analysis, and refactoring by IDE.
It is difficult to know what properties there.

You say so, it seems better to replace it with a private property with getter/setter methods.
I created Issue #65.

src/Models/UserIdentityModel.php
kenjis: Since everything else is passing `$userId`, I am not comfortable making this the only place where `User`|`Authenticatable` is used.

If we change it to pass the User, I don't think it will be a problem, but please make it a different PR.

src/Models/UserIdentityModel.php
kenjis: Changed to `getAccessTokenByRawToken()`.

src/Models/UserIdentityModel.php
kenjis: Done.

src/Models/UserIdentityModel.php
kenjis: Done.

src/Models/UserIdentityModel.php
kenjis: I am not sure what to do when it does not exist.

src/Models/UserIdentityModel.php
kenjis: Can't do that, because here is `first()`.

src/Authentication/Actions/Email2FA.php
MGatner: Acknowledged.

src/Entities/User.php
MGatner: Thanks! We'll get to that, shouldn't be difficult.

src/Models/UserIdentityModel.php
MGatner: That's a little better, at list it differentiates.

src/Models/UserIdentityModel.php
MGatner: Sounds good, I will create an issue.

MGatner: https://github.com/lonnieezell/codeigniter-shield/issues/66

src/Models/UserIdentityModel.php
MGatner: They should be functionally equivalent since they will end up using the same indexes. Is it a performance concern, loading all identities then returning just the first?

src/Models/UserIdentityModel.php
MGatner: I think your solution is fine, there's a lot of precedence for delete operations being idempotent - I would just update the comment:
```php
     * Delete any access tokens for the given rawToken
```

kenjis: Done!

src/Models/UserIdentityModel.php
kenjis: Right now it does not make sense to shorten the code of a small method. 
It is also not yet certain that the methods we have created here are really necessary.

We are still in the process of modifying and finalizing the api.

MGatner (approved):

MGatner:
@lonnieezell this is ready

MGatner:
🥳

PR #64, opened 2022-04-23
fix: [interface] User ID type

kenjis:
~~Needs to rebase after merging #63~~

It is too strict that only `int` can be used for user object ID.

### DIFF
--- a/src/Authentication/AuthenticatorInterface.php
+++ b/src/Authentication/AuthenticatorInterface.php
@@ -45,9 +45,11 @@ interface AuthenticatorInterface
      *
      * @see https://codeigniter4.github.io/CodeIgniter4/extending/authentication.html
      *
+     * @param int|string $userId
+     *
      * @return mixed
      */
-    public function loginById(int $userId);
+    public function loginById($userId);
 
     /**
      * Logs the current user out.
--- a/src/Authentication/AuthenticatorInterface.php
+++ b/src/Authentication/AuthenticatorInterface.php
@@ -46,10 +46,8 @@ interface AuthenticatorInterface
      * @see https://codeigniter4.github.io/CodeIgniter4/extending/authentication.html
      *
      * @param int|string $userId
-     *
-     * @return mixed
      */
-    public function loginById($userId);
+    public function loginById($userId): void;
 
     /**
      * Logs the current user out.
--- a/src/Interfaces/Authenticatable.php
+++ b/src/Interfaces/Authenticatable.php
@@ -15,7 +15,7 @@ interface Authenticatable
      * the User object for authentication purposes.
      * Typically the user's id.
      *
-     * @return mixed
+     * @return int|string|null
      */
     public function getAuthId();
 
--- a/src/Interfaces/Authenticatable.php
+++ b/src/Interfaces/Authenticatable.php
@@ -13,9 +13,9 @@ interface Authenticatable
     /**
      * Returns the unique identifier of
      * the User object for authentication purposes.
-     * Typically the user's id.
+     * Typically, the user's id.
      *
-     * @return int|string|null
+     * @return int|string
      */
     public function getAuthId();
 
--- a/src/Interfaces/Authenticatable.php
+++ b/src/Interfaces/Authenticatable.php
@@ -12,7 +12,7 @@ interface Authenticatable
 
     /**
      * Returns the unique identifier of
-     * the object for authentication purposes.
+     * the User object for authentication purposes.
      * Typically the user's id.
      *
      * @return mixed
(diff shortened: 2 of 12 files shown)


src/Authentication/AuthenticatorInterface.php
```diff
@@ -45,9 +45,11 @@ public function login(Authenticatable $user);
      *
      * @see https://codeigniter4.github.io/CodeIgniter4/extending/authentication.html
      *
+     * @param int|object|string $userId
```
MGatner, line 48: What would an `object` even be here? I propose dropping that.

MGatner (changes requested):

MGatner:
Most of the review will need to wait for rebase.

src/Authentication/AuthenticatorInterface.php
kenjis: Since it is an interface, I think it is better to include the possibilities.
I assumed a user who uses a value object.

MGatner: I'd actually argue the opposite: since it is an interface we should keep it tight. Adding the possibility for an `object` means everyone's authentication handler is going to have to check for and deal with an `object` with an unspecified structure. If someone wants to pass an object they could use a `Stringable` and still fit within this interface.

kenjis: Okay, this is Auth system. `int|string` is enough for User ID for authentication.
Fixed.

kenjis:
Rebased.

MGatner (changes requested):

src/Authentication/AuthenticatorInterface.php
```diff
@@ -45,9 +45,11 @@ public function login(Authenticatable $user);
      *
      * @see https://codeigniter4.github.io/CodeIgniter4/extending/authentication.html
      *
+     * @param int|string $userId
+     *
      * @return mixed
      */
-    public function loginById(int $userId);
+    public function loginById($userId);
```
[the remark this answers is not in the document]
kenjis: I think `void` would be better. These methods do not usually fail.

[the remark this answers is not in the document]
lonnieezell: It probably didn't come through, but I agree with `void` here.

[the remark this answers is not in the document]
kenjis: Okay, I will change it.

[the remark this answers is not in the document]
kenjis: Done.

MGatner (approved):

kenjis:
Rebased after namespace change.

src/Interfaces/Authenticatable.php
```diff
@@ -12,10 +12,10 @@ public function getAuthIdColumn(): string;
 
     /**
      * Returns the unique identifier of
-     * the object for authentication purposes.
+     * the User object for authentication purposes.
      * Typically the user's id.
      *
-     * @return mixed
+     * @return int|string|null
```
MGatner, line 18: In what scenario would this be `null`?

MGatner (changes requested):

src/Interfaces/Authenticatable.php
kenjis: Agreed. I like valid state objects.

In this case, the User object should have vaild ID, and if not it is a bug.

Done.

kenjis: Done!

src/Models/UserModel.php
```diff
@@ -114,6 +116,7 @@ public function addToDefaultGroup(User $user)
     public function fake(Generator &$faker)
     {
         return [
+            'id'       => $faker->randomNumber(5, true),
```
MGatner, line 119: I assume you added this to generate "valid" User instances but it will prevent auto-increment from working and could create collision if tests make multiple accounts.

MGatner (commented):
Not sure if it needs to change or not, see `fake()` comment.

src/Models/UserModel.php
kenjis: Ah, it is used to create records in the database?

I added it, because test failed.

kenjis:
```
There was 1 failure:

1) Tests\Authentication\AuthHelperTest::testUserIdReturnsId
User ID must not be null. in /Users/kenji/work/codeigniter/codeigniter-shield/src/Authentication/Traits/Authenticatable.php:31
```

MGatner: Yes, much better! Thanks.

MGatner: Note that another option is to use `fake()`'s third parameter to `$persist` the object into the database, which will assign it an ID: https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Helpers/test_helper.php

MGatner (approved):

PR #71, opened 2022-04-24
refactor: move getLoginRedirect(), getLogoutRedirect(), getRedirectURL() to Config\Auth

kenjis:
I feel uncomfortable with instantiating a controller.

### DIFF
--- a/src/Authentication/Actions/Email2FA.php
+++ b/src/Authentication/Actions/Email2FA.php
@@ -3,7 +3,7 @@
 namespace Sparks\Shield\Authentication\Actions;
 
 use CodeIgniter\HTTP\IncomingRequest;
-use Sparks\Shield\Controllers\LoginController;
+use Sparks\Shield\Config\Auth as AuthConfig;
 use Sparks\Shield\Models\UserIdentityModel;
 
 /**
@@ -109,6 +109,6 @@ class Email2FA implements ActionInterface
         session()->remove('auth_action');
 
         // Get our login redirect url
-        return redirect()->to(LoginController::getLoginRedirect());
+        return redirect()->to(AuthConfig::getLoginRedirect());
     }
 }
--- a/src/Authentication/Actions/Email2FA.php
+++ b/src/Authentication/Actions/Email2FA.php
@@ -109,6 +109,6 @@ class Email2FA implements ActionInterface
         session()->remove('auth_action');
 
         // Get our login redirect url
-        return redirect()->to(AuthConfig::getLoginRedirect());
+        return redirect()->to(AuthConfig::loginRedirect());
     }
 }
--- a/src/Authentication/Actions/Email2FA.php
+++ b/src/Authentication/Actions/Email2FA.php
@@ -109,8 +109,6 @@ class Email2FA implements ActionInterface
         session()->remove('auth_action');
 
         // Get our login redirect url
-        $loginController = new LoginController();
-
-        return redirect()->to($loginController->getLoginRedirect());
+        return redirect()->to(LoginController::getLoginRedirect());
     }
 }
--- a/src/Authentication/Actions/Email2FA.php
+++ b/src/Authentication/Actions/Email2FA.php
@@ -3,7 +3,6 @@
 namespace Sparks\Shield\Authentication\Actions;
 
 use CodeIgniter\HTTP\IncomingRequest;
-use Sparks\Shield\Config\Auth as AuthConfig;
 use Sparks\Shield\Models\UserIdentityModel;
 
 /**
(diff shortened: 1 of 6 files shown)


MGatner (changes requested):
Definitely agree. I'd go even further and say this doesn't belong in the controller at all. What about moving it to be a method on the Auth Config? 

Whatever we do should apply to `getLogoutRedirect()` as well.

lonnieezell:
After having lived with this for a bit working on Bonfire, I agree with moving it to the config file.

kenjis:
Moved to the Config file.

src/Authentication/Actions/Email2FA.php
```diff
@@ -109,8 +109,6 @@ public function verify(IncomingRequest $request)
         session()->remove('auth_action');
 
         // Get our login redirect url
-        $loginController = new LoginController();
-
-        return redirect()->to($loginController->getLoginRedirect());
+        return redirect()->to(AuthConfig::loginRedirect());
```
lonnieezell, line 112: You have this hard-coded to the Shield version of the config. I think this needs to use `config()` to ensure it can access the overridden version that will often be in `app/Config`. I think this might necessitate it being a non-static method, also.

```suggestion
        return redirect()->to(config('Auth')->loginRedirect());
```

kenjis: Oh, thanks. Indeed.

kenjis:
I don't know why but I got the following error twice in the past.
I ran `composer test` again, all tests passed.
Any clue?

```
1) Tests\Unit\PasswordsTest::testEmptyPassword
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'A Password is required.'
+'Auth.errorPasswordEmpty'
```

src/Authentication/Actions/Email2FA.php
kenjis: Fixed.

MGatner: Yes glad this was updated already but Config methods should always be on the instance since they could come from who-knows-where in the wild land of Config discovery 😅

MGatner:
@kenjis I've seen the language file failure a few times across different repos, not sure what causes it.

src/Config/Auth.php
```diff
@@ -325,4 +325,30 @@ class Auth extends BaseConfig
      * @var class-string<\Sparks\Shield\Interfaces\UserProvider>
      */
     public $userProvider = 'Sparks\Shield\Models\UserModel';
+
+    /**
+     * Returns the URL that a user should be redirected
+     * to after a successful login.
+     */
+    public static function loginRedirect(): string
```
MGatner, line 333: These will work on instances but I'm worried having them static might encourage developers to use them statically - any reason not to drop the static altogether?
```suggestion
    public function loginRedirect(): string
```

src/Config/Auth.php
```diff
@@ -325,4 +325,30 @@ class Auth extends BaseConfig
...
+    {
+        $url = setting('Auth.redirects')['login'];
```
MGatner, line 335: It feels very strange to call `setting()` when we already have the property instance. I suppose that's the only way we can allow Settings overrides in this case, unless anyone can think of a different approach? Maybe the Settings library needs an "auto-Registrar" option?

MGatner (changes requested):

src/Config/Auth.php
kenjis: In this PR, there is no other way to use the values in the database.

It seems Settings needs to get values from the database and set them when a Config class is instantiates.
Is it the "auto-Registrar" option?

kenjis:
It happened.

```
1) Tests\Unit\PasswordsTest::testEmptyPassword
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'A Password is required.'
+'Auth.errorPasswordEmpty'

/home/runner/work/codeigniter-shield/codeigniter-shield/tests/Unit/PasswordsTest.php:22
```
https://github.com/lonnieezell/codeigniter-shield/runs/6167178735?check_suite_focus=true

kenjis:
@MGatner @lonnieezell  I think this is complete.

lonnie-vault (approved):

src/Config/Auth.php
MGatner: My "auto-Registrar" idea is a way to have Settings participate in the Registrar stage of Config initialization, so instead of asking for the "settings-ified" value via `settings('Foo.bar')` devs would have the option of having those values already included in the Config file: `config('Foo')->bar`. This would also let Config files use their own properties without having to reroute

PR #73, opened 2022-04-25
Corrected typo error

Sammyskills:
Updated the call to the `setting('Auth.allowMagicLinkLogin')` to `setting('Auth.allowMagicLinkLogins')` as stated in the `Config/Auth.php` file.


kenjis (approved):

MGatner (approved):
Good catch!

PR #86, opened 2022-04-29
refactor: rename handler to authenticator

### LINKED ISSUE #67: Handler or Authenticator ?
In docs and in the namespace, `Handler` is used.
https://github.com/lonnieezell/codeigniter-shield/blob/develop/docs/2%20-%20authentication.md#available-handlers
https://github.com/lonnieezell/codeigniter-shield/tree/develop/src/Authentication/Handlers

But the interface name uses `Authenticator`.
https://github.com/lonnieezell/codeigniter-shield/blob/develop/src/Authentication/AuthenticatorInterface.php

If these terms refer to the same thing, they should be unified.

kenjis:
Fixes #67

### DIFF
--- a/src/Auth.php
+++ b/src/Auth.php
@@ -14,7 +14,7 @@ class Auth
     protected Authentication $authenticate;
 
     /**
-     * The handler to use for this request.
+     * The Authenticator to use for this request.
      */
     protected ?string $handler = null;
 
@@ -27,7 +27,7 @@ class Auth
     }
 
     /**
-     * Sets the handler that should be used for this request.
+     * Sets the Authenticator that should be used for this request.
      */
     public function setHandler(?string $handler = null)
     {
@@ -39,7 +39,7 @@ class Auth
     }
 
     /**
-     * Returns the handler name.
+     * Returns the Authenticator alias.
      *
      * @return string
      */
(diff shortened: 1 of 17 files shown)


MGatner (changes requested):
Looks good! Two minor preferences, open to debate:
* The only place I favor "Handlers" is for the folder name. This keeps it consistent with the main framework where the structure for a component directory is "provider/service, interface, handlers directory". 
* As a variable name "authenticatorAlias" is kind of bulky, especially where our assignment alignment style has to expand to match it. I don't think we are aliasing anything else - can we consider simple `$alias` instead?

src/Auth.php
```diff
@@ -27,25 +27,25 @@ public function __construct(Authentication $authenticate)
...
     }
 
     /**
-     * Returns the handler name.
+     * Returns the Authenticator alias.
      *
      * @return string
      */
-    public function getHandler()
+    public function getAlias()
```
kenjis, line 46: This method seems to be unused. Can we remove? @lonnieezell

kenjis:
Renamed authenticatorAlias to alias.

MGatner (approved):
Thanks! See if Lonnie chimes in on `getAlias()`

src/Auth.php
lonnieezell: You're right that we don't currently use it. I'm trying to think of use cases where it might be handy and I'm drawing a blank currently. Unless you can think of a use case when extending it or daily usage, it's probably safe to delete.

kenjis: Removed.

lonnieezell (approved):

PR #87, opened 2022-04-29
Use a PAT for hopefully less Github authentication issues.

lonnieezell:


.github/workflows/deptrac.yml
```diff
@@ -36,7 +36,7 @@ jobs:
           extensions: intl, json, mbstring, xml
           coverage: none
         env:
-          COMPOSER_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          COMPOSER_TOKEN: ${{ secrets. }}
```
kenjis, line 39: Missing variable name.

lonnieezell: Good catch. Thanks. Fixed.

kenjis (approved):

MGatner (commented):
Note that `GITHUB_TOKEN` is the recommendation here (https://github.com/shivammathur/setup-php). I think you may have reintroduced the auth issues with this.

lonnieezell:
Hm, well everywhere I look that talks about issues with authentication using GITHUB_TOKEN says to use a PAT if you're having problems. But this obviously didn't help out like we'd hoped. Damn. There's got to be a way to fix this. As it stands there's no way to reliably know if a PR messes something up or not.

MGatner:
Yeah, I'm not saying I know for sure as this has been going on across many different workflow iterations across many different repos - this is just what `ssteup-php` recommends (more precise link: https://github.com/shivammathur/setup-php#composer-github-oauth).

PR #91, opened 2022-04-30
Remove drop indexes. Fixes #90

### LINKED ISSUE #90: Migration rollback error
Running a `migrate:rollback` command returns the error below:

`[mysqli_sql_exception]`
`You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ''auth_identities_user_id'' at line 1`.

Further checks show that it has to do with the `auth_identies` table `('DROP INDEX IF EXISTS \'auth_identities_user_id\'')` at line `347`.

Please check.

lonnieezell:
The code wasn't needed as RDBMS will drop the index when the table is dropped.

### DIFF
--- a/src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
+++ b/src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
@@ -338,20 +338,6 @@ class CreateAuthTables extends Migration
 
         $this->db->disableForeignKeyChecks();
 
-        // drop constraints first to prevent errors
-        if ($this->db->tableExists('auth_remember_tokens')) {
-            $this->forge->dropForeignKey('auth_remember_tokens', 'auth_remember_tokens_user_id_foreign');
-        }
-        if ($this->db->tableExists('auth_identities')) {
-            $this->forge->dropForeignKey('auth_identities', 'auth_identities_user_id_foreign');
-        }
-        if ($this->db->tableExists('auth_groups_users')) {
-            $this->forge->dropForeignKey('auth_groups_users', 'auth_groups_users_user_id_foreign');
-        }
-        if ($this->db->tableExists('auth_permissions_users')) {
-            $this->forge->dropForeignKey('auth_permissions_users', 'auth_permissions_users_user_id_foreign');
-        }
-
         $this->forge->dropTable('users', true);
         $this->forge->dropTable('auth_logins', true);
         $this->forge->dropTable('auth_remember_tokens', true);
--- a/src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
+++ b/src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
@@ -344,16 +344,12 @@ class CreateAuthTables extends Migration
         }
         if ($this->db->tableExists('auth_identities')) {
             $this->forge->dropForeignKey('auth_identities', 'auth_identities_user_id_foreign');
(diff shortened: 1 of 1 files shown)


MGatner:
Test failures suggest there is something more going on here than just the FKs. 

In general I am not a fan of using "if exists..." in migrations (either the method or the second parameter) because it can hide issues. MigrationRunner should handle if/when the file runs, and at that point if it is "up" the table exists something has gone wrong prior.

kenjis:
It seems `group` causes an error on SQLite. This seems to be CI4 SQLite driver's bug.

```
ALERT - 2022-04-30 07:32:13 --> ALTER TABLE `db_auth_groups_users` RENAME TO `db_temp_auth_groups_users`
ALERT - 2022-04-30 07:32:13 --> CREATE TABLE `db_auth_groups_users` (
	`id` INTEGER NULL DEFAULT NULL,
	`user_id` INT NULL DEFAULT NULL,
	`group` varchar NULL DEFAULT NULL,
	`created_at` datetime NULL DEFAULT NULL,
	CONSTRAINT `pk_db_auth_groups_users` PRIMARY KEY(`id`)
)
ALERT - 2022-04-30 07:32:13 --> INSERT INTO db_auth_groups_users(id, user_id, group, created_at) SELECT id, user_id, group, created_at FROM db_temp_auth_groups_users
ERROR - 2022-04-30 07:32:13 --> ErrorException: SQLite3::exec(): near "group": syntax error in .../codeigniter-shield/vendor/codeigniter4/framework/system/Database/SQLite3/Connection.php:116
```
See `copyData()` in `system/Database/SQLite3/Table.php`

src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
```diff
@@ -344,17 +344,13 @@ public function down()
         }
         if ($this->db->tableExists('auth_identities')) {
             $this->forge->dropForeignKey('auth_identities', 'auth_identities_user_id_foreign');
-            $this->db->query('DROP INDEX IF EXISTS ?', [$prefix . 'auth_identities_user_id']);
         }
         if ($this->db->tableExists('auth_groups_users')) {
-            //            $this->forge->dropForeignKey('auth_groups_users', 'auth_groups_users_user_id_foreign');
+            $this->forge->dropForeignKey('auth_groups_users', 'auth_groups_users_user_id_foreign');
```
kenjis, line 349: Comment out this line.
Then the error will not occur.

lonnieezell:
> In general I am not a fan of using "if exists..." in migrations (either the method or the second parameter) because it can hide issues. MigrationRunner should handle if/when the file runs, and at that point if it is "up" the table exists something has gone wrong prior.

I get that and partially feel the same way, but I've had enough times where something fails in the middle of a migration during development that going back and cleaning up whatever had been done up to the failure point is a PITA. :)

lonnieezell:
That SQL for generating the table looks weird, not sure what the first `null` in each line is about, but something in there is causing an issue as it tries to recreate the table. 

While we should dig into that bug in `copyTable` for now I've removed all drop foreign key instructions. Locally tests pass in SQLite, and I confirmed that the foreign keys are removed from MySQL's `information_schema:key_column_usage` table. I thought they stuck around but that doesn't appear to be the case.

kenjis (approved):

kenjis:
I noticed that because `disableForeignKeyChecks()` is called, we don't need to delete foreign keys.
If tables are dropped, all keys also are deleted.

PR #94, opened 2022-05-01
getAuthEmail should check for getter method.

### LINKED ISSUE #83: strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated
```
There was 1 error:

1) Tests\Controllers\ActionsTest::testEmailActivateShow
ErrorException: strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated

/home/runner/work/shield/shield/vendor/codeigniter4/framework/system/Email/Email.php:728
/home/runner/work/shield/shield/vendor/codeigniter4/framework/system/Email/Email.php:541
/home/runner/work/shield/shield/src/Authentication/Actions/EmailActivator.php:44
/home/runner/work/shield/shield/src/Controllers/ActionController.php:54
/home/runner/work/shield/shield/src/Controllers/ActionController.php:42
/home/runner/work/shield/shield/vendor/codeigniter4/framework/system/CodeIgniter.php:830
/home/runner/work/shield/shield/vendor/codeigniter4/framework/system/CodeIgniter.php:419
/home/runner/work/shield/shield/vendor/codeigniter4/framework/system/CodeIgniter.php:326
/home/runner/work/shield/shield/vendor/codeigniter4/framework/system/Test/FeatureTestTrait.php:188
/home/runner/work/shield/shield/vendor/codeigniter4/framework/system/Test/FeatureTestTrait.php:222
/home/runner/work/shield/shield/tests/Controllers/ActionsTest.php:215
```
https://github.com/codeigniter4/shield/runs/6204722220?check_suite_focus=true

lonnieezell:
Fixes #83

### DIFF
--- a/src/Authentication/Traits/Authenticatable.php
+++ b/src/Authentication/Traits/Authenticatable.php
@@ -50,6 +50,11 @@ trait Authenticatable
     public function getAuthEmail()
     {
         $column = $this->getAuthEmailColumn();
+        $getter = 'get' . ucfirst($column);
+
+        if (method_exists($this, $getter)) {
+            return $this->{$getter}();
+        }
 
         return $this->{$column} ?? null;
     }


MGatner:
If the Authenticatable is a CI Entity this should already be handled by `$user->{$column}`.

kenjis:
> I think we will still need to make the action conditional on null emails.

Yes, we need it.
And the current failed test provides invalid User object that has null email.

lonnieezell:
> If the Authenticatable is a CI Entity (which Shield's User is) this should already be handled by the magic getter when calling $user->{$column}. Have you confirmed this fixes the issue?

Yes, I would have thought it should work, also. But I did confirm this was happening. I believe this is because we now have class properties on the User entity. Since the entity can be found, the null value is returned and `getEmail` is never given a chance to run.

MGatner:
Entity's `__get()` will always prefer a `getXxxx()` method if no public property is available: https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Entity/Entity.php#L474-487

If you experienced this issue then something else is probably wrong and I'd be inclined to dig into the issue. Nothing wrong with this change, but it is theoretically unnecessary.

kenjis:
This code is in the `User` entity. The magic `__get()` won't work.
The User has private $email property.

The `getAuthEmailColumn()` method seems to be no longer needed.
Does it really need?

MGatner:
The magic getter should work because it will short circuit on the `getEmail()` method when calling `$user->email`. My link above was missing an `L` to point this out, updated it now.

kenjis:
Not `$user->email`, but `$this->email`. `__get` is never called.

MGatner:
Ohhhh yes you are right. Okay this all makes sense now, and I'm out of helpful suggestions. At this point I think we should decide whether we will commit to `Authenticatable` being an `Entity` or `User` and then we can adjust accordingly.

kenjis:
I don't understand "Authenticatable being an Entity or User".
Authenticatable is an interface for User, is abstraction of User,
and User extends an Entity.

What's the difference between being an Entity and User?

MGatner:
@kenjis basically, get rid of `Authenticatable` the interface and just use `User`; developers who want to bring their own implementations will extend `User` to fulfill our other interfaces. Same with `UserProvider`, just use `UserModel`

kenjis:
Okay, I got it.

kenjis:
I believe this PR is out of dated, and the issue was already fixed.
Closing.

kenjis:
I believe this PR is out of dated, and the issue was already fixed.
Closing.

PR #99, opened 2022-05-01
Trigger didLogin Event on 2FA and EmailActivator actions

sammyskills:
Currently, the `didLogin` event is only set when there is a normal login.

Here's the current flow for the login 2FA action:
1. Users enters their details
2. A 2FA code is sent
3. Users retrieve the 2FA code and enter in the form
4. They are automatically **logged in** if code is valid.

I believe the `didLogin` event should also be fired at this point, since there is an actual login.

Same thing applies to the email activation process. Users are automatically logged in to the application after verifying their email address. The `didLogin` event should also be fired.


lonnieezell:
I agree with this addition. However, it looks like you have some styling issues that are being reported. Trying running `composer style` to have it auto-update the styles for you and commit any changes.

Sammyskills:
I am having some errors trying to run the `composer style` command:

`admin/setup.sh: line 2: $'\r': command not found`

Based on some research online, it turns out that because I'm using a windows machine and line-ends in windows is treated as `\r\n` instead of `\r` which of for classic macOS. 

If I make any changes to the `admin/setup.sh` file, since it is not included in the `.gitignore` file, it will be pushed to the repo, and that might cause a problem for other mac users.

Also, if I include the `admin/setup.sh` file in `.gitignore`, that means that subsequent contributors who fork the repo will not get the file.

Please,  what do you suggest ?

kenjis:
It seems you automatically convert LF to CRLF in your git repository.
Please check git config `core.autocrlf`.

Sammyskills:
This is what I have in my git config:

`[core]`
	`repositoryformatversion = 0`
`	filemode = false`
`	bare = false`
`	logallrefupdates = true`
`	symlinks = false`
`	ignorecase = true`
`[remote "origin"]`
`	url = https://github.com/Sammyskills/codeigniter-shield.git`
`	fetch = +refs/heads/*:refs/remotes/origin/*`
`[branch "develop"]`
`	remote = origin`
`	merge = refs/heads/develop`

kenjis:
Try and check the value of `core.autocrlf`.
```
$ git config -l
```
If it is true, set it false, and check out files again.

```
$ git config --local core.autocrlf false
```

Sammyskills:
Thanks @kenjis 

I just tried your suggestion and it worked.

kenjis:
@sammyskills I sent a PR to fix Events including this problems: #126
Please check it, if you can.

sammyskills:
Yeah @kenjis 

Your PR is more comprehensive. I'll just close this PR instead.

Thanks.

PR #102, opened 2022-05-03
Ensure 'except' routes are not in routes at all. Fixes #72

### LINKED ISSUE #72: Problem with route configuration
Hi,

I have been able to setup this package in my CI4 project to an extent, but I'm having issues with the routes configuration. 

From the [docs](https://github.com/lonnieezell/codeigniter-shield/blob/develop/docs/0%20-%20install.md#route-configuration) _(which doesn't explain much unfortunately)_, it states
`If you need to customize how any of the auth features are handled, you will likely need to update the routes to point to the correct controllers. You can still use the auth()->routes() helper, but you need to pass a list of routes to an except option:`
`auth()->routes($routes, ['except' => ['login', 'register']]);`

I have tried adding this in my App/Config/Routes.php file, but it returned an error: `Call to undefined function Config\auth()`.

I have already extended the controllers in my App/Controllers, but I don't know how to update the route to point to the new controllers.

Please help.

lonnieezell:


kenjis (approved):

MGatner (approved):
Very fine tests!

PR #108, opened 2022-05-11
fix: EmailActivator does not send email

### LINKED ISSUE #107: Email-based Account Activation does not send email

kenjis:
~~Needs to rebase after merging #44~~

Fixes  #107


MGatner:
Review will have to wait for rebase.

lonnieezell:
I think this would have been fixed by #94 if we are good to merge that.

kenjis:
I think we don't need `getAuth*()` including`getAuthEmail()` any more.

PR #112, opened 2022-05-11
refactor: move Traits/Authenticatable into User

kenjis:
Follow up #44


lonnieezell (commented):
Do the methods for things like `getAuthEmail`, `getAuthEmailFileld`, etc still make sense? If everything is based on the User entity now, then there's no chances they're changing the field for the id or email. 

Does it make sense to get rid of those methods and refactor the app to just user `->email` and `->password` instead?

kenjis:
The `get*Column()` methods do not make sense at all. Because now the user table does not have such columns.
I will remove them in another PR.

The `getAuth*()` methods may make sense in the future when users want to integrate this package into their own system, but now these don't make sense. Because this package is not yet very extensible. I think it is fine to remove once for now.

kenjis:
I sent a PR to remove `get*Column()`: #118

MGatner (approved):

PR #114, opened 2022-05-11
chore: update composer command and pre-commit

kenjis:
- update with reference to CodeIgniter4's


MGatner (approved):
This is good, but I'm also fine to remove the precommit hook. I don't use it myself, find it very pushy.

PR #110, opened 2022-05-11
Add error checks in Actions

kenjis:
- add User null check
- add check to see if the email sent

PR #119, opened 2022-05-13
Add email null check

kenjis:
~~Needs to rebase after merging #110~~

- add null check for email address


src/Authentication/Actions/EmailActivator.php
```diff
@@ -23,6 +24,13 @@ public function show(): string
             throw new RuntimeException('Cannot get the User.');
         }
 
+        $userEmail = $user->getAuthEmail();
```
lonnie-vault, line 27: At this point calling `getAuthEmail()` is the same as just calling `->email`, right? With no trait that needs the `getAuth*` methods do we even need them anymore?

kenjis: Yes, it is the same as calling `getEmail()`, so it is the same as `->email`.

About `getAuth*()` methods, we don't need them now. But when we introduce Interface for *User*, probably we will need them. So I am okay to remove `getAuth*()` now.

MGatner (approved):

PR #124, opened 2022-05-13
refactor: change return type to void in RememberModel

kenjis:
- change return type to void when *command* methods 
- and throw Exception if error occurs

### DIFF
--- a/src/Models/RememberModel.php
+++ b/src/Models/RememberModel.php
@@ -3,6 +3,7 @@
 namespace CodeIgniter\Shield\Models;
 
 use CodeIgniter\Model;
+use CodeIgniter\Shield\Exceptions\RuntimeException;
 use DateTime;
 use stdClass;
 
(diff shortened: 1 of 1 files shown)


lonnieezell (approved):

src/Models/RememberModel.php
```diff
@@ -24,25 +25,36 @@ class RememberModel extends Model
...
+
+    private function checkQueryReturn(bool $return): void
+    {
+        if ($return === false) {
+            $error   = $this->db->error();
+            $message = 'Query error: ' . $error['code'] . ', '
+                . $error['message'] . ', query: ' . $this->db->getLastQuery();
+
+            throw new RuntimeException($message, $error['code']);
+        }
```
MGatner, line 51: The similarities here make me wonder if we should consider a central error handling library. Maybe for the next version? I know Lonnie had a very nice PR on the main repo expanding and centralizing exception handling that might be applicable.

src/Models/RememberModel.php
```diff
@@ -24,25 +25,36 @@ class RememberModel extends Model
...
     }
 
     /**
      * Returns the remember-me token info for a given selector.
      *
-     * @return mixed
+     * @return stdClass|null
      */
     public function getRememberToken(string $selector)
```
MGatner, line 59: Can this be official?
```php
... $selector): ?stdClass
```

MGatner:
I didn't realize this one was merge already 🤦‍♂️

src/Models/RememberModel.php
kenjis: I think it is not good that the behavior of the database class changes depending on the value of `DBDebug`.
It should always throw an Exception on database errors.

MGatner: Agree! But as you know from recent conversations, I'm much more into throwing exceptions 🤗

PR #126, opened 2022-05-14
fix: Event triggers

kenjis:
Supersedes  #99

- fix Event names: now `failedLogin`, `login`, `logout`, `register`
- `login` is fired only once after login process is completed
- fix error invalidActivateToken does not show
- change return types
- add `Auth::activateUser()` and `UserModel::activate()`
- refactor

### DIFF
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -27,14 +27,17 @@ class Session implements AuthenticatorInterface
      */
     protected UserModel $provider;
 
+    /**
+     * The user logged in
+     */
     protected ?User $user = null;
-    protected LoginModel $loginModel;
 
     /**
      * Should the user be remembered?
      */
     protected bool $shouldRemember = false;
 
+    protected LoginModel $loginModel;
     protected RememberModel $rememberModel;
 
     public function __construct(UserModel $provider)
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -78,6 +78,7 @@ class Session implements AuthenticatorInterface
 
             // Fire an event on failure so devs have the chance to
             // let them know someone attempted to login to their account
+            unset($credentials['password']);
             Events::trigger('failedLoginAttempt', $credentials);
 
             return $result;
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -106,6 +106,8 @@ class Session implements AuthenticatorInterface
     }
 
     /**
+     * Check token in Action
+     *
      * @param string $type  Action type. 'email_2fa' or 'email_activate'
      * @param string $token Token to check
      */
@@ -134,6 +136,14 @@ class Session implements AuthenticatorInterface
         Events::trigger('login', $user);
 
         return true;
+    }
+
+    /**
+     * Activate a User
+     */
+    public function activateUser(User $user): void
+    {
+        $this->provider->activate($user);
     }
 
     /**
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -126,7 +126,7 @@ class Session implements AuthenticatorInterface
 
         // Remove the password from credentials so we can
         // check afterword.
-        $givenPassword = $credentials['password'] ?? null;
+        $givenPassword = $credentials['password'];
         unset($credentials['password']);
 
         // Find the existing user
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -79,7 +79,7 @@ class Session implements AuthenticatorInterface
             // Fire an event on failure so devs have the chance to
             // let them know someone attempted to login to their account
             unset($credentials['password']);
-            Events::trigger('failedLoginAttempt', $credentials);
+            Events::trigger('failedLogin', $credentials);
 
             return $result;
         }
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -315,7 +315,7 @@ class Session implements AuthenticatorInterface
     private function startLogin(User $user): void
     {
         // Update the user's last used date on their password identity.
-        $this->user->touchIdentity($this->user->getEmailIdentity());
+        $user->touchIdentity($user->getEmailIdentity());
 
         // Regenerate the session ID to help protect against session fixation
         if (ENVIRONMENT !== 'testing') {
@@ -323,7 +323,7 @@ class Session implements AuthenticatorInterface
         }
 
         // Let the session know we're logged in
-        session()->set(setting('Auth.sessionConfig')['field'], $this->user->getAuthId());
+        session()->set(setting('Auth.sessionConfig')['field'], $user->getAuthId());
 
         /** @var Response $response */
         $response = service('response');
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -105,13 +105,17 @@ class Session implements AuthenticatorInterface
         return $result;
     }
 
-    public function checkAction(string $token): bool
+    /**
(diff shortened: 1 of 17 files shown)


src/Authentication/Authenticators/Session.php
```diff
@@ -88,9 +93,59 @@ public function attempt(array $credentials): Result
...
+        $identityModel = model(UserIdentityModel::class);
+
+        // On success - remove the identity and clean up session
+        $identityModel->deleteIdentitiesByType($user->getAuthId(), $type);
+
+        // Clean up our session
+        session()->remove('auth_action');
+
+        // a successful login
+        Events::trigger('login', $user);
```
lonnieezell, line 136: Would this always be a login event?

lonnieezell: Looks like there's no event fired after EmailActivation.

src/Controllers/RegisterController.php
```diff
@@ -92,8 +92,10 @@ public function registerAction(): RedirectResponse
         }
 
         // Set the user active
-        $user->active = true;
-        $users->save($user);
+        auth()->activateUser($user);
+
+        // a successful login
+        Events::trigger('login', $user);
```
lonnieezell, line 98: I don't think `login` event should be fired during registration.

tests/Controllers/RegisterTest.php
```diff
@@ -57,13 +58,15 @@ public function testRegisterActionSuccess()
         $this->seeInDatabase('users', [
             'username' => 'JohnDoe',
         ]);
+
         // User has email/password identity
-        $user = model('UserModel')->where('username', 'JohnDoe')->first(); // @phpstan-ignore-line
+        $user = model(UserModel::class)->where('username', 'JohnDoe')->first();
```
lonnieezell, line 63: This uses the full class name to locate the class which doesn't allow app-level overrides. I think this should be reverted.

src/Authentication/Authenticators/Session.php
kenjis: The method is called in the action `Email2FA` or `EmailActivator`.
The `login` event is fired after token checking.
The event is fired after EmailActivation.

src/Controllers/RegisterController.php
kenjis: Registration without email activation logs in the user (call `login()` method).
The user can do without any restriction.

If we think registration is not login, and should not fire a login event,
then what do we do after email activation?

tests/Controllers/RegisterTest.php
kenjis: The full class name does not mean loading the specified class. If you have `App\Models\UserModel`, it will be loaded.
See https://codeigniter4.github.io/CodeIgniter4/concepts/factories.html#example

And this is a test for this package. I don't think app-level overrides does not matter.

lonnieezell: I had forgotten Factories would do that. Thanks for the reminder. And you're right - it's a test. Ignore me lol

src/Controllers/RegisterController.php
lonnieezell: I think that line can be deleted, actually. We have a `didRegister` event fired above (which should probably be renamed to simply `register` to match the login ones. And the `login()` method of the Session handler now fires off a login event for us. This is now redundant, I believe.

src/Authentication/Authenticators/Session.php
lonnieezell: The `login` should only be fired when a user manually logged in. Just checking the token (like a remember-me check) should not fire it off. 

Also, if the user has activated their email, the login action will fire the event for them. No need to do it here. If we still need it fired for the `Email2FA` action I'd move the event out to that class.

src/Controllers/RegisterController.php
kenjis: I removed firing login event in `Session::login()`, because when we use 2FA login is not completed at that point.

src/Authentication/Authenticators/Session.php
kenjis: In this case, the login event is fired after 2FA token checking.

>  If we still need it fired for the Email2FA action I'd move the event out to that class.

This seems to be a major point of disagreement.

I think the Action classes (`Email2FA`) are parts of controllers, and controllers are not a part of Auth library.
But Authenticators are a part of Auth library.
Auth library manages authentications and should fire login events.

So I removed `Events::trigger()` in controllers as much as possible.

src/Authentication/Authenticators/Session.php
```diff
@@ -88,9 +93,59 @@ public function attempt(array $credentials): Result
...
+     * @param string $type  Action type. 'email_2fa' or 'email_activate'
+     * @param string $token Token to check
+     */
+    public function checkAction(string $type, string $token): bool
+    {
+        $user = $this->loggedIn()
+            ? $this->getUser()
+            : null;
+
+        $identity = $user->getIdentity($type);
```
MGatner, line 120: If `$user` ever is actually `null` this will cause an error.

src/Models/UserModel.php
```diff
@@ -184,4 +185,27 @@ public function findByRememberToken($id, string $token): ?User
...
+    private function checkQueryReturn(bool $return): void
+    {
+        if ($return === false) {
+            $error   = $this->db->error();
+            $message = 'Query error: ' . $error['code'] . ', '
+                . $error['message'] . ', query: ' . $this->db->getLastQuery();
+
+            throw new RuntimeException($message, $error['code']);
+        }
+    }
```
MGatner, line 210: I love this dedicated error handling!

MGatner (changes requested):
It sounds like we still have some workflow refining/clarifying to do but I will defer to Lonnie. One small logic error.

src/Authentication/Authenticators/Session.php
kenjis: Added null check.

src/Authentication/Authenticators/Session.php
lonnieezell: That's fair, but login doesn't seem appropriate after EmailActivation. And that method covers all actions, doesn't it? So it would fire login after 2FA, but also after registration, which is not desirable. Especially since the `login` event is now triggered in the `login()` method itself. So this seems unnecessary, is all I'm saying. Am I understanding the flow correctly?

src/Controllers/RegisterController.php
lonnieezell: Oh, I missed that point. The latest code I pulled down yesterday had it in there I'm pretty sure.

https://github.com/codeigniter4/shield/blob/develop/src/Authentication/Authenticators/Session.php#L290

src/Authentication/Authenticators/Session.php
kenjis: > And that method covers all actions, doesn't it?

Yes.

> So it would fire login after 2FA, but also after registration, which is not desirable.

Yes. I think to fire login event after registration is okay. 

> Same thing applies to the email activation process.  Users are automatically logged in to the application after verifying their email address. The didLogin event should also be fired. https://github.com/codeigniter4/shield/pull/99

But if it should not, I will skip the `Events::trigger()` when 'email_activate'.

> Especially since the login event is now triggered in the login() method itself.

I removed the trigger in `login()` method in this PR. 
But the issue is what `login()` method does, what `login` is.

If `login` means authentication is complete, i.e. the user has no further Auth steps to take (see https://github.com/codeigniter4/shield/issues/128#issuecomment-1126691010),
and the `login()` method makes a user state so, the `login` event should be fired in it.

MGatner (approved):
The code looks very good, pending the discussion about events.

src/Authentication/Authenticators/Session.php
lonnieezell: > I removed the trigger in login() method in this PR.

You're right. I missed that. 

I think I'm convinced enough to allow that to stay. And I agree with a pending state until complete authorization.

kenjis:
I believe no major objections to this PR.
We still have some points to decide and implement, so I will merge this for now.

PR #127, opened 2022-05-14
refactor: extract Models from Authorizable

kenjis:
- extract GroupModel
- extract PermissionModel

### DIFF
--- a/src/Authorization/Models/GroupModel.php
+++ b/src/Authorization/Models/GroupModel.php
@@ -1,0 +1,58 @@
+<?php
+
+namespace CodeIgniter\Shield\Authorization\Models;
+
+use CodeIgniter\Model;
+
+class GroupModel extends Model
+{
+    protected $table          = 'auth_groups_users';
+    protected $primaryKey     = 'id';
+    protected $returnType     = 'array';
+    protected $useSoftDeletes = false;
+    protected $allowedFields  = [
+        'user_id',
+        'group',
+        'created_at',
+    ];
+    protected $useTimestamps      = false;
+    protected $validationRules    = [];
+    protected $validationMessages = [];
+    protected $skipValidation     = false;
+
+    /**
+     * @param int|string $userId
+     */
+    public function getByUserId($userId): array
+    {
+        $groups = $this->builder()
+            ->select('group')
+            ->where('user_id', $userId)
+            ->get()
+            ->getResultArray();
+
+        return array_column($groups, 'group');
+    }
+
+    /**
+     * @param int|string $userId
+     */
+    public function deleteAll($userId): void
+    {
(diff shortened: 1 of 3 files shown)


MGatner (approved):
This looks good. What I thought this PR was going to be was moving these models into **src/Models** which made me happy to the point that maybe we should consider it.

kenjis:
@MGatner Your English is a bit difficult to me. I don't understand what you exactly mean. 
Do you want to move the models into src/Models?

MGatner:
Sorry @kenjis. It is a compliment on your English, because usually I am mindful of my writing with non-native speakers, but you work so well in English sometimes I forget. 

Restated... I did not know that I would like to move the models until this PR. Probably at this point so close to release it would be unwise to do that; and this is still Lonnie's canvas and he clearly organized them like this one purpose. But I often go to the wrong folder looking for a model and have to stop to think about where it is.

MGatner:
ご迷惑

lonnieezell:
I didn't have a chance to comment on this before it was merged but I think this is a bad idea. While it makes sense logically, it forces everyone to always use the full namespace when creating a new instance, and is not able to just do: 

```
$groups = model('GroupModel');
```

I don't know what everyone else's practices are, but this is my preferred way of doing it so that if I extend the class in the App namespace it can be found without any code changes.

MGatner:
To be fair, these models didn't exist before this PR so nobody could use them with the `model()` command. Now they are available internally, or with the namespaced-class.

My comments above were partially about this same issue - that consolidating all Models to **src/Models/** makes them accessible via their shorthand. But it was mostly about where to go to look for the files :)

Also note, by default `Factories::model()` uses `preferApp` so even if you give the full namespace it will still use a version in **app/Models/**:
```php
$model = model('CodeIgniter\Shield\Authorization\Models\GroupModel');
echo $model::class; // App\Models\GroupModel
```

lonnieezell:
oh, dang. I keep forgetting the way your updates with `preferApp` work since I wasn't involved much in that. Ignore my comment :)

kenjis:
It was me that created `Shield\Authorization\Models\`, because the all code of Authorization was located in the `Authorization` folder, and it was easy to move the code to the models. `Shield\Models\` are bit far away.

But as a whole package the Models folder seems not to be good.
I will move the models to `Shield\Models\` later.

kenjis:
@MGatner "ご迷惑" is a bit funny!
It is a polite and correct Japanese word, but that word alone does not complete the meaning.
If you say "ご迷惑をおかけしました", it means "Sorry for the inconvenience".

MGatner:
Haha oops! Thanks for the correction, and ご迷惑をおかけしました

PR #129, opened 2022-05-15
fix: add missing unique key to `auth_identities`

kenjis:
Access Token or Magic Link Token must be unique.

### DIFF
--- a/src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
+++ b/src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
@@ -46,6 +46,7 @@ class CreateAuthTables extends Migration
             'updated_at'   => ['type' => 'datetime', 'null' => true],
         ]);
         $this->forge->addPrimaryKey('id');
+        $this->forge->addUniqueKey(['type', 'secret']);
         $this->forge->addKey('user_id');
         $this->forge->addForeignKey('user_id', 'users', 'id', '', 'CASCADE');
         $this->forge->createTable('auth_identities', true);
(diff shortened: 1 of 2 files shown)


lonnieezell (commented):
This would also ensure email is also unique for email/password identities, which I'm good with. 

However, email2FA uses the secret for the token also. Which currently does NOT ensure there is a unique value created since the number is tied with a user id and secured on the session currently. 

If we make this unique, which I'm ok with, we need both of those to generate a random number, then check the database to ensure it's unique, and repeat until it gets a unique value. Otherwise it will definitely have collisions at some point.

kenjis:
> This would also ensure email is also unique for email/password identities, which I'm good with.

Email must be unique, because it is a ID for a user. 

> However, email2FA uses the secret for the token also. Which currently does NOT ensure there is a unique value created since the number is tied with a user id and secured on the session currently.

Good point! I forgot about it.

> If we make this unique, which I'm ok with, we need both of those to generate a random number, then check the database to ensure it's unique, and repeat until it gets a unique value. Otherwise it will definitely have collisions at some point.

You are correct. It is only 6 digits.

I think Access Token and Magic Link Token should be unique. This is because if they conflict, one user could log in as another user.

MGatner (approved):
Good catch!

lonnieezell:
Agree. Make it unique - but that requires the other generated tokens are ensured they are unique as mentioned above. 

At first I was concerned we might need it not to be unique key when used with third-party integrations, but Oauth2 would just save an access token, as would JWT and probably most other uses, so this is a good change. Thanks.

PR #132, opened 2022-05-17
Add User State in Authenticators\Session

kenjis:
- add `$userState` in `Session`
- now `Session::loggedIn()` does not return `true` for pending login user.
- now `Session::getUser()` does not return pending login user, so `auth()->user()` also does not return.
- add `Session::getPendingUser()`
- add `Session::isPending()` that returns `true` when a user is in pending login state.
- add `Session::isAnonymous()`
- add `ActionInterface::getType()`
- refactor

### DIFF
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -452,7 +452,7 @@ class Session implements AuthenticatorInterface
     {
         if (! $this->user instanceof User) {
             throw new InvalidArgumentException(
-                self::class . '::recordActiveDate() requires logged in user before calling.'
+                __METHOD__ . '() requires logged in user before calling.'
             );
         }
 
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -22,15 +22,29 @@ use stdClass;
 
 class Session implements AuthenticatorInterface
 {
+    private const STATE_UNKNOWN   = 0;
+    private const STATE_ANONYMOUS = 1;
+    private const STATE_LOGGED_IN = 3;
+
     /**
      * The persistence engine
      */
     protected UserModel $provider;
 
     /**
-     * The user logged in
+     * Authenticated or authenticating (pending login) User
      */
     protected ?User $user = null;
+
+    /**
+     * The User auth state
+     */
+    private int $userState = self::STATE_UNKNOWN;
+
+    /**
+     * Pending login error message
+     */
+    private ?string $pendingMessage = null;
 
     /**
      * Should the user be remembered?
@@ -146,6 +160,8 @@ class Session implements AuthenticatorInterface
 
     private function completeLogin(User $user): void
     {
+        $this->userState = self::STATE_LOGGED_IN;
+
         // a successful login
         Events::trigger('login', $user);
     }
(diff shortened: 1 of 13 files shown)


src/Authentication/Authenticators/Session.php
```diff
@@ -249,28 +285,85 @@ public function loggedIn(): bool
         if ($userId !== null) {
             $this->user = $this->provider->findById($userId);
 
-            return $this->user instanceof User;
+            $identities = $this->userIdentityModel->getIdentitiesByTypes(
+                $this->user->getAuthId(),
+                ['email_2fa', 'email_activate']
```
lonnieezell, line 290: I think you need to get the list of actions assigned in the config file here, otherwise if some wants to do, says, 2FA with SMS messages, this logic would be skipped.

src/Authentication/Authenticators/Session.php
```diff
@@ -249,28 +285,85 @@ public function loggedIn(): bool
...
+    }
+
+    public function isPending(): bool
+    {
+        $this->checkUserState();
+
+        return $this->userState === self::STATE_PENDING;
+    }
+
+    public function isAnonymous(): bool
```
lonnieezell, line 335: Please add a quick docblock with why a user might be anonymous

src/Controllers/RegisterController.php
```diff
@@ -80,19 +83,21 @@ public function registerAction(): RedirectResponse
 
         Events::trigger('register', $user);
 
-        auth()->login($user);
+        /** @var Session $authenticator */
+        $authenticator = auth('session')->getAuthenticator();
+
+        $authenticator->login($user);
```
lonnieezell, line 89:
```suggestion
        auth('session')->getAuthenticator()->login($user);
```

Looks like this is the only time the variable is being used so no need to create one for this.

lonnieezell (changes requested):
Looks pretty good. I had a couple of notes, and it also needs information about the pending state added to the docs.

src/Authentication/Authenticators/Session.php
kenjis: Since `isPending()` was necessary, I also created `isAnonymous()` by the way.
But don't we need it? We are not using it at the moment.

lonnieezell: I don't have any concerns with the method/state. Those are both fine. But adding a description of what the `anonymous` represents within this library would save a developer a bit of time trying to parse through everything and understand.

MGatner (commented):
This is too big for me to review on mobile. It looks like Lonnie has been here so just let me know if I need to do anything on desktop later.

src/Authentication/Authenticators/Session.php
kenjis: I removed the hard coded action types.

src/Controllers/RegisterController.php
kenjis: It is used later.
https://github.com/codeigniter4/shield/blob/b7c308d72181dadc277a6d007263c4caa7ebde4b/src/Controllers/RegisterController.php#L100

src/Authentication/Authenticators/Session.php
kenjis: Added.

kenjis:
I believe this PR can now be merged.
Pending login status has been introduced and the `loggedIn()` behavior has been fixed.

One more refactoring would make the code more readable.

src/Controllers/RegisterController.php
lonnieezell: Good deal. I missed that.

lonnieezell (approved):

PR #134, opened 2022-05-17
refactor: remove unused methods in User

kenjis:
- remove `getRememberToken()` and `setRememberToken()`


lonnieezell (approved):
Side note - can we disable mutation testing for this? I find them very annoying :)

PR #140, opened 2022-05-17
style: fix indentation in view files

kenjis:
Ref https://github.com/codeigniter4/CodeIgniter4/pull/6039

- update editorconfig
- use spaces instead of tabs in Views

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,5 +11,5 @@ end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
 
-[*.md]
-trim_trailing_whitespace = false
+[*.yml]
+indent_size = 2
--- a/src/Views/email_2fa_verify.php
+++ b/src/Views/email_2fa_verify.php
@@ -5,32 +5,32 @@
 <?= $this->section('main') ?>
 
 <div class="container d-flex justify-content-center p-5">
-	<div class="card col-5 shadow-sm">
-		<div class="card-body">
-			<h5 class="card-title mb-5"><?= lang('Auth.emailEnterCode') ?></h5>
+    <div class="card col-5 shadow-sm">
+        <div class="card-body">
+            <h5 class="card-title mb-5"><?= lang('Auth.emailEnterCode') ?></h5>
 
-			<p><?= lang('Auth.emailConfirmCode') ?></p>
+            <p><?= lang('Auth.emailConfirmCode') ?></p>
 
-			<?php if (session('error') !== null) : ?>
-			<div class="alert alert-danger"><?= session('error') ?></div>
-			<?php endif ?>
+            <?php if (session('error') !== null) : ?>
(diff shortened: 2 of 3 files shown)


MGatner (approved):

src/Views/.editorconfig
```diff
@@ -0,0 +1,10 @@
+; This file is for unifying the coding style for different editors and IDEs.
+; More information at http://editorconfig.org
+
+[*]
+charset = utf-8
+indent_size = 2
+indent_style = tab
```
lonnieezell, line 7:
```suggestion
indent_size = 4
indent_style = space
```

Let's keep this consistent with CI itself.

kenjis: CI4 uses tabs in Views:
https://github.com/codeigniter4/CodeIgniter4/blob/develop/app/Views/welcome_message.php
https://github.com/codeigniter4/CodeIgniter4/blob/develop/app/Views/errors/html/error_404.php

lonnieezell: Is there something that automates that? Or did that just never get changed when we made the switch to follow PSR-12? I don't know of any way to enforce it only in views and not other files, and that just gets confusing for people.

kenjis: I think we did not change the indentation in Views, because these are not pure PHP source code.
In fact these are HTML files. So we still use tabs. And I think it is okay.

The `.editorconfig` enforces it only in this directory if users use `.editorconfig` aware IDEs.

MGatner: I'm not sure if it was an oversight or desired behavior, but the switch to spaces was never applied to Views because they are all exempt in our CS Fixer. I have maintained this behavior in my own projects, not out of some intentional preference but just to remain consistent with the framework core.

kenjis: I don't have strong preference for tab indentation. But I feel 2 space sized indentation is more common in HTML/CSS.

I also sent a PR to CI4 main repos. https://github.com/codeigniter4/CodeIgniter4/pull/6010

lonnieezell: Wouldn't it be simpler just to keep everything the same?

kenjis: @lonnieezell First, let's decide the rules for the main repository.
https://github.com/codeigniter4/CodeIgniter4/pull/6010

kenjis:
Now 4 spaces in Views.

lonnieezell (approved):

MGatner:
@kenjis How did you apply these changes? Did you open all view files and do a search-and-replace? Or find+sed? Or CS Fixer? I want to use this new style for my projects and I'm trying to decide the best way to apply it.

kenjis:
open all view files and do a search-and-replace.

I think php-cs-fixer does not fix non PHP files.

PR #145, opened 2022-05-18
refactor: change session variable

kenjis:
- change the session variable structure:
  - now uses one key
```php
[
    'user' => [
        'auth_action' => 'CodeIgniter\Shield\Authentication\Actions\Email2FA',
        'auth_action_message' => 'You must complete a two-factor verification.',
        'id' => '1',
    ]
]
```

- if there are more than one identity for actions, use the first one.
- if session has the user info, use it to check the user state.
- add code to set `auth_action` after checking valid remember-me token.
- refactor

### DIFF
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -136,6 +136,32 @@ class Session implements AuthenticatorInterface
     }
 
     /**
+     * If an action has been defined, start it up.
+     *
+     * @param string $type 'register'
+     */
+    public function startUpAction(string $type, User $user): bool
+    {
+        $actionClass = setting('Auth.actions')[$type] ?? null;
+
+        if ($actionClass === null) {
+            return false;
+        }
+
+        $action = Factories::actions($actionClass); // @phpstan-ignore-line
+
+        // E.g., afterRegister()
+        $method = 'after' . ucfirst($type);
+        if (method_exists($action, $method)) {
+            $action->{$method}($user);
+        }
+
+        $this->setSessionUser('auth_action', $actionClass);
+
+        return true;
+    }
+
+    /**
      * Check token in Action
      *
      * @param string $type  Action type. 'email_2fa' or 'email_activate'
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -118,17 +118,8 @@ class Session implements AuthenticatorInterface
         $this->recordLoginAttempt($credentials, true, $ipAddress, $userAgent, $user->getAuthId());
 
         // If an action has been defined for login, start it up.
-        $actionClass = setting('Auth.actions')['login'] ?? null;
-
-        if (! empty($actionClass)) {
-            $action = Factories::actions($actionClass); // @phpstan-ignore-line
-
-            if (method_exists($action, 'afterAttempt')) {
-                $action->afterAttempt($user);
-            }
-
-            $this->setSessionUser('auth_action', $actionClass);
-        } else {
+        $hasAction = $this->startUpAction('login', $user);
+        if (! $hasAction) {
             $this->completeLogin($user);
         }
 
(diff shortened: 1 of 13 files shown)


src/Authentication/Authenticators/Session.php
```diff
@@ -281,51 +315,63 @@ private function checkUserState(): void
...
 
             return;
         }
 
+        // No User Info in Session.
         // Check remember-me token.
         if (setting('Auth.sessionConfig')['allowRemembering']) {
-            $this->checkRememberMe();
+            if ($this->checkRememberMe()) {
+                $this->setAuthAction();
```
lonnieezell, line 340: I don't think I follow the logic here. Why only set the action if remember me is allowed and they checked it? If remember-me wasn't allowed you would still want to fire the action.

src/Authentication/Authenticators/Session.php
```diff
@@ -281,51 +315,63 @@ private function checkUserState(): void
...
     }
 
+    /**
+     * Gets identities for action from database, and set session.
+     */
+    private function setAuthAction()
+    {
+        // Get identities for action
+        $identities = $this->userIdentityModel->getIdentitiesByTypes(
+            $this->user->getAuthId(),
```
lonnieezell, line 356:
```suggestion
            $this->user->id,
```

Weren't we going to get rid of the `getAuthX` methods?

src/Authentication/Authenticators/Session.php
kenjis: The `checkRememberMe()` returns `true` if the user logged in by remember-me token.
If it returns `false`, there was no valid remember-me token. No session, no remember-me token,
that is the user is anonymous.

`setAuthAction()` needs user id to find identities for action. So it does not work when the user is anonymous.

src/Authentication/Authenticators/Session.php
kenjis: The param was changed to User. Rebased and fixed.

`getAuthId()` is still used in a lot of places.
We certainly do not need those methods now and I would like to remove them, 
but the task is not high on my priority list.

src/Authentication/Authenticators/Session.php
```diff
@@ -459,6 +508,38 @@ private function startLogin(User $user): void
...
+    {
+        $sessionUserInfo = $this->getSessionUserInfo();
+
+        return $sessionUserInfo[$key] ?? null;
+    }
+
+    /**
+     * @param int|string|null $value
+     */
+    private function setSessionUser(string $key, $value): void
```
MGatner, line 529: These two make me think they will get/set the user. What about `getSessionKey()` or `Data()` or `Field()`?

src/Authentication/Authenticators/Session.php
```diff
@@ -459,6 +508,38 @@ private function startLogin(User $user): void
...
+    {
+        $sessionUserInfo       = $this->getSessionUserInfo();
+        $sessionUserInfo[$key] = $value;
+        session()->set(setting('Auth.sessionConfig')['field'], $sessionUserInfo);
+    }
+
+    private function removeSessionUser(string $key): void
```
MGatner, line 536: Oops, same comment here.

MGatner (changes requested):
I like the refactor. A naming suggestion. 

Also, this has me thinking... there is quite a lot of "workflow logic" around actions that is not specific to Session. Is there a scenario where we use Actions for other authenticators? Or want to make them available for developers to leverage themselves? If so it might make sense to abstract them even more into a sort of "authentication middleware".

src/Authentication/Authenticators/Session.php
kenjis: Renamed to `getSessionKey()`.

src/Authentication/Authenticators/Session.php
kenjis: Done.

lonnieezell (approved):

MGatner (approved):

PR #146, opened 2022-05-18
Refactor: Object Parameters

### LINKED ISSUE #66: Pass class instances when available
I ascribe to the "pass large" design philosophy, where you use the largest logical parameter for method input (e.g. passing `User $user` or `Authenticatable $user` instead of `string|int $userId`). This lets your validation logic remain outside the current method (I.e. you know that User exists, has a certain structure) and helps static analysis know your true intent (much less likely to pass a random string than a whole User).

I would be in favor of doing a refactor of the current models. One "CodeIgniter-specific" issue this solves is our inconsistency around Entity primary keys - sometimes int, sometimes autoincrement (natural), sometimes string (UUID or other).

MGatner:
Changes classes to prefer whole-object parameters instead of the less-precise `int|string` ID.

Fixes #66.

### DIFF
--- a/src/Authentication/Actions/Email2FA.php
+++ b/src/Authentication/Actions/Email2FA.php
@@ -63,7 +63,7 @@ class Email2FA implements ActionInterface
         /** @var UserIdentityModel $identityModel */
         $identityModel = model(UserIdentityModel::class);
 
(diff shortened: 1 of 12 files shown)


kenjis (approved):
LGTM!

lonnieezell (approved):

Issue #138, opened 2022-05-17
Table `auth_activation_attempts` is not used


kenjis:
We have `EmailActivator`, so we should use the table for it?

But the table has the following:
```php
        [
            'id'         => ['type' => 'int', 'constraint' => 11, 'unsigned' => true, 'auto_increment' => true],
            'ip_address' => ['type' => 'varchar', 'constraint' => 255],
            'user_agent' => ['type' => 'varchar', 'constraint' => 255],
            'token'      => ['type' => 'varchar', 'constraint' => 255, 'null' => true],
            'created_at' => ['type' => 'datetime', 'null' => false],
        ]
```

On the other hand, `auth_logins` has the following:
```php
        [
            'id'         => ['type' => 'int', 'constraint' => 11, 'unsigned' => true, 'auto_increment' => true],
            'ip_address' => ['type' => 'varchar', 'constraint' => 255, 'null' => true],
            'user_agent' => ['type' => 'varchar', 'constraint' => 255, 'null' => true],
            'email'      => ['type' => 'varchar', 'constraint' => 255, 'null' => true],
            'user_id'    => ['type' => 'int', 'constraint' => 11, 'unsigned' => true, 'null' => true], // Only for successful logins
            'date'       => ['type' => 'datetime'],
            'success'    => ['type' => 'tinyint', 'constraint' => 1],
        ]
```

It seems `auth_activation_attempts` needs `user_id` and `success`.

MGatner:
Let's hold until #131.

lonnieezell:
That was a remnant from Myth:Auth's tables, I think. And definitely from before I switched to using identities, which is probably where it got dropped. 

I think it probably would be good to log those attempts, though, in order to trace attempted hacks if needed. Adding user_id and success is fine by me.

kenjis:
> I think it probably would be good to log those attempts, though, in order to trace attempted hacks if needed. Adding user_id and success is fine by me.

Yes. But for now I sent a PR to remove this table.

We should log not only activation attempts but also 2FA code post attempts.

PR #148, opened 2022-05-19
update migration file

kenjis:
- remove `auth_activation_attempts` (Ref #138)
- remove unique key for `users.email`
- fix the column name for key (Follow-up  #142)


lonnieezell (approved):
Looks like GitHub decided we shouldn't be able to authenticate again today, but the changes look good.

PR #149, opened 2022-05-19
CodeIgniter DevKit

MGatner:
* Switches dependencies to use the CodeIgniter DevKit
* Applies a few updated configurations, notably Rector
* Removes Infection from GitHub Workflows (still locally optional)
* Applies a few updated static analysis fixes

Note: Once Psalm is integrated in the DevKit (https://github.com/codeigniter4/devkit/pull/27) it can additionally be removed from **composer.json** here


lonnieezell:
Oh nice - thanks for removing infection from running during workflows. That makes me happy :)

MGatner:
@lonnieezell I knew it would ;) A few more fixes incoming.

kenjis (approved):

PR #151, opened 2022-05-19
Bugfix: Empty Identities

### LINKED ISSUE #150: Login/registration error after update
Hi,

I recently updated to the most recent version and when I try to login or register an account, I get the following error:

`mysqli_sql_exception #1064`
`You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ')' at line 4`.

When I check the database after registration, I see the records being inserted/updated. But the page dies almost immediately with the above error.

Where can I check ?

MGatner:
Addresses an issue where identity actions are not set resulting in a SQL query error.

Fixes #150

### DIFF
--- a/src/Models/UserIdentityModel.php
+++ b/src/Models/UserIdentityModel.php
@@ -115,6 +115,10 @@ class UserIdentityModel extends Model
      */
     public function getIdentitiesByTypes(User $user, array $types): array
     {
+        if ($types === []) {
+            return [];
+        }
+
         return $this->where('user_id', $user->getAuthId())
             ->whereIn('type', $types)
             ->findAll();


lonnieezell (approved):
This is a good safety measure to have in place but doesn't answer why an empty array is being sent there in the first place. Will merge this, though.

kenjis:
It comes from here:
https://github.com/codeigniter4/shield/blob/e92712645e19ca6ce9db54bc4065d51fbebdb7f2/src/Authentication/Authenticators/Session.php#L289-L292

lonnieezell:
Oh, I just realized that the actions are both null by default, which would cause that error. Ok, I feel better. Was concerned something bigger might have been happening.

PR #152, opened 2022-05-20
Ensures email/password/password_hash are saved when user is saved.

### LINKED ISSUE #103: Password should be rehashed on saving user, if present.
At current if someone sets an `email`, `password` or `password_hash` field on a `User` entity, it will not get saved when the entity is saved. This can be fixed using model events.

lonnieezell:
Fixes #103

### DIFF
--- a/src/Entities/User.php
+++ b/src/Entities/User.php
@@ -57,6 +57,10 @@ class User extends Entity
         return count($identities) ? array_shift($identities) : null;
     }
 
+    /**
+     * ensures that all of the user's identities are loaded
+     * into the instance for faster access later.
+     */
     private function populateIdentities(): void
     {
         if ($this->identities === null) {
@@ -125,6 +129,51 @@ class User extends Entity
     }
 
     /**
+     * If $user, $password, or $password_hash have been updated,
+     * will update the user's email identity record with the
+     * correct values.
+     */
+    public function saveEmailIdentity(): bool
+    {
+        if (empty($this->email) && empty($this->password) && empty($this->password_hash)) {
+            return true;
+        }
+
+        $identity = $this->getEmailIdentity();
+        if ($identity === null) {
+            // Ensure we reload all identities
+            $this->identities = null;
+
+            $this->createEmailIdentity([
+                'email'    => $this->email,
+                'password' => '',
+            ]);
+            $identity = $this->getEmailIdentity();
+        }
+
+        if (empty($identity)) {
+            return true;
+        }
+
+        if (! empty($this->email)) {
+            $identity->secret = $this->email;
+        }
+
+        if (! empty($this->password)) {
+            $identity->secret2 = service('passwords')->hash($this->password);
+        }
+
+        if (! empty($this->password_hash) && empty($this->password)) {
+            $identity->secret2 = $this->password_hash;
+        }
+
+        /** @var UserIdentityModel $identityModel */
+        $identityModel = model(UserIdentityModel::class);
+
+        return $identityModel->save($identity);
+    }
+
+    /**
      * Update the last used at date for an identity record.
      */
     public function touchIdentity(UserIdentity $identity): void
@@ -161,9 +210,18 @@ class User extends Entity
         return $this->password;
     }
 
-    public function setPassword(string $password): void
+    public function setPassword(string $password): User
     {
         $this->password = $password;
+
+        return $this;
+    }
+
+    public function setPasswordHash(string $hash): User
+    {
+        $this->password_hash = $hash;
+
+        return $this;
     }
 
     /**
--- a/src/Entities/User.php
+++ b/src/Entities/User.php
@@ -151,10 +151,6 @@ class User extends Entity
             $identity = $this->getEmailIdentity();
         }
 
-        if (empty($identity)) {
-            return true;
-        }
-
         if (! empty($this->email)) {
             $identity->secret = $this->email;
         }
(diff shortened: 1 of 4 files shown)


tests/Unit/UserTest.php
```diff
@@ -87,8 +94,70 @@ public function testLastLogin()
 
         $last = $this->user->lastLogin();
 
-        $this->assertInstanceOf(Login::class, $last); // @phpstan-ignore-line
+        $this->assertInstanceOf(Login::class, $last);
```
kenjis, line 97: Revert, please.

```
GitHub Actions / PHP 7.4 Static Analysis

tests/Unit/UserTest.php#L97

Call to method PHPUnit\Framework\Assert::assertInstanceOf() with 'CodeIgniter\\Shield\\Entities\\Login' and null will always evaluate to false.
```

src/Entities/User.php
```diff
@@ -124,6 +128,51 @@ public function getEmailIdentity(): ?UserIdentity
...
+            $this->createEmailIdentity([
+                'email'    => $this->email,
+                'password' => '',
+            ]);
+            $identity = $this->getEmailIdentity();
+        }
+
+        if (empty($identity)) {
+            return true;
+        }
```
kenjis, line 156: It seems `$identity` is never empty here. When is it empty?

src/Models/UserModel.php
```diff
@@ -208,4 +210,41 @@ private function checkQueryReturn(bool $return): void
...
     }
+
+    /**
+     * Override the BaseModel's `save()` method to allow
+     * updating of user email, password, or password_hash
+     * fields if they've been modified.
+     *
+     * @param mixed $data
+     */
+    public function save($data): bool
```
kenjis, line 221: Not `mixed` but `User`?

```suggestion
    public function save(User $data): bool
```

MGatner: If we specify `User` then we won't be able to save arrays. I would recommend the docblock be specified though: `User|array`

src/Models/UserModel.php
```diff
@@ -208,4 +210,41 @@ private function checkQueryReturn(bool $return): void
...
+    {
+        try {
+            $result = parent::save($data);
+
+            if ($result && $data instanceof User) {
+                /** @var User $user */
+                $user = $data->id === null
```
MGatner, line 228: We use the method everywhere else, might as well be consistent. 
```suggestion
                $user = $data->getAuthId() === null
```

tests/Unit/UserTest.php
```diff
@@ -23,6 +23,13 @@ final class UserTest extends TestCase
     protected $namespace;
     protected $refresh = true;
 
+    protected function tearDown(): void
+    {
+        parent::tearDown();
+
+        db_connect()->table('auth_identities')->truncate();
```
MGatner, line 30: Was this not already handled by `$refresh = true`?

tests/Unit/UserTest.php
MGatner: Same as the other PR 😂 I think Lonnie has an outdated local installation of PHPStan

MGatner (changes requested):
I'm curious your decision against going with Model events? This seems to do it but I thought that was an elegant approach.

tests/Unit/UserTest.php
lonnieezell: Damn you Stan. :)

src/Models/UserModel.php
lonnieezell: But we're talking about getting rid the `getAuthX()` methods, so why use it more? But I'll change for now.

lonnieezell:
> I'm curious your decision against going with Model events? This seems to do it but I thought that was an elegant approach.

Because CodeIgniter's events don't pass along the original object so we don't have access the variables, unfortunately. Something we should probably change at some point.

tests/Unit/UserTest.php
lonnieezell: It can probably be removed again. I was hitting weird errors during tests and tried that to make sure, but I think I fixed it a different way.

src/Models/UserModel.php
lonnieezell: Then it also complains about not being compatible with the method that we're overriding.

src/Entities/User.php
lonnieezell: Good catch. changed.

src/Models/UserModel.php
lonnieezell: Actually - turns out we can't use that because of the assert call. When creating a `new User()` and attempting to save it, there is no id, but the assertion yells anyway. Reverting.

src/Models/UserModel.php
```diff
@@ -208,4 +210,41 @@ private function checkQueryReturn(bool $return): void
             throw new RuntimeException($message, $error['code']);
         }
     }
+
+    /**
+     * Override the BaseModel's `save()` method to allow
+     * updating of user email, password, or password_hash
+     * fields if they've been modified.
+     *
+     * @param mixed $data
```
kenjis, line 219:
```suggestion
     * @param User|array $data
```

MGatner:
Looking good! My only qualm is that this is entirely tied to `save()`. I almost never use this method myself, preferring to call updates more specifically. Theoretically this shouldn't really be a problem since this is mostly for internal handling, but I could see it being a "gotcha" down the road.

MGatner (approved):

kenjis (approved):

lonnieezell:
That's a good point. Definitely worth looking into.

PR #154, opened 2022-05-20
fix: token unique constraint error

### LINKED ISSUE #143: Need to ensure Email2FA/EmailActivator token is unique
Otherwise it will definitely have collisions at some point.

See https://github.com/codeigniter4/shield/pull/129#issuecomment-1126880258

kenjis:
Fixes #143 

- if UNIQUE constraint error occurs, repeat insertion max 5 times.

### DIFF
--- a/src/Authentication/Actions/CreateIdentityTrait.php
+++ b/src/Authentication/Actions/CreateIdentityTrait.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace CodeIgniter\Shield\Authentication\Actions;
+
+use CodeIgniter\Shield\Entities\User;
+use CodeIgniter\Shield\Models\DatabaseException;
+use CodeIgniter\Shield\Models\UserIdentityModel;
+
+/**
+ * Create an identity for auth action
+ */
+trait CreateIdentityTrait
+{
+    /**
+     * Create an identity for auth action
+     */
+    private function createIdentity(User $user, string $name, string $extra): string
+    {
+        helper('text');
+
+        /** @var UserIdentityModel $userIdentityModel */
+        $userIdentityModel = model(UserIdentityModel::class);
+
+        // Delete any previous identities for action
+        $userIdentityModel->deleteIdentitiesByType($user, $this->type);
+
+        // Create an identity for the action
+        $maxTry = 5;
+        $data   = [
+            'user_id' => $user->getAuthId(),
+            'type'    => $this->type,
+            'name'    => $name,
+            'extra'   => $extra,
+        ];
+
+        while (true) {
+            $data['secret'] = $this->generateSecretCode();
+
+            try {
+                $userIdentityModel->create($data);
+
+                break;
+            } catch (DatabaseException $e) {
+                $maxTry--;
+
+                if ($maxTry === 0) {
+                    throw $e;
+                }
+            }
+        }
+
+        return $data['secret'];
+    }
+
+    private function generateSecretCode(): string
+    {
+        return random_string('nozero', 6);
+    }
+}
--- a/src/Authentication/Actions/CreateIdentityTrait.php
+++ b/src/Authentication/Actions/CreateIdentityTrait.php
@@ -12,9 +12,9 @@ use CodeIgniter\Shield\Models\UserIdentityModel;
 trait CreateIdentityTrait
 {
     /**
-     * Create an identity for auth action
+     * Create an identity with 6 digits code for auth action
      */
-    private function createIdentity(User $user, string $name, string $extra): string
+    private function createCodeIdentity(User $user, string $name, string $extra): string
     {
         helper('text');
 
--- a/src/Authentication/Actions/CreateIdentityTrait.php
+++ b/src/Authentication/Actions/CreateIdentityTrait.php
@@ -16,6 +16,8 @@ trait CreateIdentityTrait
      */
     private function createCodeIdentity(User $user, string $name, string $extra): string
     {
+        assert($user->id !== null);
+
         helper('text');
 
         /** @var UserIdentityModel $userIdentityModel */
@@ -27,7 +29,7 @@ trait CreateIdentityTrait
         // Create an identity for the action
         $maxTry = 5;
         $data   = [
-            'user_id' => $user->getAuthId(),
+            'user_id' => $user->id,
             'type'    => $this->type,
             'name'    => $name,
             'extra'   => $extra,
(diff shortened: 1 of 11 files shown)


MGatner (approved):
I hadn't noticed CheckQueryReturn before, seems like a bit of a hack 😬 Not really the issue at hand though, this PR fixes the uniqueness.

kenjis:
I don't know how do other people handle things like this, but it is not easy.

1. The CI DB driver behavior (Exceptions or Error values) changes by `DBDebug` value.
2. Exception type thrown by CI DB driver is not the same.
https://github.com/codeigniter4/CodeIgniter4/issues/6007

src/Authentication/Actions/CreateIdentityTrait.php
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Create an identity for auth action
+ */
+trait CreateIdentityTrait
+{
+    /**
+     * Create an identity for auth action
+     */
+    private function createIdentity(User $user, string $name, string $extra): string
```
lonnieezell, line 17: It seems this only works with actions that require a code and nothing else, as there's no way to pass either `secret` or `secret2`. I would suggest that either: 

a) the name and docblock are changed to something more specific
or
b) It accepts a $data array where the calling function can specify the bits, and then a flag that specifies whether or not to create a random string. 

My personal preference would be option b.

src/Authentication/Actions/CreateIdentityTrait.php
```diff
@@ -0,0 +1,59 @@
...
+        ];
+
+        while (true) {
+            $data['secret'] = $this->generateSecretCode();
+
+            try {
+                $userIdentityModel->create($data);
+
+                break;
+            } catch (DatabaseException $e) {
```
lonnieezell, line 43: Instead of trying to insert the record and catching an exception, would it be simpler to just check if an identity of that type with that code exists in the database? No need to worry about what types of exceptions are thrown then.

src/Authentication/Actions/CreateIdentityTrait.php
```diff
@@ -0,0 +1,59 @@
...
+namespace CodeIgniter\Shield\Authentication\Actions;
+
+use CodeIgniter\Shield\Entities\User;
+use CodeIgniter\Shield\Models\DatabaseException;
+use CodeIgniter\Shield\Models\UserIdentityModel;
+
+/**
+ * Create an identity for auth action
+ */
+trait CreateIdentityTrait
```
lonnieezell, line 12: Why create a new trait? The User entity already has several methods dealing with traits. Seems to me like it should just go there.

src/Authentication/Actions/CreateIdentityTrait.php
```diff
@@ -0,0 +1,59 @@
...
+        /** @var UserIdentityModel $userIdentityModel */
+        $userIdentityModel = model(UserIdentityModel::class);
+
+        // Delete any previous identities for action
+        $userIdentityModel->deleteIdentitiesByType($user, $this->type);
+
+        // Create an identity for the action
+        $maxTry = 5;
+        $data   = [
+            'user_id' => $user->getAuthId(),
```
lonnieezell, line 30:
```suggestion
            'user_id' => $user->id,
```

lonnieezell (changes requested):

src/Authentication/Actions/CreateIdentityTrait.php
kenjis: Even If we check that the code exists in the database, we still need to catch the exception.
Because it is a race condition.

Process A: generate code "123456"
Process B: generate code "123456"
Process A: read the database --> OK
Process B: read the database --> OK
Process A: insert --> OK
Process B: insert --> Error

Alternatively, the table needs to be locked before reading, but CI4 has no way to do that.
And locking does not seem a good way.

src/Authentication/Actions/CreateIdentityTrait.php
kenjis: Because I thought that the two methods in actions are the same knowledge.

Since these methods are only used by Actions, there is no need for `User` to have them.

src/Authentication/Actions/CreateIdentityTrait.php
kenjis: This method is only for the two action classes. I've changed the method name.

If we need another method to pass different arguments, we can consider adding more methods or commonizing them at that time.

src/Authentication/Actions/CreateIdentityTrait.php
kenjis: I thought `$user->getAuthId()` and `$user->id` are identical, so replacing all of them later would be okay.

But `$user->id` might be null, and `$user->getAuthId()` is never null.
In this case, it must not null.
Shouldn't `getAuthId()` be left as is?

lonnieezell: Now that there's no interface necessitating the `getAuthX` methods, I find them redundant and confusing, since there's nothing auth-specific to it. 

If anyone is calling this method they would expect the user id to be there. I would do a check at the beginning of the function and throw an exception if it's not there. 

I think they need to go away. And I figured it's easier to use what we know it's going to be when writing new code, especially if you need it not to be null.

src/Authentication/Actions/CreateIdentityTrait.php
lonnieezell: With the slight refactoring I mentioned earlier the method could be useful in more situations than the current one to all developers.

src/Authentication/Actions/CreateIdentityTrait.php
lonnieezell: Good point.

MGatner: I understand the race condition, but realistically what we are coding around is generating two random six-digits strings that are the same within milliseconds of each other. The theoretical scope of unlikelihood has to be on the magnitude of trillions. I'm fine with simplifying this to a database read to clear up the DBDebug stuff - we could add a comment in the code about race condition to remember that it was not overlooked.

src/Authentication/Actions/CreateIdentityTrait.php
kenjis: Done.

src/Authentication/Actions/CreateIdentityTrait.php
kenjis: I will remove this Trait in the next PR.

MGatner (approved):

PR #158, opened 2022-05-20
lang file for farsi language(fa-IR)

datamweb:
Hi,
Language file for Persian(fa-IR).


lonnieezell:
Thanks for this addition! I obviously don't speak the language, but I spot-check a dozen or so and they translate well. Merging.

PR #160, opened 2022-05-21
fix: add/improve lang messages

kenjis:
When localized and checked, some messages were missing or difficult to understand the meaning.

- add missing lang messages
- fix some lang messages

PR #164, opened 2022-05-23
fix: translate fa-IR 

datamweb:
Hello **engineers!**
The language file was modified for Persian.
(related to PR #160)

### DIFF
--- a/src/Language/fa/Auth.php
+++ b/src/Language/fa/Auth.php
@@ -1,5 +1,14 @@
 <?php
 
+/**
+ * This file is part of CodeIgniter 4 framework.
+ *
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view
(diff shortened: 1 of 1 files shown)


datamweb:
@kenjis , Considering that Persian is RTL.
Takes **PHPCSFixer** errors.
Is there a suggestion to fix this problem by keeping the comments?
Or should I delete English comments altogether?

kenjis:
@datamweb RTL and the English comments have nothing to do with.

>    1) codeigniter-shield/src/Language/fa/Auth.php (single_blank_line_at_eof)

You just forgot a line break at the end of the file.

Run cs-fix and you will get the fixed file.
```
$ composer cs-fix
```

kenjis (approved):

datamweb:
@kenjis Problem cs was resolved, but the translation was broken, so I deleted the English comets.

kenjis (dismissed):

kenjis (approved):
I've confirmed all items are translated.

PR #165, opened 2022-05-24
test: add Translations Tests

kenjis:
from https://github.com/codeigniter4/translations/tree/develop/tests

### DIFF
--- a/tests/Language/AbstractTranslationTestCase.php
+++ b/tests/Language/AbstractTranslationTestCase.php
@@ -1,0 +1,413 @@
+<?php
+
+/**
+ * This file is part of CodeIgniter 4 framework.
+ *
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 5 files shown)


tests/Language/AbstractTranslationTestCase.php
```diff
@@ -0,0 +1,413 @@
...
+            $count > 1 ? 'keys' : 'key',
+            implode('", "', $keysNotConfigured),
+            $locale,
+            $count > 1 ? 'are' : 'is'
+        ));
+    }
+
+    /**
+     * This tests that all included language keys in a language file for the current
+     * locale that have corresponding keys in the main CI4 repository are really translated
```
lonnieezell, line 214: Just to be a bit nit-picky, but all of the comments in the file says it tests that they match the CI4 repository, which isn't what's being tested here. Should specify the Shield repo.

lonnieezell (approved):

tests/Language/AbstractTranslationTestCase.php
kenjis: I think it it better that the file is in other repository like `codeigniter4/translation-tests`.
So we can share it.

PR #173, opened 2022-05-25
refactor: remove $_SESSION

kenjis:
`$session->get()` does not return all `$_SESSION` items (flashdata and tempdata are excluded), 
but I think this change is fine.

### DIFF
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -655,15 +655,17 @@ class Session implements AuthenticatorInterface
 
         // Destroy the session data - but ensure a session is still
         // available for flash messages, etc.
-        if (isset($_SESSION)) {
-            foreach (array_keys($_SESSION) as $key) {
-                $_SESSION[$key] = null;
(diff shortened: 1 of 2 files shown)


lonnieezell (approved):

lonnieezell:
I agree - flashdata and tempdata is fine since they will naturally expire very quickly.

src/Authentication/Authenticators/Session.php
```diff
@@ -655,15 +655,17 @@ public function logout(): bool
...
         // Destroy the session data - but ensure a session is still
         // available for flash messages, etc.
-        if (isset($_SESSION)) {
-            foreach (array_keys($_SESSION) as $key) {
-                $_SESSION[$key] = null;
-                unset($_SESSION[$key]);
+        /** @var \CodeIgniter\Session\Session $session */
+        $session     = session();
+        $sessionData = $session->get();
+        if (isset($sessionData)) {
```
MGatner, line 661: I believe `get()` without a parameter always returns an array - this should be unnecessary.

PR #176, opened 2022-05-26
feat: add `auth_token_logins` table

kenjis:
- add `auth_token_logins` table to record Bearer Token type login attempts
- use `auth_token_logins` in AccessTokens Authenticator
- fix `auth_token` column type to not null
- refactor


MGatner (approved):
It's a little odd to have duplicate table structure (thanks to PHPCPD pointing that out) but the alternative seems messy.

lonnieezell (approved):

PR #177, opened 2022-05-26
fix: lastLogin()

### LINKED ISSUE #155: User::lastLogin() does not return login data that logged in by username/password
https://github.com/codeigniter4/shield/blob/e92712645e19ca6ce9db54bc4065d51fbebdb7f2/src/Entities/User.php#L187-L200

kenjis:
Fixes #155

- returns the last successful login in the `auth_logins` table of the user
- refactor
- ~~When `$allowFailed` is false, returns the last successful login in the `auth_logins` table~~
- ~~When `$allowFailed` is true, finds the last login by email or username in the `auth_logins` table~~

### DIFF
--- a/src/Entities/User.php
+++ b/src/Entities/User.php
@@ -226,11 +226,11 @@ class User extends Entity
     /**
      * Returns the last login information for this user as
      */
-    public function lastLogin(bool $allowFailed = false): ?Login
+    public function lastLogin(): ?Login
     {
         /** @var LoginModel $logins */
         $logins = model(LoginModel::class);
 
-        return $logins->lastLogin($this, $allowFailed);
+        return $logins->lastLogin($this);
     }
 }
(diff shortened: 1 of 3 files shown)


MGatner:
I just responded on the other thread... I'm still in favor of splitting these out. How about `lastLogin()` and `lastAttempt()`/`lastFailure()`? I think it is clearer that way, and I honestly don't see when a developer would use `$user->lastLogin(false)`.

kenjis:
Yes, I also think we don't need the "failed" flag in the method.

lonnieezell:
I'm not sure what the use case I was originally thinking of might be for pulling the last failed attempt, honestly. I did a quick search of both Shield and Bonfire code and didn't find it used. I'm fine with removing that feature, unless the user in the other thread has a good idea of it's use.

kenjis:
I removed the param.

The param will be needed only when we want to find the last successful or failed login of a user.
but i seems no use case like that and it is not easy to know who belongs a failed login.

lonnie-vault (approved):

MGatner (approved):

PR #181, opened 2022-05-27
feat: add `id_type` in logins table

kenjis:
~~Need to rebase after merging  #177~~

Identifiers can be added, and it will be not possible to determine what they are just from the identifier string.
Currently, `token: ` is also added to access tokens. 
Also, if there is no constraint, it is not possible to tell whether it is an `email` or a `username` just from the string.

It is easier to know what it is if we add a column for identifier type.

- add `id_type` in `auth_logins` and `auth_token_logins`
  - type: `email_password`, `username`, `magic-link`, and `access_token`
- add validation error check in `CheckQueryReturnTrait`
- add const `ID_TYPE_*` in `Session` and `AccessTokens`

### DIFF
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -239,7 +239,7 @@ class Session implements AuthenticatorInterface
         string $userAgent,
         $userId = null
     ): void {
-        $idType = ! isset($credentials['email']) && isset($credentials['username'])
+        $idType = (! isset($credentials['email']) && isset($credentials['username']))
             ? 'username'
             : 'email';
 
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -239,7 +239,13 @@ class Session implements AuthenticatorInterface
         string $userAgent,
         $userId = null
     ): void {
+        $idType = 'email';
+        if (! isset($credentials['email']) && isset($credentials['username'])) {
+            $idType = 'username';
+        }
+
         $this->loginModel->recordLoginAttempt(
+            $idType,
             $credentials['email'] ?? $credentials['username'],
             $success,
             $ipAddress,
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -25,10 +25,20 @@ use stdClass;
 
 class Session implements AuthenticatorInterface
 {
-    private const STATE_UNKNOWN   = 0;
-    private const STATE_ANONYMOUS = 1;
-    private const STATE_PENDING   = 2;
-    private const STATE_LOGGED_IN = 3;
+    /**
+     * @var string Special ID Type.
+     *             `username` is stored in `users` table, so no `auth_identities` record.
+     */
+    public const ID_TYPE_USERNAME = 'username';
+
+    public const ID_TYPE_EMAIL_PASSWORD = 'email_password';
+    public const ID_TYPE_MAGIC_LINK     = 'magic-link';
+    public const ID_TYPE_EMAIL_2FA      = 'email_2fa';
+    public const ID_TYPE_EMAIL_ACTIVATE = 'email_activate';
+    private const STATE_UNKNOWN         = 0;
+    private const STATE_ANONYMOUS       = 1;
+    private const STATE_PENDING         = 2;
+    private const STATE_LOGGED_IN       = 3;
 
     /**
      * The persistence engine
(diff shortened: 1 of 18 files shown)


lonnieezell (approved):

src/Authentication/Authenticators/Session.php
```diff
@@ -239,7 +239,13 @@ private function recordLoginAttempt(
         string $userAgent,
         $userId = null
     ): void {
+        $idType = 'email';
+        if (! isset($credentials['email']) && isset($credentials['username'])) {
+            $idType = 'username';
+        }
```
MGatner, line 245:
```suggestion
        $idType = ! isset($credentials['email']) && isset($credentials['username']))
            ? 'username'
            : 'email';
```

src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
```diff
@@ -57,6 +57,7 @@ public function up(): void
             'id'         => ['type' => 'int', 'constraint' => 11, 'unsigned' => true, 'auto_increment' => true],
             'ip_address' => ['type' => 'varchar', 'constraint' => 255],
             'user_agent' => ['type' => 'varchar', 'constraint' => 255, 'null' => true],
+            'id_type'    => ['type' => 'varchar', 'constraint' => 255],
```
MGatner, line 60: This should be added to the `identifier` key since reverse lookup will always require the ID type now too.

MGatner (changes requested):
I like this better than the prefixes. A few suggestions.

src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
kenjis: When and why do we need reverse lookup?

Table `auth_identities` already has `type`: `email_password`, `magic-link`, `access_token`
Use this as `id_type`?

kenjis: Changed to use `auth_identities` type.

Replaced strings with constants In the source code. https://github.com/codeigniter4/shield/pull/181/commits/2464affbf2466ed50ba24ee895364c4a11ae90c7

MGatner: I'm not sure where the reverse lookup happens but we already have the index for it:
```php
$this->forge->addKey('identifier');
```
Since this is not a unique field we theoretically could have two of the same identifiers for two different ID types, so the type should always be used with the identifier for reverse lookup - which makes the current index insufficient.

src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
```diff
@@ -76,6 +77,7 @@ public function up(): void
...
```
MGatner, line 80: Same thing for here.

src/Models/LoginModel.php
```diff
@@ -28,7 +29,8 @@ class LoginModel extends Model
     protected $useTimestamps   = false;
     protected $validationRules = [
         'ip_address' => 'required',
-        'identifier' => 'required',
+        'id_type'    => 'required',
+        'identifier' => 'permit_empty|string',
```
MGatner, line 33: When can this be empty?

src/Models/LoginModel.php
```diff
@@ -78,6 +87,7 @@ public function fake(Generator &$faker): Login
     {
         return new Login([
             'ip_address' => $faker->ipv4,
+            'id_type'    => 'email',
```
MGatner, line 90:
```suggestion
            'id_type'    => Session:: ID_TYPE_EMAIL_PASSWORD,
```

MGatner (changes requested):

src/Models/LoginModel.php
kenjis: When failed login, it may be empty.

MGatner: Would that even get as far as a login attempt? I would think basic validation errors (like entering `1234` in an email field) should not be recorded as "attempts", especially since they won't ever link to an actual user. In this case I would be in favor of always requiring the identifier field.

kenjis: It is a failed attempt. I don't know why it should not be recorded.
Failed attempts can't link to an actual user.

src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
kenjis: I got what you mean!

Done.

src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
kenjis: Done.

src/Models/LoginModel.php
kenjis: Good catch! Thanks.

Done.

src/Models/LoginModel.php
MGatner: I won't push it.

MGatner (approved):
Looks good!

PR #182, opened 2022-05-27
fix: Models return type

kenjis:
The return type of write-type query methods are not unified. 
Failures of writing to database are abnormal things, so Exceptions should be thrown.

- change return type to `void`
- add query return value checks

> If you forget to check an error, you’ll get silent bugs. Unchecked error values are worse than uncaught exceptions.
https://programmingduck.com/articles/exceptions-error-values

### DIFF
--- a/src/Models/UserModel.php
+++ b/src/Models/UserModel.php
@@ -195,6 +195,8 @@ class UserModel extends Model
      * fields if they've been modified.
      *
      * @param array|User $data
+     *
+     * @TODO can't change the return type to void.
      */
     public function save($data): bool
     {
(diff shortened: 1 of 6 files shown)


lonnieezell (approved):

src/Models/UserModel.php
```diff
@@ -195,6 +195,8 @@ public function activate(User $user): void
      * fields if they've been modified.
      *
      * @param array|User $data
+     *
+     * @TODO can't change the return type to void.
```
MGatner, line 199: I still think we should find a way to handle this with model callback events so we can catch `update()` etc. That would solve the return typing.

MGatner (approved):
IMO this is a deficiency of the framework that we are always working around. Being in the Exceptions camp I am definitely in favor of this 🤗 but glad to see that Lonnie approves as well.

PR #184, opened 2022-05-27
Provide a rate limiting filter.

lonnieezell:
Provides a default rate-limiting controller filter, auto-registers it through the Registrar, and has docs with suggested way of implementing it.


docs/2 - authentication.md
```diff
@@ -275,20 +275,37 @@ if ($user->tokenCant('forums.manage')) {
...
+
+### Rate Limiting
+
+To help protect your authentication forms from being spammed by bots, it is recommended that you use
+the `auth-limit` filter on all of your authentication routes. This can be done with the following
+filter setup:
+
+```php
+public $filters = [
+    'auth-rates' => [
```
kenjis, line 305: `'auth-limit'`?

src/Config/Registrar.php
```diff
@@ -16,9 +17,10 @@ public static function Filters(): array
     {
         return [
             'aliases' => [
-                'session' => SessionAuth::class,
-                'tokens'  => TokenAuth::class,
-                'chain'   => ChainAuth::class,
+                'session'    => SessionAuth::class,
+                'tokens'     => TokenAuth::class,
+                'chain'      => ChainAuth::class,
+                'auth-limit' => AuthRates::class,
```
kenjis, line 23: Why is the classname `AuthRates`, not `AuthLimit`?

src/Filters/AuthRates.php
```diff
@@ -0,0 +1,57 @@
...
+use CodeIgniter\HTTP\RequestInterface;
+use CodeIgniter\HTTP\Response;
+use CodeIgniter\HTTP\ResponseInterface;
+use CodeIgniter\Shield\Auth;
+
+/**
+ * Chain Authentication Filter.
+ *
+ * Checks all authentication systems specified within
+ * `Config\Auth->authenticationChain`
```
kenjis, line 16: Not updated.

src/Filters/AuthRates.php
```diff
@@ -0,0 +1,57 @@
...
+ */
+class AuthRates implements FilterInterface
+{
+    /**
+     * Intened for use on auth form pages to restrict the number
+     * of attempts that can be generated. Restricts it to 10 attempts
+     * per minute, which is what auth0 uses.
+     *
+     * @see https://auth0.com/docs/troubleshoot/customer-support/operational-policies/rate-limit-policy/database-connections-rate-limits
```
kenjis, line 25: Auth0 does not matter.

kenjis: Do you mean the implementation of this filter referred to Auth0?

> If one IP address makes 20 login attempts in one minute to the same user account, the rate limit comes into effect. After that, Auth0 allows the user 10 attempts per minute. Any combination of successful and failed login attempts count toward this limit. 
https://auth0.com/docs/troubleshoot/customer-support/operational-policies/rate-limit-policy/database-connections-rate-limits

src/Config/Registrar.php
lonnieezell: Better question - why is it `auth-limit` when the class is named `AuthRates` :)  I started it as auth limit then changed my mind and missed that. Fixed.

src/Filters/AuthRates.php
lonnieezell: Done

src/Filters/AuthRates.php
lonnieezell: Yes, that's why I linked the article. For us to see where I chose the rate limiting amount from. And to be able to go back and reference their other pages about how they protect against other attacks as inspiration. 

I understand that the company Auth0 doesn't matter for what we're building here, but as a large company with lots of data around the subject, they seemed like a good source of inspiration for how much to limit.

docs/2 - authentication.md
```diff
@@ -275,20 +275,37 @@ if ($user->tokenCant('forums.manage')) {
...
+    'auth-rates' => \CodeIgniter\Shield\Filters\AuthRates::class,
 ];
 ```
 
 These can be used in any of the normal filter config settings, or within the routes file.
+
+### Rate Limiting
+
+To help protect your authentication forms from being spammed by bots, it is recommended that you use
+the `auth-limit` filter on all of your authentication routes. This can be done with the following
```
kenjis, line 300:
```suggestion
the `auth-rates` filter on all of your authentication routes. This can be done with the following
```

src/Config/Registrar.php
kenjis, line 23:
```suggestion
                'auth-rates' => AuthRates::class,
```

src/Filters/AuthRates.php
kenjis: Okay, I got it. That makes sense.

src/Language/fa/Auth.php
```diff
@@ -24,6 +24,7 @@
     'oldToken'             => 'توکن دسترسی منقضی شده است.',
     'noUserEntity'         => 'برای اعتبار سنجی هویت کاربر بایستی رمز عبور ارائه شود',
     'invalidEmail'         => 'امکان تایید ایمیلی که با آدرس ایمیل ثبت شده یکسان نیست، وجود ندارد.',
+    'throttled'            => 'درخواست های بسیار زیادی از این آدرس IP انجام شده است. می توانید در {0} ثانیه دوباره امتحان کنید.',
```
datamweb, line 27: 'throttled'            => 'درخواست های بسیار زیادی از این آدرس IP انجام شده است. می توانید بعد از {0} ثانیه دوباره امتحان کنید.',

datamweb (changes requested):

MGatner (changes requested):
@kenjis Did a very good review, I won't go through all of it. My main question is if we should probably be enabling this by default? We can use **src/Config/Filters.php** to enable the filter but then the problem becomes how to turn it off - a Config toggle or list of exempt routes perhaps?

Semi-related, I think the **2 - authentication.md** docs have gotten too big and I worry the wall of text is going to make it less likely for developers to notice things like "oh I should turn on this rate filter". Maybe splitting the docs out and highlighting these setup steps in a quick start?

datamweb:
> Maybe splitting the docs out and highlighting these setup steps in a quick start?

Something a person whose native language is not English and is interested in. 👍

src/Filters/AuthRates.php
```diff
@@ -0,0 +1,56 @@
...
+     * @return RedirectResponse|void
+     */
+    public function before(RequestInterface $request, $arguments = null)
+    {
+        $throttler = service('throttler');
+
+        // Restrict an IP address to no more than 10 requests
+        // per minute on any auth-form pages (login, register, forgot, etc).
+        if ($throttler->check(md5($request->getIPAddress()), 10, MINUTE, 1) === false) {
+            return service('response')->setStatusCode(
```
datamweb, line 37: Friends, can you create an event here?
For when the amount of after for a bucket to completely fill. We can send a message to the user (email, SMS, etc.)
Or do you suggest a solution to this issue please?
`Events::trigger('afterAuthRatesBucketFill', $user);`

kenjis:
>  the problem becomes how to turn it off - a Config toggle or list of exempt routes perhaps?

Once force to enable it and turn it off? It's not efficient. I don't like it.

I hope improvement of the install guide.
Or to provide a command to update `app/Config/Filters.php` to install it?

datamweb:
> > the problem becomes how to turn it off - a Config toggle or list of exempt routes perhaps?
> 
> Once force to enable it and turn it off? It's not efficient. I don't like it.
> 
> I hope improvement of the install guide. Or to provide a command to update `app/Config/Filters.php` to install it?

I don't know if this can be helpful in this section.
I have already faced such a challenge.
And I developed a command with this title.

Screenshot 2022-05-28 173339

I believe that **CI** it should provide a CLI command that all packages can execute with what they need to install.
See the command to understand the output description.

output

kenjis:
@datamweb Awesome! It is what we need.

src/Config/Registrar.php
lonnieezell: Done.

docs/2 - authentication.md
lonnieezell: done

src/Config/Registrar.php
lonnieezell: Done.

src/Language/fa/Auth.php
lonnieezell: Done.

lonnieezell:
> My main question is if we should probably be enabling this by default? We can use src/Config/Filters.php to enable the filter but then the problem becomes how to turn it off - a Config toggle or list of exempt routes perhaps?

I'm personally leery of making this more complex than it needs to be. The simplest solution is simply having them take 30 seconds to add it themselves, instead of us building layers of complexity into something. 

> Semi-related, I think the 2 - authentication.md docs have gotten too big and I worry the wall of text is going to make it less likely for developers to notice things like "oh I should turn on this rate filter". Maybe splitting the docs out and highlighting these setup steps in a quick start?

I split it out to the install docs page. Good call.

src/Filters/AuthRates.php
lonnieezell: The response already includes a message telling them how long is left until they can try again, which can be inspected on the front end and displayed to the user. I'm not convinced this has any benefit, honestly. It seems like the `failedLogin` event is already giving enough information outside of the time limit.

lonnieezell:
@datamweb It sounds like a lot of what that script is doing can be done with [Registrars](https://codeigniter.com/user_guide/general/configuration.html) and the [Publisher](https://codeigniter.com/user_guide/libraries/publisher.html). Are you familiar with these? Does your script do more?

src/Filters/AuthRates.php
kenjis: @datamweb You can create your own filter based on this.

kenjis (approved):

src/Filters/AuthRates.php
datamweb: Okay, I got it.

datamweb:
@lonnieezell am not familiar with what has been said. So I do not know if they work the same with this script or not. This script edits the necessary files and adds the necessary items to the files: filter, generator, validation, etc. I're looking forward to the initial release.

MGatner (approved):

PR #185, opened 2022-05-29
feat: add debug toolbar

kenjis:
- ported from Myth:Auth

Screenshot 2022-05-29 21 19 02

### DIFF
--- a/src/Collectors/Auth.php
+++ b/src/Collectors/Auth.php
@@ -3,6 +3,7 @@
 namespace CodeIgniter\Shield\Collectors;
 
 use CodeIgniter\Debug\Toolbar\Collectors\BaseCollector;
+use CodeIgniter\Shield\Auth as ShieldAuth;
 
 /**
  * Debug Toolbar Collector for Auth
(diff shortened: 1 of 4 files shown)


src/Collectors/Auth.php
```diff
@@ -0,0 +1,104 @@
...
+    protected $title = 'Auth';
+
+    //--------------------------------------------------------------------
+
+    /**
+     * Returns any information that should be shown next to the title.
+     */
+    public function getTitleDetails(): string
+    {
+        return get_class(service('auth'));
```
MGatner, line 51: Would the class of the actual authenticator used be more helpful?

MGatner (changes requested):

src/Collectors/Auth.php
kenjis: Nice!

Done.

MGatner (approved):

kenjis:
Rebased.

MGatner (approved):

PR #194, opened 2022-06-01
test: add folders and move test files

kenjis:
Too many files in a folder.


tests/Authentication/Authenticators/AccessTokenAuthenticatorTest.php
```diff
@@ -28,7 +28,7 @@ protected function setUp(): void
 
         $config = new Auth();
         $auth   = new Authentication($config);
-        $auth->setProvider(model(UserModel::class)); // @phpstan-ignore-line
+        $auth->setProvider(\model(UserModel::class)); // @phpstan-ignore-line
```
MGatner, line 31: Our style guide decided against slashing root functions. Did you add these, or did they slip in via CS Fixer/Rector somehow?

MGatner (changes requested):

MGatner (changes requested):
I'm good with the folder move but please revert the function slashing. If you have a compelling case for them we should discuss it for the style guide first.

tests/Authentication/Authenticators/AccessTokenAuthenticatorTest.php
kenjis: It seems PhpStorm did it when refactoring.

kenjis: Removed.

MGatner (approved):

PR #195, opened 2022-06-01
feat: JWT Authenticator

kenjis:
~~Needs #703~~
~~Needs to rebase after merging #194, #199~~

Add:
- `Config\AuthJWT`
- `Authentication\Authenticators\JWT`
- `Filters\JWTAuth`
- `Authentication\JWTManager` = `service((jwtmanager)`

How to Test/Use:
- https://github.com/codeigniter4/shield/pull/195#issuecomment-1374456938
- https://github.com/kenjis/codeigniter-shield/blob/feat-jwt/docs/addons/jwt.md#jwt-authentication

Sample Test App:
- https://github.com/kenjis/ci4-shield-test/tree/test-jwt#how-to-test-json-web-token-jwt-authentication
- https://github.com/kenjis/ci4-shield-test/commit/0ef2d5d64afaab49576765d28560f34c31d8c4dc

TODO:
- [x] login recording specification
- [x] update docs


src/Authentication/Authenticators/JWT/Firebase.php
```diff
@@ -0,0 +1,41 @@
...
+        $key       = $config['secretKey'];
+        $algorithm = $config['algorithm'];
+
+        try {
+            return JWT::decode($encodedToken, new Key($key, $algorithm));
+        } catch (BeforeValidException|ExpiredException $e) {
+            throw new RuntimeException('Expired JWT: ' . $e->getMessage(), 0, $e);
+        } catch (
+            InvalidArgumentException|DomainException|UnexpectedValueException
+            |SignatureInvalidException $e
```
MGatner, line 36: I don't think these are valid syntax for 7.4, right? Could do one broader catch or cascading statements.

src/Config/Auth.php
```diff
@@ -185,6 +188,28 @@ class Auth extends BaseConfig
...
+     *               E.g., $ php -r 'echo base64_encode(random_bytes(32));'
+     * - algorithm   JWT Signing Algorithms.
+     * - timeToLive  Specifies the amount of time, in seconds, that a token is valid.
+     *
+     * @var array<string, bool|int|string>
+     */
+    public array $jwtConfig = [
+        'issuer'     => '<Issuer of the JWT>',
+        'audience'   => '<Audience of the JWT>',
+        'secretKey'  => '<Set secret randmon string like MQ4GfWut1OYZxPY9fXAIq2YP6KzTSKOGNS7dJNcRrR8=>',
```
MGatner, line 208:
```suggestion
        'secretKey'  => '<Set secret random string like MQ4GfWut1OYZxPY9fXAIq2YP6KzTSKOGNS7dJNcRrR8=>',
```

MGatner (changes requested):
This is looking awesome! I had no idea you were working on this, I would have guessed version 1.1 or 1.2.

src/Authentication/Authenticators/JWT/Firebase.php
kenjis: > As of PHP 7.1.0, a [catch](https://www.php.net/manual/en/language.exceptions.php#language.exceptions.catch) block may specify multiple exceptions using the pipe (|) character. This is useful for when different exceptions from different class hierarchies are handled the same.
https://www.php.net/manual/en/language.exceptions.php#language.exceptions.catch

src/Authentication/Authenticators/JWT.php
```diff
@@ -0,0 +1,240 @@
...
+/**
+ * Stateless JWT Authenticator
+ */
+class JWT implements AuthenticatorInterface
+{
+    /**
+     * @var string Special ID Type.
+     *             This Authenticator is stateless, so no `auth_identities` record.
+     */
+    public const ID_TYPE_JWT = 'jwt';
```
lonnieezell, line 28: Will this ever change? With only one possible value doesn't seem necessary to use a constant.

src/Authentication/Authenticators/JWT.php
```diff
@@ -0,0 +1,240 @@
...
+    {
+        /** @var IncomingRequest $request */
+        $request = service('request');
+
+        $ipAddress = $request->getIPAddress();
+        $userAgent = $request->getUserAgent();
+
+        $result = $this->check($credentials);
+
+        if (! $result->isOK()) {
```
lonnieezell, line 64: With JWT being stateless, this is going to record a login attempt for every API request. And it looks like it's recording it in the main login table. I think this should be handled like we handled Access Tokens, which at the very least used a different table. 

Now that I'm thinking through this more, I don't know that access tokens or JWT really need to be tracked every page view anyway. Tracking login attempts is more for catching bad actors attempting to break into accounts. With PAT and JWT, I don't believe you typically see these same types of attacks, so rapidly expanding a database to very large sizes doesn't seem beneficial.

src/Authentication/Authenticators/JWT.php
```diff
@@ -0,0 +1,240 @@
...
+            );
+
+            return $result;
+        }
+
+        $user = $result->extraInfo();
+
+        $this->login($user);
+
+        $this->loginModel->recordLoginAttempt(
```
lonnieezell, line 81: Same comment as above.

src/Authentication/Authenticators/JWT/Firebase.php
```diff
@@ -0,0 +1,41 @@
...
+use Firebase\JWT\BeforeValidException;
+use Firebase\JWT\ExpiredException;
+use Firebase\JWT\JWT;
+use Firebase\JWT\Key;
+use Firebase\JWT\SignatureInvalidException;
+use InvalidArgumentException;
+use stdClass;
+use UnexpectedValueException;
+
+class Firebase implements JWTDecoderInterface
```
lonnieezell, line 16: I'm not a fan of naming this `Firebase` since that is a product name and this authenticator doesn't have anything to do with Firebase itself, other than the library was provided by them. I'd rename to something like more descriptive like simply `JWT` or similar.

src/Authentication/TokenGenerator/JWT/JWTGeneratorInterface.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace CodeIgniter\Shield\Authentication\TokenGenerator\JWT;
+
+interface JWTGeneratorInterface
```
lonnieezell, line 5: Like the DecoderInterface, it doesn't appear that we're setting things up to swap out JWT implementations so I don't think an interface is necessary. Making it possible to swap out implementations seems like an unnecessary complication for very minimal gain.

src/Config/Auth.php
```diff
@@ -185,6 +188,28 @@ class Auth extends BaseConfig
...
+     * - secretKey   The secret key. Needs more than 256 bits random string.
+     *               E.g., $ php -r 'echo base64_encode(random_bytes(32));'
+     * - algorithm   JWT Signing Algorithms.
+     * - timeToLive  Specifies the amount of time, in seconds, that a token is valid.
+     *
+     * @var array<string, bool|int|string>
+     */
+    public array $jwtConfig = [
+        'issuer'     => '<Issuer of the JWT>',
+        'audience'   => '<Audience of the JWT>',
```
lonnieezell, line 207: According to the chart of supported PHP libs on the JWT site, Firebase's implementation doesn't support audience checks. If that's the case do we need a setting for it?

lonnieezell (changes requested):
It's great to see a JWT implementation coming! Thanks for this. 

However, I do have a couple of concerns with the current implementation. 

1. Choice of implementation. Why choose Firebase's implementation? I think the best reason is probably that it's backed by Google, but according to the the JWT site, it doesn't provide the most complete implementation. That might not be all bad, though. Honestly, I kind of had it in mind that if we ever included JWT we'd roll our own since the implementation of a JWT is pretty straight-forward. Not sure if that's worth considering to remove the reliance on a third-party libs changes, etc. 

2. I think you may have over-architected it. I've commented on a few classes along the way, but in general I think we should strive for as simple of an architecture as the library allows. 

This would also require updating all docs and adding new docs. 

I don't think we'll be able to get this one in before the initial release.

kenjis:
> I don't think we'll be able to get this one in before the initial release.

There are some parts of this feature that require consideration of specifications.
I think it will take some time.

src/Authentication/Authenticators/JWT.php
kenjis: > Will this ever change?

It might change. `jwt-jws` may be accurate.
Or just `jws` is better?

It is used twice in the file, and used twice in `JWTAuthenticatorTest`.
If we use string `jwt`, IDE can't find where it is used easily.

> With only one possible value doesn't seem necessary to use a constant.

For example, if a dev want to change the value, it is easier when a constant is used,
even if it is only one possible value. 

And another value would be added in the future.

kenjis:
> Why choose Firebase's implementation?

I googled [CodeIgniter4 jwt](https://www.google.com/search?channel=crow5&client=firefox-b-d&q=CodeIgniter4+jwt), and found most tutorials use Firebase implementation.
And it is commonly used in my country. So first of all, I chose it.

But I know it doesn't provide the most complete implementation, so I made it replaceable.

datamweb:
It is also commonly used in my country.

src/Authentication/Authenticators/JWT/JWTDecoderInterface.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace CodeIgniter\Shield\Authentication\Authenticators\JWT;
+
+use stdClass;
+
+interface JWTDecoderInterface
```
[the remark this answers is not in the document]
MGatner: I like the interface in that it allows the JWT Authenticator to exist without the Firebase dependency, because we can pass around the interface name instead of an actual class.

I think the alternative would be to release JWT as a separate "extension" which gets more complex to use and maintain.

src/Config/Auth.php
MGatner: I'm not familiar with "audience", probably a symptom of only ever using Firebase's implementation. 😂

kenjis: It is a Registered Claim.
> [Registered claims](https://tools.ietf.org/html/rfc7519#section-4.1): These are a set of predefined claims which are not mandatory but recommended, to provide a set of useful, interoperable claims. Some of them are: iss (issuer), exp (expiration time), sub (subject), aud (audience), and [others](https://tools.ietf.org/html/rfc7519#section-4.1).
https://jwt.io/introduction

I'm going to change the property so that users can remove it.

```php
    public array $jwtConfig = [
        'payload' => [
            'iss' => '<Issuer of the JWT>',
            'aud' => '<Audience of the JWT>',
        ],
        // ...
    ];
```

src/Authentication/TokenGenerator/JWTGenerator.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+namespace CodeIgniter\Shield\Authentication\TokenGenerator;
+
+use CodeIgniter\I18n\Time;
+use CodeIgniter\Shield\Authentication\TokenGenerator\JWT\Firebase;
+use CodeIgniter\Shield\Authentication\TokenGenerator\JWT\JWTGeneratorInterface;
+use CodeIgniter\Shield\Entities\User;
+
+class JWTGenerator
```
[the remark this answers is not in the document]
kenjis: I combined two JWT interfaces into one.
https://github.com/codeigniter4/shield/pull/195/files#diff-cd96a8cd175de721186884550869ed27cff45cf64436c0b17d563573f37cb8c7

src/Config/Auth.php
kenjis: I changed `$jwtConfig`.
https://github.com/codeigniter4/shield/pull/195/files#diff-c48af47762f5a9ba44e0e0b95dd538d48276cf1bc59a8c157bdcddc812184d54R208

src/Filters/JWTAuth.php
```diff
@@ -0,0 +1,88 @@
...
+     *
+     * @return Response|void
+     */
+    public function before(RequestInterface $request, $arguments = null)
+    {
+        if (! $request instanceof IncomingRequest) {
+            return;
+        }
+
+        helper(['auth', 'setting']);
```
datamweb, line 42:
```suggestion
        helper('setting');
```

datamweb (changes requested):

src/Filters/JWTAuth.php
datamweb: @Kenjis, is there a reason why you don't continue? 
What are we waiting for?

src/Authentication/Authenticators/JWT.php
kenjis: First, the log is recorded in `auth_token_logins` table from the beginning. Because `tokenLoginModel` is used.

Now I added config for logging. The default is *Record only failures*.

src/Filters/JWTAuth.php
kenjis: Thanks. Updated.
I'm not waiting for. Just I didn't have time to finish.

kenjis:
I don't remember much of the implementation as a lot of time has passed, but I think the implementation itself was done in one way or another.

If there is someone who want to try JWT, please test. Of course code reviews are also welcome.
I am going to run the code and see if it really works.

MGatner:
My only JWT CI4 project currently uses Myth and I've had issues installing Shield alongside because they have some conflicting services and factories. I know some community members have been keen on this - maybe check the forums for volunteers?

kenjis:
Good idea! I've posted the forum.

hainm0912:
it finished? how to use this branch?

kenjis:
This should work. You can get the code from my repository:
https://github.com/kenjis/codeigniter-shield/tree/feat-jwt

kenjis:
> how to use this branch?

Update your `composer.json`:
```diff
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,8 @@
     "require": {
         "php": "^7.4 || ^8.0",
         "codeigniter4/framework": "^4.0",
-        "codeigniter4/shield": "^1.0@beta"
+        "codeigniter4/shield": "dev-feat-jwt",
+        "firebase/php-jwt": "^6.2"
     },
     "require-dev": {
         "fakerphp/faker": "^1.9",
@@ -36,5 +37,11 @@
         "slack": "https://codeigniterchat.slack.com"
     },
     "minimum-stability": "dev",
-    "prefer-stable": true
+    "prefer-stable": true,
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/kenjis/codeigniter-shield.git"
+        }
+    ]
 }
```
Run `composer update`.

miguel-rn:
What is the status on JWT authentication?

Any TODOs I could help with?

kenjis:
Rebased to resolve conflicts.

kenjis:
Rebased to resolve conflicts.

kenjis:
> What is the status on JWT authentication?

The implementation was finished. I need to write docs.

> Any TODOs I could help with?

Testing and review. As you see, no one has approved this PR yet.

miguel-rn:
> > What is the status on JWT authentication?
> 
> The implementation was finished. I need to write docs.
> 
> > Any TODOs I could help with?
> 
> Testing and review. As you see, no one has approved this PR yet.

Nice work! I will create an API now to help test this implementation.

kenjis:
Added docs.

kenjis:
Why this error?
```
ERROR: ReservedWord - tests/Authentication/Authenticators/JWTAuthenticatorTest.php:148:9 - 'resource' is a reserved word (see https://psalm.dev/095)
        $users->delete(1);
```
https://github.com/codeigniter4/shield/actions/runs/4696752282/jobs/8327382118?pr=195

kenjis:
After reviewing the implementation today, I felt that the current APIs are too narrow for use cases.
I appreciate your feedback on usability.

kenjis:
Various API changes and additions. I believe that more items can be customized without extending classes.

Also, it may be too late now, but I found firebase/php-jwt does not support JWE. 
If we need JWE, we need to add another package.

src/Authentication/JWT/Adapters/FirebaseAdapter.php
```diff
@@ -0,0 +1,90 @@
...
+                $algorithm = $item['alg'];
+
+                $keys[$item['kid']] = new Key($key, $algorithm);
+            }
+        }
+
+        try {
+            return JWT::decode($encodedToken, $keys);
+        } catch (BeforeValidException|ExpiredException $e) {
+            throw new RuntimeException('Expired JWT: ' . $e->getMessage(), 0, $e);
```
MGatner, line 50: Should these be a library exception instead?

src/Authentication/JWT/Adapters/FirebaseAdapter.php
```diff
@@ -0,0 +1,90 @@
...
+        } catch (
+            InvalidArgumentException|DomainException|UnexpectedValueException
+            |SignatureInvalidException $e
+        ) {
+            throw new RuntimeException('Invalid JWT: ' . $e->getMessage(), 0, $e);
```
MGatner, line 55: Should these be a library exception instead?

src/Authentication/JWT/JWSAdapterInterface.php
```diff
@@ -0,0 +1,35 @@
...
+     * Issues Signed JWT (JWS)
+     *
+     * @param array<mixed>               $payload The payload.
+     * @param string                     $keyset  The key group.
+     *                                            The array key of Config\AuthJWT::$keys.
+     * @param array<string, string>|null $headers An array with header elements to attach.
+     *
+     * @return string JWT (JWS)
+     */
+    public static function encode(
```
MGatner, line 21: I'm curious your decision to make the interface and adapters use static methods. This seems like a missed opportunity for injectable dependencies and maybe even a service.

src/Config/Services.php
```diff
@@ -36,4 +37,16 @@ public static function passwords(bool $getShared = true): Passwords
 
         return new Passwords(config('Auth'));
     }
+
+    /**
+     * JWT Manager.
+     */
+    public static function jwtmanager(bool $getShared = true): JWTManager
```
MGatner, line 44: I see why you didn't make the adapters as a service now; I still think they might work better non-statically.

MGatner (commented):
I haven't actually used this yet. Added a few comments, but overall it looks very good and there appear to have been multiple reviews and people actually using it already.

src/Authentication/JWT/JWSAdapterInterface.php
kenjis: Maybe because the Firebase methods were static and the adapter did not have state, I made the interface static.
Removing static is easy. However, I am not sure if it really should be removed.

src/Config/Services.php
kenjis: If we make the adapter a service, it would not have enough functionality. 
The adapter can only encode and decode JWS; JWT also has JWE, which may need to be supported in the future.

src/Authentication/JWT/Adapters/FirebaseAdapter.php
kenjis: Added `InvalidTokenException`.

src/Authentication/JWT/Adapters/FirebaseAdapter.php
kenjis: Done.

src/Authentication/JWTManager.php
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * Issues Signed JWT (JWS) for a User
+     *
+     * @param array                      $claims  The payload items.
+     * @param int|null                   $ttl     Time to live in seconds.
+     * @param string                     $keyset  The key group.
+     *                                            The array key of Config\AuthJWT::$keys.
+     * @param array<string, string>|null $headers An array with header elements to attach.
+     */
+    public function generateAccessToken(
```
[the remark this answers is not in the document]
kenjis: Changed to `generateToken()`.

src/Authentication/JWT/Adapters/FirebaseAdapter.php
```diff
@@ -0,0 +1,121 @@
...
+use Firebase\JWT\Key;
+use Firebase\JWT\SignatureInvalidException;
+use InvalidArgumentException;
+use stdClass;
+use UnexpectedValueException;
+
+class FirebaseAdapter implements JWSAdapterInterface
+{
+    /**
+     * {@inheritDoc}
```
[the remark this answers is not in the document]
kenjis: It does help IDEs, and we use `{@inheritDoc}` in the CI4 repository.

Screenshot 2023-04-18 13 24 40
Screenshot 2023-04-18 13 29 46

docs/addons/jwt.md
```diff
@@ -0,0 +1,144 @@
+# JWT Authentication
+
+To use JWT Authentication, you need additional setup and configuration.
```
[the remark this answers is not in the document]
kenjis: Added explanation of JWT and how-to.

kenjis:
Added a lot of explanations in the docs: https://github.com/kenjis/codeigniter-shield/blob/feat-jwt/docs/addons/jwt.md#jwt-authentication

src/Authentication/JWT/JWSAdapterInterface.php
kenjis: There is no need to be `static`, so I removed.
It seems I don't like static.

src/Authentication/JWT/Adapters/FirebaseAdapter.php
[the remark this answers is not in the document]
lonnieezell: Fair enough.

lonnieezell (approved):
Docs look much better. Thank you! 

I haven't tried it live, but it looks good to me.

docs/addons/jwt.md
```diff
@@ -0,0 +1,312 @@
...
+
+Set your secret key in the `$keys` property, or set it in your `.env` file.
+
+E.g.:
+```dotenv
+authjwt.keys.default.0.secret = 8XBFsF6HThIa7OV/bSynahEch+WbKrGcuiJVYPiwqPE=
+```
+
+It needs more than 256 bits random string. You can get a secure random string
+with the following command:
```
michalsn, line 86: I think we should show this in a larger context so that users have no doubt about how it works.

```suggestion
It needs at least 256 bits random string. The length of the secret depends on the 
algorithm we use. The default one is `HS256`, so to ensure that the hash value is 
secure and not easily guessable, the secret key should be at least as long as the 
hash function's output - 256 bits (32 bytes). You can get a secure random string 
with the following command:
```

kenjis: Thanks! It is true.

michalsn (approved):

kenjis:
@MGatner @datamweb Can you review?

datamweb:
@kenjis will try to do today.

docs/addons/jwt.md
```diff
@@ -0,0 +1,315 @@
...
+     * Authenticate Existing User and Issue JWT.
+     */
+    public function jwtLogin(): ResponseInterface
+    {
+        // Get the validation rules
+        $rules = $this->getValidationRules();
+
+        // Validate credentials
+        if (! $this->validateData($this->request->getPost(), $rules)) {
+            return $this->failValidationErrors($this->validator->getErrors(), 422);
```
datamweb, line 134:
```php
return $this->failValidationErrors($this->validator->getErrors(), 422);
```
Here, 422 cannot be a `int`. Maybe using `setStatusCode(422)` is better.
Considering that I don't know if you intended to set the status code or just a 422 error, I didn't add the change suggestion.

(https://github.com/datamweb/CodeIgniter4/blob/e151d3a25f3ed6a2ff29d022499ed21cd3ade259/system/API/ResponseTrait.php#L238-L248)

docs/addons/jwt.md
```diff
@@ -0,0 +1,315 @@
...
+        /** @var Session $authenticator */
+        $authenticator = auth('session')->getAuthenticator();
+
+        // Check the credentials
+        $result = $authenticator->check($credentials);
+        if (! $result->isOK()) {
+            return $this->failUnauthorized($result->reason());
+        }
+
+        // Login is successful.
```
datamweb, line 151: The successful/unsuccessful login of the user must be recorded in the database. Even if it is only to receive a token, in any case, the user will login first with credentials. Also, pay attention to this issue, we cannot use filter `session` /`tokens` on **auth/jwt**.

docs/addons/jwt.md
```diff
@@ -0,0 +1,315 @@
...
+        ];
+    }
+}
+```
+
+You could send a request with the existing user's credentials by curl like this:
+
+```console
+curl --location 'http://localhost:8080/auth/jwt' \
+--header 'Content-Type: application/x-www-form-urlencoded' \
```
datamweb, line 194: If I were the author of this topic, I would write the code so that the data is sent as **JSON**.

src/Authentication/JWTManager.php
```diff
@@ -0,0 +1,85 @@
...
+     * @param array<string, string>|null $headers An array with header elements to attach.
+     */
+    public function generateToken(
+        User $user,
+        array $claims = [],
+        ?int $ttl = null,
+        $keyset = 'default',
+        ?array $headers = null
+    ): string {
+        $payload = array_merge(
```
datamweb, line 48: We've used **array unpacking** before, is there a specific reason you haven't used it?

src/Config/AuthJWT.php
```diff
@@ -0,0 +1,87 @@
...
+     * Record Login Attempts
+     * --------------------------------------------------------------------
+     * Whether login attempts are recorded in the database.
+     *
+     * Valid values are:
+     * - Auth::RECORD_LOGIN_ATTEMPT_NONE
+     * - Auth::RECORD_LOGIN_ATTEMPT_FAILURE
+     * - Auth::RECORD_LOGIN_ATTEMPT_ALL
+     */
+    public int $recordLoginAttempt = Auth::RECORD_LOGIN_ATTEMPT_FAILURE;
```
datamweb, line 86: Kenjis took this time setting from me, at first I thought there was a problem with the code not registering a successful login.
Why is it not set to `Auth::RECORD_LOGIN_ATTEMPT_ALL` by default?

src/Language/fa/Auth.php
```diff
@@ -20,6 +20,10 @@
     'unableSendEmailToUser' => 'متاسفانه, در ارسال ایمیل مشکلی پیش آمد. ما نتوانستیم ایمیلی را به "{0}" ارسال کنیم.',
     'throttled'             => 'درخواست های بسیار زیادی از این آدرس IP انجام شده است. می توانید بعد از {0} ثانیه دوباره امتحان کنید.',
     'notEnoughPrivilege'    => 'شما مجوز لازم برای انجام عملیات مورد نظر را ندارید.',
+    // JWT Exceptions
+    'invalidJWT'     => '(To be translated) The token is invalid.',
+    'expiredJWT'     => '(To be translated) The token has expired.',
+    'beforeValidJWT' => '(To be translated) The token is not yet available.',
```
datamweb, line 26:
```suggestion
    // JWT Exceptions
    'invalidJWT'     => 'توکن معتبر نمی باشد.',
    'expiredJWT'     => 'توکن منقضی شده است.',
    'beforeValidJWT' => 'در حال حاضر امکان استفاده از توکن وجود ندارد.',
```

datamweb (changes requested):

datamweb:
Question, how can the site administrator make the tokens expire in general? (I think he should change the secret code. If so, I'd prefer you explain it in the documentation.)

And the next question is there a way to expire the token for a specific user?

Please update the [README](https://github.com/codeigniter4/shield/blob/develop/README.md#authentication-methods) file, the reference to support **JWT** is good.

kenjis:
> Question, how can the site administrator make the tokens expire in general? (I think he should change the secret code. If so, I'd prefer you explain it in the documentation.)
> And the next question is there a way to expire the token for a specific user?

Tokens are to be validated by defining the conditions that make it invalid.

If you want to invalidate tokens to a specific user, you can do it by specifying the user ID and issued at.

Also, as you say, If you change the key, all tokens signed with that key will be invalidated.

docs/addons/jwt.md
```diff
@@ -0,0 +1,315 @@
...
+
+To use JWT Authentication, you need a controller that issues JWTs.
+
+Here is a sample controller. When a client posts valid credentials (email/password),
+it returns a new JWT.
+
+```php
+// Routes.php
+$routes->post('auth/jwt', '\App\Controllers\Auth\LoginController::jwtLogin');
+
```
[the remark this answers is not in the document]
kenjis: Why? Routes.php does not have `declare(strict_types=1)` by default.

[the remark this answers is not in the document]
datamweb: Kenjis I meant to add in the controller(LoginController).

docs/addons/jwt.md
kenjis: Thanks! You are correct.

422 came from https://github.com/codeigniter4/shield/blob/779bb12aac715460abde88d84e02bab9776b96ad/docs/guides/mobile_apps.md?plain=1#L44
But it seems the code is not good for.

> [15.5.21. ](https://www.rfc-editor.org/rfc/rfc9110.html#section-15.5.21)[422 Unprocessable Content](https://www.rfc-editor.org/rfc/rfc9110.html#name-422-unprocessable-content)
>
> The 422 (Unprocessable Content) status code indicates that the server understands the content type of the request content (hence a [415 (Unsupported Media Type)](https://www.rfc-editor.org/rfc/rfc9110.html#status.415) status code is inappropriate), and the syntax of the request content is correct, but it was unable to process the contained instructions. For example, this status code can be sent if an XML request content contains well-formed (i.e., syntactically correct), but semantically erroneous XML instructions.
>
> http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml
https://www.rfc-editor.org/rfc/rfc9110.html#section-15.5.21

docs/addons/jwt.md
kenjis: If we use `$authenticator->attempt($credentials)`, the login attempt will be stored in the database.
But it adds auth data in the session. The client is actually logged in with the Session Authenticator.

I thought the client does not need to be logged in actually, the current code just checks the credentials.

But I found that even if the user is banned, a new JWT will be issued.

docs/addons/jwt.md
kenjis: Changed the code.

src/Config/Auth.php
```diff
@@ -122,6 +127,7 @@ class Auth extends BaseConfig
     public array $authenticators = [
         'tokens'  => AccessTokens::class,
         'session' => Session::class,
+        'jwt'     => JWT::class,
```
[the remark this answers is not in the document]
kenjis: I think not. If a dev wants to use, they can add by themself.

JWT Auth is an optional feature. It requires additional setup.
So the line 130 should be removed, and add the instruction to add it by themself in the docs?

src/Config/AuthJWT.php
kenjis: The login here means the JWT is valid and authenticated. It is not the login attempt with id/password.
That is, if you set `Auth::RECORD_LOGIN_ATTEMPT_ALL`, every request to the protected pages will be recorded in the database.

See also https://github.com/codeigniter4/shield/pull/195#discussion_r886792961

docs/addons/jwt.md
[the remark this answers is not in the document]
kenjis: Done.

docs/addons/jwt.md
```diff
@@ -0,0 +1,315 @@
...
+## Issuing JWTs
+
+To use JWT Authentication, you need a controller that issues JWTs.
+
+Here is a sample controller. When a client posts valid credentials (email/password),
+it returns a new JWT.
+
+```php
+// Routes.php
+$routes->post('auth/jwt', '\App\Controllers\Auth\LoginController::jwtLogin');
```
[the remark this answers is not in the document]
kenjis: Done.

src/Config/Auth.php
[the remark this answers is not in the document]
kenjis: Changed: https://github.com/codeigniter4/shield/pull/195/commits/200c5ea5049b6d387cc663b7c47ae92293873a0b

docs/addons/jwt.md
kenjis: Changed to JSON request.

src/Authentication/JWTManager.php
kenjis: I don't use array unpacking, because it is not fully compatible with array_merge().
We need to wait for droping PHP 8.0 and before.
https://www.php.net/releases/8.1/en.php#array_unpacking_support_for_string_keyed_arrays

kenjis: And it seems array_merge() is faster than unpacking.

The benchmark result on my mac.
```
$ php test.php 
array_merge:0.76221704483032
unpacking:1.3178927898407
```
See https://qiita.com/shigakin/items/5e4a2784bbd6b227f4d3

datamweb: Thanks, the links were interesting.

src/Config/AuthJWT.php
datamweb: Hmm, I got it!

docs/addons/jwt.md
datamweb: thanks!

datamweb (approved):
@kenjis, thanks, everything seems to be working fine now.

docs/addons/jwt.md
datamweb: > 422 came from

There, too, it seems that it should change.

kenjis:
@datamweb Thank you for the detailed review!

kenjis:
@MGatner Can you approve? Without your approve, I cannot merge this.

MGatner (approved):

kenjis:
Thank you all!

PR #196, opened 2022-06-01
refactor: change logout() return type to void

kenjis:
- nobody checks the return value of `logout()`
  - normally, the logout process cannot fail.
  -  if something is wrong, it should throw an Exception.
- `Events::trigger()` may return false, but it does not mean errors.
  - when one listener returns false, the execution of listeners stops.
  - https://github.com/codeigniter4/CodeIgniter4/blob/ad956b426a1595a490f6b83b9e74338e0eb219e4/system/Events/Events.php#L149-L161


lonnieezell (approved):

MGatner (approved):
Yes I agree. Logout should be an idempotent action, which makes any return largely unhelpful.

PR #198, opened 2022-06-02
chore: update rector to ^0.13.3

kenjis:


src/Controllers/LoginController.php
```diff
@@ -39,7 +39,7 @@ public function loginAction()
             return redirect()->route('login')->withInput()->with('error', $result->reason());
         }
 
-        $user = $result->extraInfo();
+        $result->extraInfo();
```
pjsde, line 42: I think this line can be removed as it is not being used for anything

src/Controllers/LoginController.php
```diff
@@ -57,7 +57,7 @@ public function loginAction()
      */
     public function logoutAction()
     {
-        $user = auth()->user();
+        auth()->user();
```
pjsde, line 60: I think this line can be removed as it is not being used for anything

MGatner (changes requested):
Looks good after the removals that @pjsde noted

src/Controllers/LoginController.php
kenjis: Removed.

src/Controllers/LoginController.php
kenjis: Removed.

MGatner (approved):

PR #204, opened 2022-06-03
fix: display message magic link controller if email not send

datamweb:
Fixed issue with displaying success message if email was not sent.
The **success message should not** be displayed when the **email is not sent** to the user for any reason.


datamweb:
As usual, I forgot to run `cs-fix`.

src/Language/ja/Auth.php
```diff
@@ -4,18 +4,19 @@
...
+    'badAttempt'            => 'ログインできません。認証情報を確認してください。', // 'Unable to log you in. Please check your credentials.',
+    'noPassword'            => 'パスワードのないユーザーは認証できません。', // 'Cannot validate a user without a password.',
+    'invalidPassword'       => 'ログインできません。パスワードを確認してください。', // 'Unable to log you in. Please check your password.',
+    'noToken'               => 'すべてのリクエストは、AuthorizationヘッダーにBearerトークンが必要です。', // 'Every request must have a bearer token in the Authorization header.',
+    'badToken'              => 'アクセストークンが無効です。', // 'The access token is invalid.',
+    'oldToken'              => 'アクセストークンの有効期限が切れています。', // 'The access token has expired.',
+    'noUserEntity'          => 'パスワード検証のため、Userエンティティを指定する必要があります。', // 'User Entity must be provided for password validation.',
+    'invalidEmail'          => 'メールアドレスが一致しません。', // 'Unable to verify the email address matches the email on record.',
+    'unableSendEmailToUser' => '申し訳ありませんが、メールの送信に問題がありました。 "{0}"にメールを送信できませんでした。', // 'Sorry, there was a problem sending the email. We could not send an email to "{0}".',
```
datamweb, line 18: @kenjis please check it.

kenjis: Okay.

src/Controllers/MagicLinkController.php
```diff
@@ -77,14 +77,18 @@ public function loginAction()
...
             ->setSubject(lang('Auth.magicLinkSubject'))
             ->setMessage(view(setting('Auth.views')['magic-link-email'], ['token' => $token]))
             ->send();
 
-        return $this->displayMessage();
+        if ($emailSend) {
+            return $this->displayMessage();
+        }
+
+        return redirect()->route('magic-link')->with('error', lang('Auth.unableSendEmailToUser', [$user->email]));
```
kenjis, line 91: This code is different from Email2FA/EmailActivator:
https://github.com/codeigniter4/shield/blob/c66012f887d9e9ab14183472b162114f48ad4561/src/Authentication/Actions/Email2FA.php#L80-L82

The error message is user friendly, but it seems there is no log of the mail sending failure.

datamweb: @kenjis , I really disagree with that.
I like the error message is user friendly.
Why? Just imagine yourself instead of the user.
If necessary, the site administrator should see the logs of file **writable\logs**.
Friends, **what do you think about this?**
**Do you agree with the following code?**
```
 if ($return === false) { 
     throw new RuntimeException('Cannot send email for user: ' . $user->email); 
 } 
```

kenjis: I forgot that the Email class logs properly.

Okay, this implementation has no problem.

kenjis: Can you change the code like this (the variable name) as with other classes?
```php
$return = emailer()->setFrom(setting('Email.fromEmail'), setting('Email.fromName') ?? '')
```
https://github.com/codeigniter4/shield/blob/c66012f887d9e9ab14183472b162114f48ad4561/src/Authentication/Actions/Email2FA.php#L74

datamweb: **Done.**

kenjis (approved):
LGTM!

Issue #157, opened 2022-05-20
Composer Installation Not working


vimkaf:
Running composer require codeigniter4/shield throws 
 ` Could not find a version of package codeigniter4/shield matching your minimum-stability (stable). Require it with an expli  
  cit version constraint allowing its desired stability.`

Is shield ready for production use and what is the stable version?

datamweb:
Hi, for now, use the following command.
`composer require codeigniter4/shield:dev-develop`

lonnieezell:
Since there hasn't been a release yet, there is no stable version. We're very close to a release, though. 

For now, I would add these lines to your composer.json: 

```
"minimum-stability": "dev",
"prefer-stable": true,
```

vimkaf:
Working now. looking forward to a release. Thanks for all the work you do.

PR #205, opened 2022-06-04
feat: add sheild:setup command for initial setup

kenjis:
```
$ php spark shield:setup
```
- install config files: `Auth.php` and `AuthGroups.php`
- update `BaseController` and `routes.php`

TODO:
- [x] add tests
- [x] add running migrations
- [x] add docs

### DIFF
--- a/docs/0 - install.md
+++ b/docs/0 - install.md
@@ -20,7 +20,7 @@ table to store configuration options. As such, you should run the migrations.
 
 ### Command Setup
 
-1. Run the following command. This command does *Manual Setup* 1. through 3. and runs the migrations.
+1. Run the following command. This command handles steps 1-3 of *Manual Setup* and runs the migrations.
 
 ```
 > php spark shield:setup
--- a/docs/0 - install.md
+++ b/docs/0 - install.md
@@ -18,6 +18,18 @@ table to store configuration options. As such, you should run the migrations.
 
 ## Initial Setup
 
+### Command Setup
+
+1. Run the following command. This command does *Manual Setup* 1. through 3. and runs the migrations.
+
+```
+> php spark shield:setup
+```
+
+2. If you are running CodeIgniter v4.1.x, go to *Manual Setup* 4.
+
+### Manual Setup
+
 There are a few setup items to do before you can start using Shield in
 your project.
 
--- a/src/Commands/Setup.php
+++ b/src/Commands/Setup.php
@@ -104,8 +104,8 @@ class Setup extends BaseCommand
         $file     = 'Config/Auth.php';
         $replaces = [
             'namespace CodeIgniter\Shield\Config'    => 'namespace Config',
-            "use CodeIgniter\\Config\\BaseConfig;\n" => '',
-            'extends BaseConfig'                     => 'extends \\CodeIgniter\\Shield\\Config\\Auth',
+            "use CodeIgniter\\Config\\BaseConfig;" => 'CodeIgniter\\Shield\\Config\\Auth as ShieldAuth;',
+            'extends BaseConfig'                     => 'extends ShieldAuth',
         ];
 
         $this->copyAndReplace($file, $replaces);
--- a/src/Commands/Setup.php
+++ b/src/Commands/Setup.php
@@ -103,9 +103,9 @@ class Setup extends BaseCommand
     {
         $file     = 'Config/Auth.php';
         $replaces = [
-            'namespace CodeIgniter\Shield\Config'    => 'namespace Config',
-            "use CodeIgniter\\Config\\BaseConfig;" => 'CodeIgniter\\Shield\\Config\\Auth as ShieldAuth;',
-            'extends BaseConfig'                     => 'extends ShieldAuth',
+            'namespace CodeIgniter\Shield\Config'  => 'namespace Config',
+            'use CodeIgniter\\Config\\BaseConfig;' => 'CodeIgniter\\Shield\\Config\\Auth as ShieldAuth;',
+            'extends BaseConfig'                   => 'extends ShieldAuth',
         ];
 
         $this->copyAndReplace($file, $replaces);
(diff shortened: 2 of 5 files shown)


MGatner (approved):
It's a shame not to use Publisher for this, but you would need to do some intermediate steps to get the text replaced (or maybe do that as follow-up?). I've thought namespace replacement would be a good feature for Publisher down the road, maybe your replacement arrays is a more flexible solution.

kenjis:
This command is based on the Myth:Auth command with the addition of text replacement.
Publisher does not have the ability to replace code in existing files.
So I would like to change to use Publisher after adding that.

src/Commands/Setup.php
```diff
@@ -0,0 +1,236 @@
...
+        $this->publishConfig();
+    }
+
+    private function publishConfig()
+    {
+        $this->publishConfigAuth();
+        $this->publishConfigAuthGroups();
+
+        $this->setupHelper();
+        $this->setupRoutes();
```
datamweb, line 81: $this->setupMigrates();

src/Commands/Setup.php
```diff
@@ -0,0 +1,236 @@
...
+    private function setupRoutes()
+    {
+        $file = 'Config/Routes.php';
+
+        $check   = 'service(\'auth\')->routes($routes);';
+        $pattern = '/(.*)(\n' . preg_quote('$routes->', '/') . '[^\n]+?\n)/su';
+        $replace = '$1$2' . "\n" . $check . "\n";
+
+        $this->replace($file, $check, $pattern, $replace);
+    }
```
datamweb, line 235: private function setupMigrates(): void
    {
        // Call those commands!
        command('migrate -n CodeIgniter\\\\Settings\\\\');
        command('migrate -n CodeIgniter\\\\Shield\\\\');
    }

datamweb (changes requested):
If the proposed changes are made. Everything needed to install the shield is done with this command.
No confusion and quick installation.
@kenjis, What do you think about this issue?

src/Commands/Setup.php
kenjis: Good suggestion!

src/Commands/Setup.php
MGatner: I like adding migrations to this but probably would make more sense to use the class directly instead of using `command()`.

datamweb: I agree, it was logical.

src/Commands/Setup.php
```diff
@@ -0,0 +1,215 @@
...
+        $this->writeFile($file, $content);
+    }
+
+    private function publishConfigAuth()
+    {
+        $file     = 'Config/Auth.php';
+        $replaces = [
+            'namespace CodeIgniter\Shield\Config'    => 'namespace Config',
+            "use CodeIgniter\\Config\\BaseConfig;\n" => '',
+            'extends BaseConfig'                     => 'extends \\CodeIgniter\\Shield\\Config\\Auth',
```
lonnie-vault, line 108:
```suggestion
            "use CodeIgniter\\Config\\BaseConfig;" => 'CodeIgniter\\Shield\\Config\\Auth as ShieldAuth',
            'extends BaseConfig'                     => 'extends ShieldAuth',
```

src/Commands/Setup.php
```diff
@@ -0,0 +1,215 @@
...
+        $this->copyAndReplace($file, $replaces);
+    }
+
+    private function publishConfigAuthGroups()
+    {
+        $file     = 'Config/AuthGroups.php';
+        $replaces = [
+            'namespace CodeIgniter\Shield\Config'    => 'namespace Config',
+            "use CodeIgniter\\Config\\BaseConfig;\n" => '',
+            'extends BaseConfig'                     => 'extends \\CodeIgniter\\Shield\\Config\\AuthGroups',
```
lonnie-vault, line 120:
```suggestion
            "use CodeIgniter\\Config\\BaseConfig;" => 'CodeIgniter\\Shield\\Config\\AuthGroups as ShieldAuthGroups',
            'extends BaseConfig'                     => 'extends ShieldAuthGroups',
```

src/Commands/Setup/ContentReplacer.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace CodeIgniter\Shield\Commands\Setup;
+
+class ContentReplacer
+{
+    /**
+     * @param array $replaces [search => replace]
+     */
+    public function replace(string $content, array $replaces): string
```
lonnie-vault, line 10: Not sure it matters, but you could probably replace this entire method with a single `strtr()` call.

lonnie-vault (commented):
Looks good so far, thanks @kenjis!

I wonder if it would make sense to have some advanced flags that allow for: 

1. Copying the view files into the project (and modify the config file to reflect this)
2. Copying the controllers into the project
3. Expand the routes by default (instead of inserting the single line for routes, add all routes as individual calls that can then be modified by the developer). If they copied the controllers, modify these to match. 

I know I would use these features for quite a few projects, honestly.

src/Commands/Setup.php
kenjis: Done.

src/Commands/Setup.php
kenjis: Done.

src/Commands/Setup/ContentReplacer.php
kenjis: Yes, also with a single `str_replace()`.
They are a bit differnt, but `strtr()` is fine for this case.
https://www.php.net/manual/ja/function.strtr.php#117169

Done.

kenjis:
I've finished the implementation.
The last thing is the documentation.

@lonnieezell I don't oppose to add the advanced flags, but it seems a few needs.
So I'm not going to include them in this PR.

src/Commands/Setup/ContentReplacer.php
datamweb: The further we go, the more I enjoy your teamwork. this is fabulous.

docs/0 - install.md
```diff
@@ -18,6 +18,18 @@ table to store configuration options. As such, you should run the migrations.
 
 ## Initial Setup
 
+### Command Setup
+
+1. Run the following command. This command does *Manual Setup* 1. through 3. and runs the migrations.
```
MGatner, line 23: Yours is perfectly fine but I had to read it twice to figure out the punctuation. My suggestion:
```suggestion
1. Run the following command. This command handles steps 1-3 of *Manual Setup* and runs the migrations.
```

MGatner (changes requested):

MGatner (approved):

datamweb:
@kenjis I saw the changes in the documents. it is good.
But what do you think about creating a file **QuickInstallation.md** at this point in time?
We have had several reports of installation failures. Look [forum](https://forum.codeigniter.com/showthread.php?tid=82003&pid=396837#pid396837), #207, #157
I suggest you create file **QuickInstallation.md** at this time with the installation via composer description(`composer require codeigniter4/shield:dev-develop`) and command `php spark shield:setup` .
After the release of the stable version, if you do not need it, we can delete this file .

lonnieezell:
I don't think we need a quick install doc. We just need to address the minimum stability in the readme and the install doc files.

lonnieezell (approved):

PR #206, opened 2022-06-04
refactor: update for PHPStan update

kenjis:
- refactor for CI4/PHPStan update


MGatner (approved):
Glad to see PHPStan handling `model()` a bit better! This is always a sore spot.

kenjis:
Yeah, the current setup is much better. I also like showing errors when using the builder's methods.

PR #209, opened 2022-06-05
feat: readable form errors if there are multiple errors

datamweb:


src/Views/register.php
```diff
@@ -15,7 +15,7 @@
                     <div class="alert alert-danger" role="alert">
                         <?php if (is_array(session('errors'))) : ?>
                             <?php foreach (session('errors') as $error) : ?>
-                                <?= $error ?>
+                                <?= $error . '<br>' ?>
```
MGatner, line 18: How about just plain HTML?
```suggestion
                                <?= $error ?>
                                <br>
```

Also I haven't kept up with the latest - are we still self-closing tags (`<br />`) or is that nonsense over?

MGatner (changes requested):

src/Views/register.php
datamweb: **Done.**
 And must have a start tag, and must not have an end tag. In XHTML documents, write this element as `<br />`.
[See](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br)

MGatner (approved):

PR #211, opened 2022-06-05
fix: Debugbar error after logout 

### LINKED ISSUE #210: Call to a member function getGroups() on null
I point out that when the logout is performed, an error occurs in the development environment with the toolbar active, as shown in the attached image. In production environment or disabling the toolbar the error does not appear.

error

kenjis:
Fixes #210

- fix: update `$this->userState` when logout
- remove unneeded helper loading

### DIFF
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -688,7 +688,8 @@ class Session implements AuthenticatorInterface
         // Trigger logout event
         Events::trigger('logout', $this->user);
 
-        $this->user = null;
+        $this->user      = null;
+        $this->userState = self::STATE_ANONYMOUS;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


MGatner (approved):

datamweb:
Thanks @kenjis  works well now.

kenjis:
When I tried to reproduce the bug in #210, I could not reproduce it, and since the error was not displayed, 
I thought that it was not a bug.
But @datamweb reproduces the error, and I looked the code again, it must be a bug.
I don't know why I did not see the error.

So I changed this PR subject, and a commit message.

datamweb:
I suggest you add the following test:

```
public function testDisplayNotLoggedInAfterLogout(): void
{
        $authenticator = service('auth')->getAuthenticator();
        assert($authenticator instanceof Session);
        $authenticator->login($this->user);
        
        $authenticator->logout($this->user);

        $output = $this->collector->display();
        $this->assertStringContainsString('Not logged in', $output);
}
```

unittest

lonnie-vault (approved):

MGatner (approved):
I love the collaboration! Thanks to all.

PR #212, opened 2022-06-06
fix: RememberMe cookie does not work with Cookie prefix

kenjis:
- fix missing `Cookie.prefix`
- fix hard coded cookie name `remember`

### DIFF
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -495,9 +495,12 @@ class Session implements AuthenticatorInterface
 
     private function getRememberMeToken(): ?string
     {
-        helper('cookie');
+        /** @var IncomingRequest $request */
+        $request = service('request');
 
-        return get_cookie('remember');
+        $cookieName = setting('Cookie.prefix') . setting('Auth.sessionConfig')['rememberCookieName'];
+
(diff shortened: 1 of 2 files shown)


src/Authentication/Authenticators/Session.php
```diff
@@ -495,9 +495,12 @@ private function checkRememberMe(): bool
 
     private function getRememberMeToken(): ?string
     {
-        helper('cookie');
+        /** @var IncomingRequest $request */
+        $request = service('request');
 
-        return get_cookie('remember');
+        $cookieName = setting('Cookie.prefix') . setting('Auth.sessionConfig')['rememberCookieName'];
```
MGatner, line 501: This might be a good candidate for a Config method, like we did with redirect URL. At this point it looks like this is the only place it is used.

MGatner (approved):
I'm not certain why the helper methods didn't work but your test looks clear and code is good.

kenjis:
> I'm not certain why the helper methods didn't work

I'm sure you already know this.
https://github.com/codeigniter4/CodeIgniter4/pull/6080

MGatner:
Yep, found that right after making this comment!

lonnie-vault (approved):

PR #215, opened 2022-06-08
fix: redirect user if `loggedIn()` and called `/login` page

datamweb:
Hello, my hardworking friends.
When the user `loggedIn`, if he calls the `/login` form again, **the form will be displayed!**
This is completely meaningless.
Suggestions came to my mind.
1. Display the message "You are already logged in."
2. User logout if this happens.
3. User transfer to `config('Auth')->loginRedirect()`.
**I chose option 3.**
what is your idea about this?

### DIFF
--- a/src/Controllers/LoginController.php
+++ b/src/Controllers/LoginController.php
@@ -12,6 +12,8 @@ class LoginController extends BaseController
 
     /**
      * Displays the form the login to the site.
+     *
+     * @return
      */
     public function loginView()
     {
--- a/src/Controllers/LoginController.php
+++ b/src/Controllers/LoginController.php
@@ -4,6 +4,7 @@ namespace CodeIgniter\Shield\Controllers;
 
 use App\Controllers\BaseController;
 use CodeIgniter\HTTP\IncomingRequest;
+use CodeIgniter\HTTP\RedirectResponse;
 use CodeIgniter\HTTP\Response;
 
 class LoginController extends BaseController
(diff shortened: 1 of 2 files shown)


src/Controllers/LoginController.php
```diff
@@ -13,8 +13,12 @@ class LoginController extends BaseController
     /**
      * Displays the form the login to the site.
```
kenjis, line 14: Please add `@return`.

lonnieezell (approved):
I agree with this. I thought this change was already here, but I must have made it in Bonfire instead. Once you've addressed @kenjis comment I'm good.

src/Controllers/LoginController.php
datamweb: @kenjis, Sorry, please help.

src/Controllers/LoginController.php
```diff
@@ -12,9 +12,15 @@ class LoginController extends BaseController
 
     /**
      * Displays the form the login to the site.
+     *
+     * @return
```
kenjis, line 16:
```suggestion
     * @return RedirectResponse|string
```
And `use CodeIgniter\HTTP\RedirectResponse;` is needed.

datamweb: **Thanks**, done.

kenjis (approved):

MGatner:
Good addition!

PR #216, opened 2022-06-08
feat: prevent logged-in users from trying to log in again

kenjis:
Basically, a logged-in user is not supposed to log in again without logging out, 
and should be protected so that he/she cannot do so with filters and/or controllers.

But a developer may misconfigure and if a logged-in user logs in with a different user account, 
the session data of the previous user is carried over.

This PR prevents such behavior by making errors.

### DIFF
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -550,6 +550,20 @@ class Session implements AuthenticatorInterface
      */
     public function startLogin(User $user): void
     {
+        /** @var int|string|null $userId */
+        $userId = $this->getSessionKey('id');
+
+        // Check if already logged in.
+        if ($userId !== null) {
+            throw new LogicException(
+                'The user has User Info in Session, so already logged in or in pending login state.'
+                . ' If a logged in user logs in again with other account, the session data of the previous'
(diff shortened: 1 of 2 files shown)


MGatner (commented):
The only exception I see is "act as" functionality - where an admin impersonates another user to simulate their experience using the website. I incorporate this into all my projects because it is very helpful during development to be able to try different user roles and it helps eventual website owners troubleshoot issues with their users' accounts. 

That said, maybe that doesn't count as an actual "login" and needs some other way of changing the underlying active user?

lonnie-vault:
I wouldn't think of that as a login, per se. In Laravel projects I've used [laravel-impersonate](https://github.com/404labfr/laravel-impersonate) and found it very nice. Maybe we need a package like this? Though I'm not sure that's a core component. 

Either way - you could probably override the filter to make this work?

kenjis:
"act as" functionality is not a normal login. It should not fire login event.
You can't use `Session::login()` for it.

MGatner (approved):

lonnieezell (approved):

kenjis:
Fixed a rector error. Merging.

Issue #214, opened 2022-06-07
Call to a member function getGroups() on null within logout


jozefrebjak:
I've got an Error within logout.

![CleanShot 2022-06-07 at 21 33 27@2x](https://user-images.githubusercontent.com/36922215/172467196-3ccf6a17-64f8-4cf2-80f9-4a0a9271f84a.png)

jozefrebjak:
I saw that there was a fix for something related with toolbar, but I had this issue today with fresh install of CI 4.2 and latest shield and PR #211 was merged two days ago.

kenjis:
How did you install Shield?

jlopes90:
The same error continues. I registered an account after logging in and it doesn't work to log out. So I did migrate:refresh then got this error.

jozefrebjak:
@kenjis installed as docs describe instalation, but it seems we need to specify develop version to use, becase beta version was installed.

kenjis:
> I registered an account after logging in and it doesn't work to log out. So I did migrate:refresh then got this error.

It seems you deleted the user's group data in the table.

jlopes90:
> > I registered an account after logging in and it doesn't work to log out. So I did migrate:refresh then got this error.
> 
> It seems you deleted the user's data in the table.

Well, now I'm stuck with this error. It's not to resolve the users table or others. It has to be corrected, so it is possible to login or register. I think myth-auth works because it checks "users" in the database.

kenjis:
> It seems you deleted the user's data in the table.

But if there is no user, nobody can't logged in. So the toolbar shows not logged in.

@jlopes90 Can you show us the output of the following commands?

```
$ php spark db:table user
$ php spark db:table auth_groups_users
```

jlopes90:
> > It seems you deleted the user's data in the table.
> 
> But if there is no user, nobody can't logged in. So the toolbar shows not logged in.
> 
> @jlopes90 Can you show us the output of the following commands?
> 
> ```
> $ php spark db:table user
> $ php spark db:table auth_groups_users
> ```

![image](https://user-images.githubusercontent.com/50026015/172717408-97b387cb-8bb6-4275-b63b-2b5adf53b817.png)

![image](https://user-images.githubusercontent.com/50026015/172717466-57352b7d-c75a-4cae-b2b8-790e18227f1f.png)

All are empty, because I did `migrate:refresh`.

I already tried to enter the address for example: http://localhost:8082/login, still the same error.

[...]

PR #222, opened 2022-06-08
fix: error occurs if a user deleted from the database has session User Info or RememberMe cookie

kenjis:
Reported in https://github.com/codeigniter4/shield/issues/214#issuecomment-1150421000

- fix `Session::loggedIn()` returns wrong result when the user was deleted.
 
**Steps to Reproduce:**
1. register a user
2. login as the user with remember me
3. delete the user from the database
4. access the site


lonnieezell (approved):

MGatner (commented):
Very nice fix.

PR #223, opened 2022-06-09
fix: redirect user after `loggedIn()` and called `/login/magic-link` page

datamweb:
Hello good guys
Ref: #215 
It seems that changes should have been made here as well.

### DIFF
--- a/tests/Controllers/MagicLinkTest.php
+++ b/tests/Controllers/MagicLinkTest.php
@@ -1,0 +1,49 @@
+<?php
+
+namespace Tests\Controllers;
+
+use CodeIgniter\Test\DatabaseTestTrait;
+use CodeIgniter\Test\FeatureTestTrait;
+use Config\Services;
+use Tests\Support\FakeUser;
+use Tests\Support\TestCase;
+
+/**
+ * @internal
+ */
+final class MagicLinkTest extends TestCase
+{
+    use DatabaseTestTrait;
+    use FeatureTestTrait;
+    use FakeUser;
+
+    protected $namespace;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        helper('auth');
+
+        // Add auth routes
+        $routes = service('routes');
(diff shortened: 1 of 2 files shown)


tests/Controllers/MagicLinkTest.php
```diff
@@ -0,0 +1,49 @@
...
+
+        helper('auth');
+
+        // Add auth routes
+        $routes = service('routes');
+        auth()->routes($routes);
+        Services::injectMock('routes', $routes);
+    }
+
+    public function testAfterLoggedInNotAllowDesplayMagicLink()
```
kenjis, line 34: I found a typo.

```suggestion
    public function testAfterLoggedInNotAllowDisplayMagicLink()
```

Also `LoginTest::testAfterLoggedInNotDesplayLoginPage`.

datamweb: Thanks,
Can I apply here or should I apply in PR separate?

kenjis: > One thing at a time: A pull request should only contain one change. That does not mean only one commit, but one change - however many commits it took. The reason for this is that if you change X and Y, but send a pull request for both at the same time, we might really want X but disagree with Y, meaning we cannot merge the request. Using the Git-Flow branching model you can create new branches for both of these features and send two requests.
https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#branching

kenjis (approved):

MGatner (approved):
Not sure what the Rector errors are - got a similar error dump on a different repo also. @samsonasik any ideas?

samsonasik:
@MGatner possibly related with node stmts refresh on multi rules, should be fixed at rector/rector:dev-main ref https://github.com/rectorphp/rector-src/pull/2453

MGatner:
That looks likely! Glad you got a fix in, we can wait it out.

PR #232, opened 2022-06-10
fix: workaround for email only registration/login

### LINKED ISSUE #224: Register without username
I register without username, gives error.

Error message:
https://user-images.githubusercontent.com/50026015/172833798-59665122-053b-407c-803d-474d617ddc9b.png

kenjis:
Fixes #224

### DIFF
--- a/src/Authentication/Passwords/NothingPersonalValidator.php
+++ b/src/Authentication/Passwords/NothingPersonalValidator.php
@@ -52,7 +52,7 @@ class NothingPersonalValidator extends BaseValidator implements ValidatorInterfa
      */
     protected function isNotPersonal(string $password, ?User $user): bool
     {
-        $userName = \strtolower($user->username);
+        $userName = \strtolower($user->username ?? '');
         $email    = \strtolower($user->email);
         $valid    = true;
 
@@ -151,6 +151,10 @@ class NothingPersonalValidator extends BaseValidator implements ValidatorInterfa
      */
     protected function isNotSimilar(string $password, ?User $user): bool
     {
+        if ($user->username === null) {
+            return true;
+        }
+
         $maxSimilarity = (float) $this->config->maxSimilarity;
         // sanity checking - working range 1-100, 0 is off
         if ($maxSimilarity < 1) {
--- a/src/Authentication/Passwords/NothingPersonalValidator.php
+++ b/src/Authentication/Passwords/NothingPersonalValidator.php
@@ -20,7 +20,7 @@ class NothingPersonalValidator extends BaseValidator implements ValidatorInterfa
      */
     public function check(string $password, ?User $user = null): Result
     {
-        $password = \strtolower($password);
+        $password = strtolower($password);
 
         if ($valid = $this->isNotPersonal($password, $user) === true) {
             $valid = $this->isNotSimilar($password, $user);
@@ -52,8 +52,8 @@ class NothingPersonalValidator extends BaseValidator implements ValidatorInterfa
      */
     protected function isNotPersonal(string $password, ?User $user): bool
     {
-        $userName = \strtolower($user->username ?? '');
-        $email    = \strtolower($user->email);
+        $userName = strtolower($user->username ?? '');
+        $email    = strtolower($user->email);
         $valid    = true;
 
         // The most obvious transgressions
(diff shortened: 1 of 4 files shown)


src/Controllers/RegisterController.php
```diff
@@ -66,6 +66,11 @@ public function registerAction(): RedirectResponse
 
         $user->fill($this->request->getPost($allowedPostFields));
 
+        // Workaround for email only registration/login
+        if ($user->username === null) {
+            $user->username = null;
+        }
```
MGatner, line 72: We should definitely think about a better solution down the road, this is pretty hacky 😅
Alternate syntax:
```suggestion
        $user->username ??= null;
```

src/Models/UserModel.php
```diff
@@ -222,6 +222,7 @@ public function save($data): bool
                 lang('Database.emptyDataset', ['update']),
             ];
             if (in_array($e->getMessage(), $messages, true)) {
+                // @TODO Why true? Shouldn't this workaround be removed?
```
MGatner, line 225: I am not a fan of the whole `save()` extension. Lonnie had trouble getting model callbacks to work because the entity isn't passed, but I think finding another way to handle these is important (eventually).

MGatner (approved):
Comment is optional, workaround looks good.

src/Controllers/RegisterController.php
kenjis: After all, the both codes are the same.
Both are funny!

This is a temporary hack, so no problem.

src/Models/UserModel.php
lonnieezell: I think the thing to do is to fix CodeIgniter to pass the original data along in the Model callbacks, honestly. I think that would be helpful for many people.

datamweb:
@kenjis It seems that changes to file `NothingPersonalValidator.php` should be applied to these **PR**.
https://github.com/codeigniter4/shield/blob/00023eee70d65c17cb9bd840a174885feb15bebd/src/Authentication/Passwords/NothingPersonalValidator.php#L15-L20

kenjis:
@datamweb What do you mean?
NothingPersonalValidator is enabled by default, and I did not see any errors.

By the way, the file has functions staring with `\`.
https://github.com/codeigniter4/shield/blob/00023eee70d65c17cb9bd840a174885feb15bebd/src/Authentication/Passwords/NothingPersonalValidator.php#L23

src/Models/UserModel.php
kenjis: > to fix CodeIgniter to pass the original data along in the Model callbacks

I agree with it.

datamweb:
> What do you mean? 

@kenjis **I have such an error!!** and I used the following code in [55](https://github.com/codeigniter4/shield/blob/00023eee70d65c17cb9bd840a174885feb15bebd/src/Authentication/Passwords/NothingPersonalValidator.php#L55) and [163](https://github.com/codeigniter4/shield/blob/00023eee70d65c17cb9bd840a174885feb15bebd/src/Authentication/Passwords/NothingPersonalValidator.php#L163) to fix it.

```
         if(is_null($user->username)){
             return true;
         }
```
err

kenjis:
@datamweb Oh, thanks! I'm using PHP 8.0, so I did not notice.

kenjis:
@datamweb Fixed the error. Is this okay?

datamweb:
> Is this okay?

@kenjis is okay.
Just for user friendliness, it's not better to make changes in

https://github.com/codeigniter4/shield/blob/00023eee70d65c17cb9bd840a174885feb15bebd/src/Collectors/Auth.php#L74

create?

kenjis:
If the username is null, the toolbar output will be blank.
I'm okay with it.

Screenshot 2022-06-12 16 42 15

datamweb:
Thanks @kenjis.

src/Authentication/Passwords/NothingPersonalValidator.php
```diff
@@ -52,8 +52,8 @@ public function check(string $password, ?User $user = null): Result
      */
     protected function isNotPersonal(string $password, ?User $user): bool
     {
-        $userName = \strtolower($user->username);
-        $email    = \strtolower($user->email);
+        $userName = strtolower($user->username ?? '');
+        $email    = strtolower($user->email);
```
MGatner, line 56: Should we anticipate users trying to work without emails?
```suggestion
        $email    = strtolower($user->email ?? '');
```

src/Models/UserModel.php
MGatner: Sounds good! I am fine with this for now.

src/Authentication/Passwords/NothingPersonalValidator.php
kenjis: It is out of scope of this PR.

The current implementation requires email. If we support users without email, I guess more to work.

lonnieezell (approved):

src/Controllers/RegisterController.php
nfaiz: Since the username column is unique, why don't use email value as username for temporary hack?

kenjis: I think we should not store data other than what the column name indicates.
It will only bring confusion.

nfaiz: Yes, but it can only save one (1) row ATM.

MGatner: This is a modification a developer could make for their instance, but using an email address as the username can be a security and privacy issue so this library has worked hard to keep them separate.

PR #238, opened 2022-06-12
fix: datetime for defaultLocale set fa

datamweb:
When `toDateTimeString() ` is used. If `defaultLocale = 'fa'`.
The value of the datetime will be `UTF-8 string (33) "۲۰۲۲-۰۶-۱۲ ۲۲:۳۰:۳۵"`.
In this case, the database enters the information as `0000-00-00 00:0:00`.
Which causes a bug in the shield.
In this PR I tried to solve this problem.

note: This method can also solve the problem. `$time = new Time('now','','en-US') `

```
Data of Table "auth_groups_users":

+----+---------+-------+--------------------+
| id | user_id | group | created_at         |
+----+---------+-------+--------------------+
| 7  | 7       | user  | 0000-00-00 00:0... |
| 22 | 22      | user  | 0000-00-00 00:0... |
| 23 | 23      | user  | 2022-06-12 19:3... |
+----+---------+-------+--------------------+
```

### DIFF
--- a/tests/Authorization/AuthorizableTest.php
+++ b/tests/Authorization/AuthorizableTest.php
@@ -2,6 +2,7 @@
 
 namespace Tests\Authorization;
 
+use CodeIgniter\CodeIgniter;
 use CodeIgniter\I18n\Time;
 use CodeIgniter\Shield\Authorization\AuthorizationException;
 use CodeIgniter\Shield\Models\UserModel;
@@ -302,22 +303,23 @@ final class AuthorizableTest extends TestCase
      */
     public function testCreatedAtIfDefaultLocaleSetFaWithAddGroup(): void
     {
+        if (version_compare(CodeIgniter::CI_VERSION, '4.2.1', '<')) {
+            $this->markTestSkipped('This test does not work on CI v4.2.0 or before due to a bug.');
+        }
+
         $currentLocale = Locale::getDefault();
         Locale::setDefault('fa');
 
-        $this->hasInDatabase('auth_groups_users', [
-            'user_id'    => $this->user->id,
-            'group'      => 'admin',
-            'created_at' => Time::now()->format('Y-m-d H:i:s'),
-        ]);
+        Time::setTestNow('March 10, 2017', 'America/Chicago');
 
         $this->user->addGroup('admin');
 
-        $this->dontSeeInDatabase('auth_groups_users', [
+        $this->seeInDatabase('auth_groups_users', [
             'user_id'    => $this->user->id,
             'group'      => 'admin',
-            'created_at' => '0000-00-00 00:00:00',
+            'created_at' => '2017-03-10 00:00:00',
         ]);
+
         Locale::setDefault($currentLocale);
     }
 }
--- a/tests/Authorization/AuthorizableTest.php
+++ b/tests/Authorization/AuthorizableTest.php
@@ -2,7 +2,6 @@
 
 namespace Tests\Authorization;
 
-use CodeIgniter\CodeIgniter;
 use CodeIgniter\I18n\Time;
 use CodeIgniter\Shield\Authorization\AuthorizationException;
 use CodeIgniter\Shield\Models\UserModel;
--- a/tests/Authorization/AuthorizableTest.php
+++ b/tests/Authorization/AuthorizableTest.php
@@ -303,6 +303,7 @@ final class AuthorizableTest extends TestCase
      */
     public function testCreatedAtIfDefaultLocaleSetFaWithAddGroup(): void
     {
+        // @phpstan-ignore-next-line
         if (version_compare(CodeIgniter::CI_VERSION, '4.2.1', '<')) {
             $this->markTestSkipped('This test does not work on CI v4.2.0 or before due to a bug.');
         }
--- a/tests/Authorization/AuthorizableTest.php
+++ b/tests/Authorization/AuthorizableTest.php
@@ -318,5 +318,6 @@ final class AuthorizableTest extends TestCase
             'group'      => 'admin',
             'created_at' => '0000-00-00 00:00:00',
         ]);
+        Locale::setDefault($currentLocale);
     }
 }
--- a/tests/Authorization/AuthorizableTest.php
+++ b/tests/Authorization/AuthorizableTest.php
@@ -302,7 +302,8 @@ final class AuthorizableTest extends TestCase
      */
     public function testCreatedAtIfDefaultLocaleSetFaWithAddGroup(): void
     {
-        locale::setDefault('fa');
+        $currentLocale = Locale::getDefault();
+        Locale::setDefault('fa');
 
         $this->hasInDatabase('auth_groups_users', [
             'user_id'    => $this->user->id,
--- a/tests/Authorization/AuthorizableTest.php
+++ b/tests/Authorization/AuthorizableTest.php
@@ -303,11 +303,6 @@ final class AuthorizableTest extends TestCase
      */
     public function testCreatedAtIfDefaultLocaleSetFaWithAddGroup(): void
     {
-        // @phpstan-ignore-next-line
-        if (version_compare(CodeIgniter::CI_VERSION, '4.2.1', '<')) {
-            $this->markTestSkipped('This test does not work on CI v4.2.0 or before due to a bug.');
-        }
-
         $currentLocale = Locale::getDefault();
         Locale::setDefault('fa');
 
--- a/tests/Authorization/AuthorizableTest.php
+++ b/tests/Authorization/AuthorizableTest.php
@@ -6,6 +6,7 @@ use CodeIgniter\I18n\Time;
 use CodeIgniter\Shield\Authorization\AuthorizationException;
 use CodeIgniter\Shield\Models\UserModel;
 use CodeIgniter\Test\DatabaseTestTrait;
+use Locale;
 use Tests\Support\FakeUser;
 use Tests\Support\TestCase;
 
@@ -294,5 +295,25 @@ final class AuthorizableTest extends TestCase
         $this->user->addGroup('superadmin');
 
         $this->assertTrue($this->user->can('admin.access'));
+    }
+
+    // see https://github.com/codeigniter4/shield/pull/238
+    public function testCreatedAtIfDefaultLocaleSetFaWithAddGroup(): void
+    {
+        locale::setDefault('fa');
+
+        $this->hasInDatabase('auth_groups_users', [
+            'user_id'    => $this->user->id,
+            'group'      => 'admin',
+            'created_at' => Time::now()->format('Y-m-d H:i:s'),
+        ]);
+
+        $this->user->addGroup('admin');
+
+        $this->dontSeeInDatabase('auth_groups_users', [
+            'user_id'    => $this->user->id,
+            'group'      => 'admin',
+            'created_at' => '0000-00-00 00:00:00',
+        ]);
     }
 }
--- a/tests/Authorization/AuthorizableTest.php
+++ b/tests/Authorization/AuthorizableTest.php
@@ -297,7 +297,9 @@ final class AuthorizableTest extends TestCase
         $this->assertTrue($this->user->can('admin.access'));
     }
 
-    // see https://github.com/codeigniter4/shield/pull/238
+    /**
+     * @see https://github.com/codeigniter4/shield/pull/238
+     */
     public function testCreatedAtIfDefaultLocaleSetFaWithAddGroup(): void
     {
         locale::setDefault('fa');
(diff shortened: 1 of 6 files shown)


datamweb:
@kenjis I'm not sure this is the best solution, please review.
I will also send the **test** after confirmation.

kenjis:
[toDateTimeString()](https://codeigniter4.github.io/CodeIgniter4/libraries/time.html#todatetimestring) returns localized datetime string. So it is not appropriate. 

Your solution is good.

datamweb:
I tried to put the test in the appropriate file. Adding this test to other files(`AccessTokens.php` , `MagicLinkController.php` ...) seems to cause bloating.
However please let me know if you need to change.

tests/Authorization/AuthorizableTest.php
```diff
@@ -295,4 +296,24 @@ public function testCanCascadesToGroupsWithWildcards(): void
 
         $this->assertTrue($this->user->can('admin.access'));
     }
+
+    // see https://github.com/codeigniter4/shield/pull/238
```
kenjis, line 300:
```suggestion
    /**
     * @see https://github.com/codeigniter4/shield/pull/238
     */
```

datamweb: Done.

tests/Authorization/AuthorizableTest.php
```diff
@@ -295,4 +296,24 @@ public function testCanCascadesToGroupsWithWildcards(): void
...
+    public function testCreatedAtIfDefaultLocaleSetFaWithAddGroup(): void
+    {
+        locale::setDefault('fa');
```
kenjis, line 305:
```suggestion
        $currentLocale = Locale::getDefault();
        Locale::setDefault('fa');
```

tests/Authorization/AuthorizableTest.php
```diff
@@ -295,4 +296,26 @@ public function testCanCascadesToGroupsWithWildcards(): void
...
+            'created_at' => Time::now()->format('Y-m-d H:i:s'),
+        ]);
+
+        $this->user->addGroup('admin');
+
+        $this->dontSeeInDatabase('auth_groups_users', [
+            'user_id'    => $this->user->id,
+            'group'      => 'admin',
+            'created_at' => '0000-00-00 00:00:00',
+        ]);
```
kenjis, line 319: Add `Locale::setDefault($currentLocale);`.

datamweb: Done.
I'm sorry to bother you so much for a small change.

tests/Authorization/AuthorizableTest.php
```diff
@@ -295,4 +296,28 @@ public function testCanCascadesToGroupsWithWildcards(): void
...
+            'group'      => 'admin',
+            'created_at' => Time::now()->format('Y-m-d H:i:s'),
+        ]);
+
+        $this->user->addGroup('admin');
+
+        $this->dontSeeInDatabase('auth_groups_users', [
+            'user_id'    => $this->user->id,
+            'group'      => 'admin',
+            'created_at' => '0000-00-00 00:00:00',
```
kenjis, line 319: Isn't `'0000-00-00 00:00:00'` MySQL specific?

Does this test fail before applying your fixes?

tests/Authorization/AuthorizableTest.php
kenjis: No problem. Thanks.

tests/Authorization/AuthorizableTest.php
```diff
@@ -295,4 +296,28 @@ public function testCanCascadesToGroupsWithWildcards(): void
...
+    public function testCreatedAtIfDefaultLocaleSetFaWithAddGroup(): void
+    {
+        $currentLocale = Locale::getDefault();
+        Locale::setDefault('fa');
+
+        $this->hasInDatabase('auth_groups_users', [
+            'user_id'    => $this->user->id,
+            'group'      => 'admin',
+            'created_at' => Time::now()->format('Y-m-d H:i:s'),
+        ]);
```
kenjis, line 312: `hasInDatabase()` inserts a record:
https://codeigniter4.github.io/CodeIgniter4/testing/database.html?highlight=hasindatabase#hasindatabase-table-data
Why do you need to insert the data here?

kenjis:
@datamweb The following test should work, but it does not work because of CI4 Time bug. https://github.com/codeigniter4/CodeIgniter4/pull/6116

```php
    public function testCreatedAtIfDefaultLocaleSetFaWithAddGroup(): void
    {
        $currentLocale = Locale::getDefault();
        Locale::setDefault('fa');

        Time::setTestNow('March 10, 2017', 'America/Chicago');

        $this->user->addGroup('admin');

        $this->seeInDatabase('auth_groups_users', [
            'user_id'    => $this->user->id,
            'group'      => 'admin',
            'created_at' => '2017-03-10 00:00:00',
        ]);

        Locale::setDefault($currentLocale);
    }
```

tests/Authorization/AuthorizableTest.php
datamweb: In fact, I was just trying to check the correct record of the date in the database.
That's why I used it.

datamweb:
> ```
>     public function testCreatedAtIfDefaultLocaleSetFaWithAddGroup(): void
>     {
>         $currentLocale = Locale::getDefault();
>         Locale::setDefault('fa');
> 
>         Time::setTestNow('March 10, 2017', 'America/Chicago');
> 
>         $this->user->addGroup('admin');
> 
>         $this->seeInDatabase('auth_groups_users', [
>             'user_id'    => $this->user->id,
>             'group'      => 'admin',
>             'created_at' => '2017-03-10 00:00:00',
>         ]);
> 
>         Locale::setDefault($currentLocale);
>     }
> ```

@kenjis, So should I replace the test?

tests/Authorization/AuthorizableTest.php
datamweb: Oh my God.
This is a ridiculous mistake, **I'm sorry.**
**Thanks for the explanation.**

MGatner: Not your fault @datamweb, I have always felt that `hasInDatabse()` was very poorly named.

MGatner:
Let's hold off on this until we resolve the issue with the framework. The User Guide seems strongly in favor of `toDateTimeString()` for database inserts, so this might be the correct code and the bug is actually that the framework is localizing the result.

datamweb:
> Let's hold off on this until we resolve the issue with the framework. The User Guide seems strongly in favor of `toDateTimeString()` for database inserts, so this might be the correct code and the bug is actually that the framework is localizing the result.

**Okya**. Please let me know if need to do something.

kenjis:
@MGatner The bug in the framework was fixed in the `develop`, but we can't use it in this repository until it is released.
What do we do?

MGatner:
This PR is still good, it is just the test case that needs the framework change, correct? Since the test case is covering an external bug I don't think it is necessary - we can proceed with the content change. If you want to leave the test in place with a conditional skip it could check the framework version on `CodeIgniter\CodeIgniter`.

MGatner:
... Alternatively I can make the bugfix release of 4.2.1 today and we can use that.

kenjis:
@datamweb Update the test, and let's move on.

```diff
--- a/tests/Authorization/AuthorizableTest.php
+++ b/tests/Authorization/AuthorizableTest.php
@@ -2,6 +2,7 @@
 
 namespace Tests\Authorization;
 
+use CodeIgniter\CodeIgniter;
 use CodeIgniter\I18n\Time;
 use CodeIgniter\Shield\Authorization\AuthorizationException;
 use CodeIgniter\Shield\Models\UserModel;
@@ -302,22 +303,23 @@ public function testCanCascadesToGroupsWithWildcards(): void
      */
     public function testCreatedAtIfDefaultLocaleSetFaWithAddGroup(): void
     {
+        if (version_compare(CodeIgniter::CI_VERSION, '4.2.1', '<')) {
+            $this->markTestSkipped('This test does not work on CI v4.2.0 or before due to a bug.');
+        }
+
         $currentLocale = Locale::getDefault();
         Locale::setDefault('fa');
 
-        $this->hasInDatabase('auth_groups_users', [
-            'user_id'    => $this->user->id,
-            'group'      => 'admin',
-            'created_at' => Time::now()->format('Y-m-d H:i:s'),
-        ]);
+        Time::setTestNow('March 10, 2017', 'America/Chicago');
 
         $this->user->addGroup('admin');
 
-        $this->dontSeeInDatabase('auth_groups_users', [
+        $this->seeInDatabase('auth_groups_users', [
             'user_id'    => $this->user->id,
             'group'      => 'admin',
-            'created_at' => '0000-00-00 00:00:00',
+            'created_at' => '2017-03-10 00:00:00',
         ]);
+
         Locale::setDefault($currentLocale);
     }
 }
```

tests/Authorization/AuthorizableTest.php
```diff
@@ -295,4 +297,29 @@ public function testCanCascadesToGroupsWithWildcards(): void
 
         $this->assertTrue($this->user->can('admin.access'));
     }
+
+    /**
+     * @see https://github.com/codeigniter4/shield/pull/238
+     */
+    public function testCreatedAtIfDefaultLocaleSetFaWithAddGroup(): void
+    {
+        if (version_compare(CodeIgniter::CI_VERSION, '4.2.1', '<')) {
```
MGatner, line 306:
```suggestion
        // @phpstan-ignore-next-line
        if (version_compare(CodeIgniter::CI_VERSION, '4.2.1', '<')) {
```

MGatner (changes requested):
PHPStan doesn't like the constant being variable.

tests/Authorization/AuthorizableTest.php
datamweb: PHPStan doesn't like `// @phpstan-ignore-next-line` . also I doesn't like PHPStan ☺ .

kenjis:
```diff
--- a/rector.php
+++ b/rector.php
@@ -20,6 +20,7 @@
 use Rector\Config\RectorConfig;
 use Rector\Core\ValueObject\PhpVersion;
 use Rector\DeadCode\Rector\ClassMethod\RemoveUnusedPromotedPropertyRector;
+use Rector\DeadCode\Rector\If_\UnwrapFutureCompatibleIfPhpVersionRector;
 use Rector\DeadCode\Rector\MethodCall\RemoveEmptyMethodCallRector;
 use Rector\DeadCode\Rector\StmtsAwareInterface\RemoveJustPropertyFetchForAssignRector;
 use Rector\EarlyReturn\Rector\Foreach_\ChangeNestedForeachIfsToEarlyContinueRector;
@@ -80,6 +81,11 @@
             __DIR__ . '/tests',
         ],
 
+        // check on constant compare
+        UnwrapFutureCompatibleIfPhpVersionRector::class => [
+            __DIR__ . '/tests/Authorization/AuthorizableTest.php',
+        ],
+
         // Ignore files that should not be namespaced to their folder
         NormalizeNamespaceByPSR4ComposerAutoloadRector::class => [
             __DIR__ . '/src/Helpers',
diff --git a/tests/Authorization/AuthorizableTest.php b/tests/Authorization/AuthorizableTest.php
index 3667f7c..0a8ea43 100644
--- a/tests/Authorization/AuthorizableTest.php
+++ b/tests/Authorization/AuthorizableTest.php
@@ -308,6 +308,7 @@ public function testCreatedAtIfDefaultLocaleSetFaWithAddGroup(): void
             $this->markTestSkipped('This test does not work on CI v4.2.0 or before due to a bug.');
         }
 
+        // @phpstan-ignore-next-line
         $currentLocale = Locale::getDefault();
         Locale::setDefault('fa');
 
```

MGatner:
4.2.1 is out, safe to roll back these workarounds

tests/Authorization/AuthorizableTest.php
```diff
@@ -295,4 +297,30 @@ public function testCanCascadesToGroupsWithWildcards(): void
...
+    /**
+     * @see https://github.com/codeigniter4/shield/pull/238
+     */
+    public function testCreatedAtIfDefaultLocaleSetFaWithAddGroup(): void
+    {
+        // @phpstan-ignore-next-line
+        if (version_compare(CodeIgniter::CI_VERSION, '4.2.1', '<')) {
+            $this->markTestSkipped('This test does not work on CI v4.2.0 or before due to a bug.');
+        }
+
```
kenjis, line 310:
```suggestion
```

kenjis:
@datamweb Removing the `if` in the test case is one last job for this PR.
Can you?

datamweb:
Hi @kenjis , I'm **sorry** for the delay. I was busy these few days.
**Thanks** for the review.

kenjis:
Okay, all green!
Merged.

PR #239, opened 2022-06-12
feat: add email validation in MagicLinkController

kenjis:
It is better to send input data to the database after validating it.


lonnieezell (approved):

MGatner (commented):
💪

PR #241, opened 2022-06-13
docs: improve installation

### LINKED ISSUE #240: SQLite3 error
Could we update the install docs to avoid the following issue?

When doing the spark migrate --all command I get the following error:

Unable to connect to the database.
Main connection [SQLite3]: Class "SQLite3" not found

I fixed it by getting rid of the test folder, but it took me half an hour to work out what the problem was. Perhaps a note in the installation section would save others some time.

kenjis:
Fixes #240

- add note about `Class "SQLite3" not found` error

### DIFF
--- a/docs/0 - install.md
+++ b/docs/0 - install.md
@@ -50,6 +50,17 @@ table to store configuration options. As such, you should run the migrations.
 > php spark migrate --all
 ```
 
+---
+
+#### **Note**:
+
+When you run `spark migrate --all`, if you get `Class "SQLite3" not found` error:
+
+1. Remove sample migration files in `tests/_support/Database/Migrations/`
+2. Or install `sqlite3` php extension
+
+---
+
 ## Initial Setup
 
 ### Command Setup
--- a/docs/0 - install.md
+++ b/docs/0 - install.md
@@ -11,7 +11,9 @@ If you have it installed as a phar, or othewise you will need to adjust the way 
 
 ---
 
-**IMPORTANT**: If you get the following error:
+#### **IMPORTANT**:
+
+If you get the following error:
 
 ```
   Could not find a version of package codeigniter4/shield matching your minimum-stability (stable).
--- a/docs/0 - install.md
+++ b/docs/0 - install.md
@@ -11,7 +11,7 @@ If you have it installed as a phar, or othewise you will need to adjust the way 
 
 ---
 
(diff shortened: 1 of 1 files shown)


docs/0 - install.md
```diff
@@ -50,6 +52,17 @@ table to store configuration options. As such, you should run the migrations.
 > php spark migrate --all
 ```
 
+---
+
+#### Note: migration error
+
+When you run `spark migrate --all`, if you get `Class "SQLite3" not found` error:
+
+1. Remove sample migration files in `tests/_support/Database/Migrations/`
```
MGatner, line 61: I don't see those files in this library. Did I miss something?

MGatner: Oh you mean in the implementing project! Got it.

kenjis: Yes!
These files cause the error.
- https://github.com/codeigniter4/appstarter/tree/master/tests/_support/Database/Migrations
- https://github.com/codeigniter4/appstarter/blob/a6f65a777efdc44b89e9bee835a8e5b703833f24/app/Config/Database.php#L65

MGatner: I responded in the other thread but I think we should remove these if they are causing more harm than good. Doesn't affect this PR though

kenjis: I don't know causing more harm than good.
It is a default sample test code. So if devs sees it, they can learn how to use.
And the error happens only in environment with no SQLite3 extension.

MGatner: Maybe the problem is that `migrate --all` is running migrations from the `Tests` namespace, which logically should never happen (and could cause a lot of problems if test migration files were present on a production instance).

datamweb:
How about this?
>**Note**: migration error
>
> When you run `spark migrate --all`, if you get `Class "SQLite3" not found` error:
>
> 1. Remove sample migration files in `tests/_support/Database/Migrations/`
> 2. Or install `sqlite3` php extension

kenjis:
Merge with this for the time being.

kenjis:
@datamweb 
The note looks good, but it is not linkable.
https://github.com/codeigniter4/shield/blob/develop/docs/0%20-%20install.md#note-migration-error

> **Note**
> This is a note

> **Warning**
> This is a warning

Issue #237, opened 2022-06-12
Blank login form (email and password) without username


jlopes90:
I tested the blank login form (email and password), it gave an error because of `username`.

![image](https://user-images.githubusercontent.com/50026015/173240436-534b49f2-046c-4711-9339-3e7236d37a16.png)

https://github.com/codeigniter4/shield/blob/cbd9bfdcdfef2d7d21d9c2571d6b28ec69bcf353/src/Authentication/Authenticators/Session.php#L256-L263

to 

```
$this->loginModel->recordLoginAttempt( 
     $idType, 
     $credentials['email'] ?? $credentials['username'] ?? '',
     $success, 
     $ipAddress, 
     $userAgent, 
     $userId 
 );
```

datamweb:
Did I get it right, you deleted both the **email** and the **username** in the login form?

jlopes90:
> Did I get it right, you deleted both the **email** and the **username** in the login form?

that's not it, the login form has an email and password, I tested it without filling it in and got this error.

datamweb:
In general, you should not send data to the database **without validation**. **This causes security problems**.
In file `RegisterController.php` you must have the following values.

 ```php
    /**
     * Returns the rules that should be used for validation.
     *
     * @return string[]
     */
    protected function getValidationRules(): array
    {
        return [
            //'username'         => 'required|alpha_numeric_space|min_length[3]|is_unique[users.username]',
            'email'            => 'required|valid_email|is_unique[auth_identities.secret]',
            'password'         => 'required|strong_password',
            'password_confirm' => 'required|matches[password]',
        ];
    }
```

datamweb:
Can you send a **PR**?

jlopes90:
> Você pode enviar um **PR** ?

I can do it, but they do it differently.

datamweb:
@jlopes90  team members are very friendly and with you.
Sending **PR** on your behalf is not a problem,
In the worst case, your **PR** will be closed, either by yourself or until the problem is solved.
**Do not be afraid to participate.** 👍 
**CI** Members are all with you.

kenjis:
Yes, missing validation in the login form is a bug, or very bad practice.

lonnieezell:
I agree - there should have been validation there.

datamweb:
I believe it is better to apply **validations to models** as well.
[...]

PR #242, opened 2022-06-13
feat: add validation in LoginController

jlopes90:
Fixes https://github.com/codeigniter4/shield/issues/237

### DIFF
--- a/src/Config/Auth.php
+++ b/src/Config/Auth.php
@@ -129,17 +129,6 @@ class Auth extends BaseConfig
 
     /**
      * --------------------------------------------------------------------
-     * Login validation rules
-     * --------------------------------------------------------------------
-     * Returns the rules that should be used for user login validation.
-     */
-    public array $loginRules = [
-        'email'    => 'required|valid_email|max_length[254]',
-        'password' => 'required',
-    ];
-
-    /**
-     * --------------------------------------------------------------------
      * Allow Registration
      * --------------------------------------------------------------------
      * Determines whether users can register for the site.
@@ -232,6 +221,18 @@ class Auth extends BaseConfig
     public array $validFields = [
         'email',
         'username',
+    ];
+
+    /**
+     * --------------------------------------------------------------------
+     * Login validation rules
+     * --------------------------------------------------------------------
+     * Returns the rules that should be used for user login validation.
+     */
+    public array $loginRules = [
+        // 'username' => 'required|max_length[30]|alpha_numeric_space|min_length[3]',
+        'email'    => 'required|valid_email|max_length[254]',
+        'password' => 'required',
     ];
 
     /**
--- a/src/Config/Auth.php
+++ b/src/Config/Auth.php
@@ -134,7 +134,7 @@ class Auth extends BaseConfig
      * Returns the rules that should be used for user login validation.
      */
     public array $loginRules = [
-        'email'    => 'required|valid_email',
+        'email'    => 'required|valid_email|max_length[254]',
         'password' => 'required',
     ];
 
--- a/src/Config/Auth.php
+++ b/src/Config/Auth.php
@@ -225,18 +225,6 @@ class Auth extends BaseConfig
 
     /**
      * --------------------------------------------------------------------
-     * Login validation rules
-     * --------------------------------------------------------------------
-     * Returns the rules that should be used for user login validation.
-     */
-    public array $loginRules = [
-        // 'username' => 'required|max_length[30]|alpha_numeric_space|min_length[3]',
-        'email'    => 'required|valid_email|max_length[254]',
-        'password' => 'required',
-    ];
-
-    /**
-     * --------------------------------------------------------------------
      * Additional Fields for "Nothing Personal"
      * --------------------------------------------------------------------
      * The NothingPersonalValidator prevents personal information from
--- a/src/Config/Auth.php
+++ b/src/Config/Auth.php
@@ -129,6 +129,17 @@ class Auth extends BaseConfig
 
     /**
      * --------------------------------------------------------------------
+     * Login validation rules
+     * --------------------------------------------------------------------
+     * Returns the rules that should be used for user login validation.
+     */
+    public array $loginRules = [
+        'email'    => 'required|valid_email',
+        'password' => 'required',
+    ];
+
+    /**
+     * --------------------------------------------------------------------
      * Allow Registration
      * --------------------------------------------------------------------
      * Determines whether users can register for the site.
--- a/src/Controllers/LoginController.php
+++ b/src/Controllers/LoginController.php
@@ -28,15 +28,12 @@ class LoginController extends BaseController
      */
     public function loginAction(): RedirectResponse
     {
-        $credentials = $this->request->getPost(setting('Auth.validFields'));
-        $credentials = array_filter($credentials);
-
-        $rules = $this->getValidationRules($credentials);
-
-        if (! $this->validate($rules)) {
+        if (! $this->validate(setting('Auth.loginRules'))) {
             return redirect()->back()->withInput()->with('errors', $this->validator->getErrors());
         }
 
+        $credentials             = $this->request->getPost(setting('Auth.validFields'));
+        $credentials             = array_filter($credentials);
         $credentials['password'] = $this->request->getPost('password');
         $remember                = (bool) $this->request->getPost('remember');
 
@@ -53,33 +50,6 @@ class LoginController extends BaseController
         }
 
         return redirect()->to(config('Auth')->loginRedirect())->withCookies();
-    }
-
-    /**
-     * Returns the rules that should be used for validation.
-     *
-     * @param array $identifier email or username
-     *
-     * @return string[]
-     */
-    protected function getValidationRules(array $identifier): array
-    {
-        $rules = [
-            'password' => 'required',
-        ];
-
-        if (isset($identifier['email'])) {
-            $rules['email'] = 'required|max_length[254]|valid_email';
-        }
-        if (isset($identifier['username'])) {
-            $rules['username'] = 'required|alpha_numeric_space|min_length[3]';
-        }
-
-        if (count($rules) === 1) {
-            $rules['email'] = 'required|max_length[254]|valid_email';
-        }
-
-        return $rules;
     }
 
     /**
--- a/src/Controllers/LoginController.php
+++ b/src/Controllers/LoginController.php
@@ -28,17 +28,25 @@ class LoginController extends BaseController
     /**
      * Attempts to log the user in.
      *
-     * @return Response|string
+     * @return RedirectResponse
      */
-    public function loginAction()
+    public function loginAction(): RedirectResponse
     {
-        /** @var IncomingRequest $request */
-        $request = service('request');
+        // Validate here first, since some things,
+        // like the password, can only be validated properly here.
+        $rules = $this->getValidationRules();
 
-        $credentials             = $request->getPost(setting('Auth.validFields'));
+        /** @var Validation $validation */
+        $validation = service('validation');
+
+        if (! $this->validate($rules)) {
+            return redirect()->back()->withInput()->with('errors', $validation->getErrors());
+        }
+
+        $credentials             = $this->request->getPost(setting('Auth.validFields'));
         $credentials             = array_filter($credentials);
-        $credentials['password'] = $request->getPost('password');
-        $remember                = (bool) $request->getPost('remember');
+        $credentials['password'] = $this->request->getPost('password');
+        $remember                = (bool) $this->request->getPost('remember');
 
         // Attempt to login
         $result = auth('session')->remember($remember)->attempt($credentials);
@@ -56,9 +64,28 @@ class LoginController extends BaseController
     }
 
     /**
+     * Returns the rules that should be used for validation.
+     *
+     * @return string[]
+     */
+    protected function getValidationRules(): array
+    {
+        $rules = [
+            'email'            => 'required',
+            'password'         => 'required',
+        ];
+
+        if (setting('Auth.validFields') === ['email']) {
+            $rules['email'] .= '|valid_email';
+        }
+
+        return $rules;
+    }
+
+    /**
      * Logs the current user out.
      *
-     * @return Response|string
+     * @return RedirectResponse
      */
     public function logoutAction()
     {
--- a/src/Controllers/LoginController.php
+++ b/src/Controllers/LoginController.php
@@ -71,8 +71,8 @@ class LoginController extends BaseController
     protected function getValidationRules(): array
     {
         $rules = [
-            'email'            => 'required',
-            'password'         => 'required',
+            'email'    => 'required',
+            'password' => 'required',
         ];
 
         if (setting('Auth.validFields') === ['email']) {
--- a/src/Controllers/LoginController.php
+++ b/src/Controllers/LoginController.php
@@ -3,9 +3,7 @@
 namespace CodeIgniter\Shield\Controllers;
 
 use App\Controllers\BaseController;
-use CodeIgniter\HTTP\IncomingRequest;
 use CodeIgniter\HTTP\RedirectResponse;
-use CodeIgniter\HTTP\Response;
 
 class LoginController extends BaseController
 {
@@ -27,24 +25,18 @@ class LoginController extends BaseController
 
     /**
      * Attempts to log the user in.
-     *
-     * @return RedirectResponse
      */
     public function loginAction(): RedirectResponse
     {
-        // Validate here first, since some things,
-        // like the password, can only be validated properly here.
-        $rules = $this->getValidationRules();
+        $credentials = $this->request->getPost(setting('Auth.validFields'));
+        $credentials = array_filter($credentials);
 
-        /** @var Validation $validation */
-        $validation = service('validation');
+        $rules = $this->getValidationRules($credentials);
 
         if (! $this->validate($rules)) {
-            return redirect()->back()->withInput()->with('errors', $validation->getErrors());
+            return redirect()->back()->withInput()->with('errors', $this->validator->getErrors());
         }
 
-        $credentials             = $this->request->getPost(setting('Auth.validFields'));
-        $credentials             = array_filter($credentials);
         $credentials['password'] = $this->request->getPost('password');
         $remember                = (bool) $this->request->getPost('remember');
 
@@ -66,18 +58,25 @@ class LoginController extends BaseController
     /**
      * Returns the rules that should be used for validation.
      *
+     * @param array $identifier email or username
+     *
      * @return string[]
      */
-    protected function getValidationRules(): array
+    protected function getValidationRules(array $identifier): array
     {
         $rules = [
-            'email'    => 'required_without[username]|permit_empty|max_length[254]|valid_email',
-            'username' => 'required_without[email]|permit_empty|alpha_numeric_space|min_length[3]',
             'password' => 'required',
         ];
 
-        if (setting('Auth.validFields') === ['email']) {
-            $rules['email'] .= '|valid_email';
+        if (isset($identifier['email'])) {
+            $rules['email'] = 'required|max_length[254]|valid_email';
+        }
+        if (isset($identifier['username'])) {
+            $rules['username'] = 'required|alpha_numeric_space|min_length[3]';
+        }
+
+        if (count($rules) === 1) {
+            $rules['email'] = 'required|max_length[254]|valid_email';
         }
 
         return $rules;
--- a/src/Controllers/LoginController.php
+++ b/src/Controllers/LoginController.php
@@ -71,7 +71,8 @@ class LoginController extends BaseController
     protected function getValidationRules(): array
     {
         $rules = [
-            'email'    => 'required',
+            'email'    => 'required_without[username]|permit_empty|max_length[254]|valid_email',
+            'username' => 'required_without[email]|permit_empty|alpha_numeric_space|min_length[3]',
             'password' => 'required',
         ];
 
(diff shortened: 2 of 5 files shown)


src/Controllers/LoginController.php
```diff
@@ -55,10 +63,29 @@ public function loginAction()
...
 
+    /**
+     * Returns the rules that should be used for validation.
+     *
+     * @return string[]
+     */
+    protected function getValidationRules(): array
+    {
+        $rules = [
+            'email'    => 'required',
```
kenjis, line 74:
```suggestion
            'email'    => 'required|max_length[254]|valid_email',
```

src/Controllers/LoginController.php
```diff
@@ -28,17 +28,25 @@ public function loginView()
...
+    public function loginAction(): RedirectResponse
     {
-        /** @var IncomingRequest $request */
-        $request = service('request');
+        // Validate here first, since some things,
+        // like the password, can only be validated properly here.
+        $rules = $this->getValidationRules();
 
-        $credentials             = $request->getPost(setting('Auth.validFields'));
+        /** @var Validation $validation */
```
kenjis, line 39: See the error.
> PHPDoc tag @var for variable $validation contains unknown class CodeIgniter\Shield\Controllers\Validation.

However the `$validation` below is also not needed, because the controller will have `$this->validator`.

src/Controllers/LoginController.php
```diff
@@ -55,10 +63,29 @@ public function loginAction()
...
+     */
+    protected function getValidationRules(): array
+    {
+        $rules = [
+            'email'    => 'required',
+            'password' => 'required',
+        ];
+
+        if (setting('Auth.validFields') === ['email']) {
+            $rules['email'] .= '|valid_email';
```
kenjis, line 79: Why do you add this only when `setting('Auth.validFields') === ['email']`?

kenjis:
The coding style violation is reported by GitHub Action.
Please fix.
See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#php-style

kenjis:
The PHPUnit test failed. Please fix.

```
 There was 1 failure:
1) Tests\Controllers\LoginTest::testLoginActionUsernameSuccess
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'https://example.com/index.php'
+'https://example.com/index.php/'
/home/runner/work/shield/shield/tests/Controllers/LoginTest.php:124
```
https://github.com/codeigniter4/shield/runs/6871580690?check_suite_focus=true

kenjis:
Please add a test to check validation error if you can.

src/Controllers/LoginController.php
jlopes90: I didn't even notice it's possible "required_without".

src/Controllers/LoginController.php
jlopes90: I was watching "myth-auth"

https://github.com/lonnieezell/myth-auth/blob/bc606fed47812773bb0647c7770fae6edc02d9f4/src/Controllers/AuthController.php#L71-L73

jlopes90:
> Please add a test to check validation error if you can.

image

jlopes90:
> Is the only way to stop validating username to match the login form?

image
image

kenjis:
How about this?
```diff
--- a/src/Controllers/LoginController.php
+++ b/src/Controllers/LoginController.php
@@ -3,9 +3,7 @@
 namespace CodeIgniter\Shield\Controllers;
 
 use App\Controllers\BaseController;
-use CodeIgniter\HTTP\IncomingRequest;
 use CodeIgniter\HTTP\RedirectResponse;
-use CodeIgniter\HTTP\Response;
 
 class LoginController extends BaseController
 {
@@ -27,24 +25,19 @@ public function loginView()
 
     /**
      * Attempts to log the user in.
-     *
-     * @return RedirectResponse
      */
     public function loginAction(): RedirectResponse
     {
+        $credentials = $this->request->getPost(setting('Auth.validFields'));
+        $credentials = array_filter($credentials);
+
         // Validate here first, since some things,
         // like the password, can only be validated properly here.
-        $rules = $this->getValidationRules();
-
-        /** @var Validation $validation */
-        $validation = service('validation');
-
+        $rules = $this->getValidationRules($credentials);
         if (! $this->validate($rules)) {
-            return redirect()->back()->withInput()->with('errors', $validation->getErrors());
+            return redirect()->back()->withInput()->with('errors', $this->validator->getErrors());
         }
 
-        $credentials             = $this->request->getPost(setting('Auth.validFields'));
-        $credentials             = array_filter($credentials);
         $credentials['password'] = $this->request->getPost('password');
         $remember                = (bool) $this->request->getPost('remember');
 
@@ -66,18 +59,20 @@ public function loginAction(): RedirectResponse
     /**
      * Returns the rules that should be used for validation.
      *
+     * @param array $identifier email or username
+     *
      * @return string[]
      */
-    protected function getValidationRules(): array
+    protected function getValidationRules(array $identifier): array
     {
         $rules = [
-            'email'    => 'required_without[username]|permit_empty|max_length[254]|valid_email',
-            'username' => 'required_without[email]|permit_empty|alpha_numeric_space|min_length[3]',
+            'email'    => 'required|max_length[254]|valid_email',
             'password' => 'required',
         ];
 
-        if (setting('Auth.validFields') === ['email']) {
-            $rules['email'] .= '|valid_email';
+        if (! isset($identifier['email']) && isset($identifier['username'])) {
+            unset($rules['email']);
+            $rules['username'] = 'required|alpha_numeric_space|min_length[3]';
         }
 
         return $rules;
```

jlopes90:
@kenjis

image

image

jlopes90:
I think input "email" passes to "login", so check if you use only email or both. It looks like myth-auth.

kenjis:
What do you mean?
I don't see any problem in your *Test email and username* result.

kenjis:
@jlopes90 You don't need to upload screenshots of code .
It can't be copy and paste and is a bit difficult to read.

Please paste the code as text as the following:
` ```php `
` paste code here`
` ``` `

jozefrebjak:
@jlopes90 @kenjis  Why not to move rules to registrar for example, something like this

Registrar:

```php
    /**
     * Registers the validation.
     */
    public static function Validation()
    {
        return [
            'login' => [
                'email'    => 'required_without[username]|permit_empty|max_length[254]|valid_email',
                'username' => 'required_without[email]|permit_empty|alpha_numeric_space|min_length[3]',
                'password' => 'required',
            ],
        ];
    }
```

and call them in Controller like:

```php
        /**
         * Perform validation here
         */
        $rules = config('Validation')->login;

        if (! $this->validate($rules)) {
            return redirect()->back()->withInput()->with('errors', $this->validator->getErrors());
        }
```

Then we don't need private function  `getValidationRules` and also we don't need

```php 
 /** @var Validation $validation */
 $validation = service('validation');
```

MGatner:
> Why not to move rules to registrar

I like how @jozefrebjak is thinking! I'll even take it one step further: since `Controller::validate()` can take a rule group string all you need is the name:
```diff

        /**
         * Perform validation here
         */
-        $rules = config('Validation')->login;
+

-        if (! $this->validate($rules)) {
+        if (! $this->validate('login')) {
            return redirect()->back()->withInput()->with('errors', $this->validator->getErrors());
        }
```

jozefrebjak:
> > Why not to move rules to registrar
> 
> I like how @jozefrebjak is thinking! I'll even take it one step further: since `Controller::validate()` can take a rule group string all you need is the name:
> 
> ```diff
>         /**
>          * Perform validation here
>          */
> -        $rules = config('Validation')->login;
> -
> -        if (! $this->validate($rules)) {
> +        if (! $this->validate('login')) {
>             return redirect()->back()->withInput()->with('errors', $this->validator->getErrors());
>         }
> ```
@MGatner Good to know, I'll update my validations, even more simple solution with less rows.

MGatner:
The only issue with this is that we would need to find a way to resolve Registrars meaningfully. In other words: if `Shield\Config` creates a validation registrar we need to be sure `App\Config` can override it. Maybe the solution is to put the actual rules in `Shield\Config\Auth` and the Registrar injects those?

MGatner:
Okay sorry for all the comments, thinking out loud here. But I think my edit is the way to go. Right now we get user input and filter it by "allowed fields" then generate rules based on that - I think we should skip that altogether and just have to e Config file generate the validation rule set based on "allowed fields".

jozefrebjak:
> The only issue with this is that we would need to find a way to resolve Registrars meaningfully. In other words: if `Shield\Config` creates a validation registrar we need to be sure `App\Config` can override it. Maybe the solution is to put the actual rules in `Shield\Config\Auth` and the Registrar injects those?
> 
> Edit: Well now that I read that it seems silly to use a Config file to inject a Validation setting into another Config file. Maybe these just go into the Auth Config instead of the local method? It could be a method like `redirectUrl()` that checks the configured allowed fields for a dependent rule set.

@MGatner What I see, we already have some validation rules in Registrar. I agree with allowing us to have an option to overwrite these rules with Auth conf, so we also need to move users validation rules to Auth conf. Registrar is fine with App not package what I see now, or in situations where we don't want to allow editing these rules. Auth conf is probably the way to go and call those rules from controllers.

kenjis:
I don't know why you want to move validation rules to Config file.

1. validation rules are not system setting. they are rules for specific one form validation.
2. if you move validation rules to config file, the file soon will be very big.
3. it is difficult to find the validation rules in a big config file. If they are in a controller method, I can go to there with one click.

kenjis:
@jozefrebjak We don't need `$validation = service('validation')` from the beginning.
`$this->validate()` will create `$this->validator`.

MGatner:
As long as the rules are derived from a configurable source I don't think it matters where they come from. What is missing from the current solution is a way to support additional rules if a developer wants to expand the "allowed fields". If we can support that dynamically in the controller method then the move to Config won't be necessary.

jozefrebjak:
@kenjis For me personally, it is better to have all of the validation rules in one place and just call them from controllers. The controller can also be a big file. Here it doesn't matter, but the config file is a better option for me, because we can build a UI around it for the admin to add or remove some rules if it's needed, with functions in control it's not possible. If we, for example, want to edit some rules to achieve some goal in future, we need to extend login controller and type code about rules. If they are in conf file, it's easier. Personally, I don't want to extend controllers and make a copy of them, when that option can be achieved with configuration.

jozefrebjak:
@kenjis @MGatner I can imagine something like I did with register validation https://github.com/codeigniter4/shield/pull/244. With login it's little bit harder, because we can login with username or email.

kenjis:
@jlopes90 It is not a bit harder. It is simple as register validation.

The default config for login form is "email and password". 
So the default value for the login rules are for email and password.

If a user wants to use "username and password", they just need to change the rules.

jlopes90:
@kenjis 

Auth.php
```php
/**
 * Login validation rules
 * Returns the rules that should be used for user login validation.
 */
public array $loginRules = [
    'email'            => 'required|valid_email',
    'password'         => 'required',
];
```

LoginController.php
```php
/**
 * Attempts to log the user in.
 */
public function loginAction(): RedirectResponse
{
    if (! $this->validate(setting('Auth.loginRules'))) {
        return redirect()->back()->withInput()->with('errors', $this->validator->getErrors());
    }

    $credentials['password'] = $this->request->getPost('password');
    $remember                = (bool) $this->request->getPost('remember');

    ...
}
```

jlopes90:
update https://github.com/codeigniter4/shield/pull/242#issuecomment-1156498101

kenjis:
This test result is not good, but the failure means there is something wrong.

```
There was 1 failure:
1) Tests\Controllers\LoginTest::testLoginActionUsernameSuccess
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'https://example.com/index.php'
+'https://example.com/index.php/'
/home/runner/work/shield/shield/tests/Controllers/LoginTest.php:124
```

src/Config/Auth.php
```diff
@@ -127,6 +127,17 @@ class Auth extends BaseConfig
         'tokens',
     ];
 
+    /**
+     * --------------------------------------------------------------------
+     * Login validation rules
```
kenjis, line 132: This would make more sense down next to `$validFields` since they influence each other.

src/Config/Auth.php
```diff
@@ -127,6 +127,17 @@ class Auth extends BaseConfig
...
+     * --------------------------------------------------------------------
+     * Returns the rules that should be used for user login validation.
+     */
+    public array $loginRules = [
+        'email'    => 'required|valid_email',
```
kenjis, line 137: According to the RFC, the maximum length of an email address is 254 characters.

```suggestion
        'email'    => 'required|valid_email|max_length[254]',
```

src/Controllers/LoginController.php
```diff
@@ -58,7 +55,7 @@ public function loginAction()
     /**
      * Logs the current user out.
      *
-     * @return Response|string
+     * @return RedirectResponse
      */
     public function logoutAction()
```
kenjis, line 60:
```suggestion
    public function logoutAction(): RedirectResponse
```
And we can remove `@return` above.

src/Config/Auth.php
```diff
@@ -127,6 +127,17 @@ class Auth extends BaseConfig
         'tokens',
     ];
 
+    /**
+     * --------------------------------------------------------------------
+     * Login validation rules
+     * --------------------------------------------------------------------
+     * Returns the rules that should be used for user login validation.
+     */
+    public array $loginRules = [
```
kenjis, line 136: It would be kind if there is `username` as a comment.
```
        // 'username' => 'required|max_length[30]|alpha_numeric_space|min_length[3]',
```

kenjis:
@jlopes90 We need to fix the test code.
`testLoginActionUsernameSuccess` login fails now.

```diff
--- a/tests/Controllers/LoginTest.php
+++ b/tests/Controllers/LoginTest.php
@@ -5,6 +5,7 @@
 use CodeIgniter\Config\Factories;
 use CodeIgniter\I18n\Time;
 use CodeIgniter\Shield\Authentication\Actions\Email2FA;
+use CodeIgniter\Shield\Config\Auth;
 use CodeIgniter\Test\DatabaseTestTrait;
 use CodeIgniter\Test\FeatureTestTrait;
 use Config\Services;
@@ -73,6 +74,7 @@ public function testLoginActionEmailSuccess(): void
             'password' => 'secret123',
         ]);
 
+        $result->assertSessionHas('user', ['id' => 1]);
         $result->assertStatus(302);
         $result->assertRedirect();
         $this->assertSame(site_url(), $result->getRedirectUrl());
@@ -109,6 +111,14 @@ public function testAfterLoggedInNotDesplayLoginPage(): void
 
     public function testLoginActionUsernameSuccess(): void
     {
+        // Change the validation rules
+        $config             = new Auth();
+        $config->loginRules = [
+            'username' => 'required|max_length[30]|alpha_numeric_space|min_length[3]',
+            'password' => 'required',
+        ];
+        Factories::injectMock('config', 'Auth', $config);
+
         $this->user->createEmailIdentity([
             'email'    => 'foo@example.com',
             'password' => 'secret123',
@@ -119,6 +129,7 @@ public function testLoginActionUsernameSuccess(): void
             'password' => 'secret123',
         ]);
 
+        $result->assertSessionHas('user', ['id' => 1]);
         $result->assertStatus(302);
         $result->assertRedirect();
         $this->assertSame(site_url(), $result->getRedirectUrl());
```

lonnieezell:
I realize I'm late to the party, but I believe the "proper" solution would be to add the rules to a Config\Validation file, as per the [user guide](https://codeigniter.com/user_guide/libraries/validation.html#saving-sets-of-validation-rules-to-the-config-file).  

If we do that, though, the question becomes how do we get that to the user? We could copy the rules into the app's Config\Validation file during setup, maybe? Or provide our own Validation config file with a note of the default rules that can be added to theirs if they wish?

kenjis:
> We could copy the rules into the app's Config\Validation file during setup, maybe?

I oppose it. `Config\Validation` is for app, not for the package.
I don't think we should mix settings without permission.

MGatner:
See my comment on the other PR. I think we should do the same thing here: add `getValidationRules()` that returns `setting('Validation.login') ?? [default rules here]` and leave it out of the Config altogether.

lonnieezell:
As mentioned on the other PR, I like @MGatner solution a lot.

src/Controllers/LoginController.php
```diff
@@ -56,11 +57,23 @@ public function loginAction()
...
+     * Returns the rules that should be used for validation.
      *
-     * @return Response|string
+     * @return string[]
+     */
+    protected function getValidationRules(): array
+    {
+        return setting('Validation.login') ?? [
+            //'username' => 'required|max_length[30]|alpha_numeric_space|min_length[3]',
+            'email'    => 'required|valid_email|max_length[254]',
```
kenjis, line 68: If too long value comes, it is better to invalidate immediately.

```suggestion
            'email'    => 'required|max_length[254]|valid_email',
```

kenjis:
@MGatner @lonnieezell @jozefrebjak 

The way that `getValidationRules()` returns `setting('Validation.login') ?? [default rules here]` seemed to be good,
but will not be able to write a test like this?
https://github.com/codeigniter4/shield/pull/242/files#diff-4ee2439849688f735f904d77d2c14d5bf4270481b2ddf24005f9330ed2a5c32eR112
The test changes validation rules temporarily, but there is no `Validation::$login`.
So we need to add `Validation::$login` dynamically.
But dynamic properties will be deprecated in PHP 8.2.
https://wiki.php.net/rfc/deprecate_dynamic_properties

MGatner:
As stated on the other PR: for now we should proceed assuming the framework will fix BaseConfig for allowing dynamic properties from Registrar.

kenjis:
@jlopes90 
```diff
--- a/src/Models/UserModel.php
+++ b/src/Models/UserModel.php
@@ -118,7 +118,7 @@ public function addToDefaultGroup(User $user): void
     public function fake(Generator &$faker): User
     {
         return new User([
-            'username' => $faker->userName,
+            'username' => str_replace('.', ' ', $faker->userName),
             'active'   => true,
         ]);
     }
diff --git a/tests/Controllers/LoginTest.php b/tests/Controllers/LoginTest.php
index ca42d43..82a55e0 100644
--- a/tests/Controllers/LoginTest.php
+++ b/tests/Controllers/LoginTest.php
@@ -5,10 +5,10 @@
 use CodeIgniter\Config\Factories;
 use CodeIgniter\I18n\Time;
 use CodeIgniter\Shield\Authentication\Actions\Email2FA;
-use CodeIgniter\Shield\Config\Auth;
 use CodeIgniter\Test\DatabaseTestTrait;
 use CodeIgniter\Test\FeatureTestTrait;
 use Config\Services;
+use Config\Validation;
 use Tests\Support\FakeUser;
 use Tests\Support\TestCase;
 
@@ -112,12 +112,13 @@ public function testAfterLoggedInNotDesplayLoginPage(): void
     public function testLoginActionUsernameSuccess(): void
     {
         // Change the validation rules
-        $config             = new Auth();
-        $config->loginRules = [
-            'username' => 'required|max_length[30]|alpha_numeric_space|min_length[3]',
-            'password' => 'required',
-        ];
-        Factories::injectMock('config', 'Auth', $config);
+        $config           = new class () extends Validation {
+            public $login = [
+                'username' => 'required|max_length[30]|alpha_numeric_space|min_length[3]',
+                'password' => 'required',
+            ];
+        };
+        Factories::injectMock('config', 'Validation', $config);
 
         $this->user->createEmailIdentity([
             'email'    => 'foo@example.com',
```

src/Models/UserModel.php
```diff
@@ -118,7 +118,7 @@ public function addToDefaultGroup(User $user): void
     public function fake(Generator &$faker): User
     {
         return new User([
-            'username' => $faker->userName,
+            'username' => str_replace('.', ' ', $faker->userName),
```
MGatner, line 121: Was this necessary? IMO Shield should support periods in usernames and *not* spaces.

tests/Controllers/LoginTest.php
```diff
@@ -109,6 +111,15 @@ public function testAfterLoggedInNotDesplayLoginPage(): void
...
     public function testLoginActionUsernameSuccess(): void
     {
+        // Change the validation rules
+        $config           = new class () extends Validation {
+            public $login = [
+                'username' => 'required|max_length[30]|alpha_numeric_space|min_length[3]',
+                'password' => 'required',
+            ];
+        };
+        Factories::injectMock('config', 'Validation', $config);
```
MGatner, line 121: This is fine but since the Config Factories helper method returns the shared instance you can also just do this:
```suggestion
        config('Validation')->login = [
            'username' => 'required|max_length[30]|alpha_numeric_space|min_length[3]',
            'password' => 'required',
        ];
```

I realize after typing this that this doesn't work anymore once dynamic properties are deprecated 🤦‍♂️

MGatner (changes requested):
@jlopes90 Did you need to change the User fake data to match existing validation, or was that something you added?

src/Models/UserModel.php
kenjis: If so, we need to change the validation rule for registration.
https://github.com/codeigniter4/shield/blob/b0ccc14d5dda68df74130723227fbc0370d88221/src/Controllers/RegisterController.php#L134

MGatner: I'm speaking anecdotally, I don't have an RFC reference or anything, but I would never anticipate allowing a space in a username and I also know many services that allow periods, especially so you can use Gmail-style credentials: `matthew.gatner@gmail.com => matthew.gatner`

kenjis: @MGatner Why don't we merge this, and after that, how about discussing the username validation rules?
This PR is a bit old, and Registration is out of scope of this PR.

MGatner (approved):

lonnieezell:
@jlopes90 @kenjis @MGatner Thanks for slogging through this! I think we ended up someplace better than it would have originally been. 

Merging.

PR #244, opened 2022-06-15
feat: allow to use custom validation rules for registration

jozefrebjak:
This PR simplifies the validation code during registration and gives us the option to change the rules if needed without the need to extend Controller.

### DIFF
--- a/docs/7 - customization.md
+++ b/docs/7 - customization.md
@@ -112,9 +112,9 @@ public $registration = [
 ```
 
 Or use `Registrar`. If you are not aware of the Registrar, you can read more about the class [here](https://codeigniter4.github.io/CodeIgniter4/general/configuration.html#registrars).
-For example:
 
 For example:
+
 
 ```php
 /**
--- a/docs/7 - customization.md
+++ b/docs/7 - customization.md
@@ -114,6 +114,8 @@ public $registration = [
 Or use `Registrar`. If you are not aware of the Registrar, you can read more about the class [here](https://codeigniter4.github.io/CodeIgniter4/general/configuration.html#registrars).
 For example:
 
+For example:
+
 ```php
 /**
  * Registers validation rules for registration.
--- a/docs/7 - customization.md
+++ b/docs/7 - customization.md
@@ -110,25 +110,3 @@ public $registration = [
     'password_confirm' => 'required|matches[password]',
 ];
 ```
-
-Or use `Registrar`. If you are not aware of the Registrar, you can read more about the class [here](https://codeigniter4.github.io/CodeIgniter4/general/configuration.html#registrars).
-
-For example:
-
-
-```php
-/**
- * Registers validation rules for registration.
- */
-public static function Validation()
-{
-    return [
-        $registration = [
-            'username'         => 'required|alpha_numeric_space|min_length[3]|is_unique[users.username]',
-            'email'            => 'required|valid_email|is_unique[auth_identities.secret]',
-            'password'         => 'required|strong_password',
-            'password_confirm' => 'required|matches[password]',
-        ];
-    ];
-}
-```
--- a/docs/7 - customization.md
+++ b/docs/7 - customization.md
@@ -116,7 +116,7 @@ For example:
 
 ```php
 /**
- * Registers the validation.
+ * Registers validation rules for registration.
  */
 public static function Validation()
 {
--- a/docs/7 - customization.md
+++ b/docs/7 - customization.md
@@ -97,7 +97,7 @@ The shield has the following rules for registration:
 ];
 ```
 
-If you need a different set of rules for registration, you can specify them in `Validation` configuration. You can use the configuration in `app\Config\Validation` and specify registration rules like:
+If you need a different set of rules for registration, you can specify them in your `Validation` configuration (**app\Config\Validation.php**) like:
 
 ```php
 //--------------------------------------------------------------------
--- a/docs/7 - customization.md
+++ b/docs/7 - customization.md
@@ -81,3 +81,51 @@ class LoginController extends ShieldLogin
     }
 }
 ```
+
+## Custom validation rules
+
+### Registration
+
+The shield has the following rules for registration:
+
+```php
+[
+    'username'         => 'required|alpha_numeric_space|min_length[3]|is_unique[users.username]',
+    'email'            => 'required|valid_email|is_unique[auth_identities.secret]',
+    'password'         => 'required|strong_password',
+    'password_confirm' => 'required|matches[password]',
+];
+```
+
+If you need a different set of rules for registration, you can specify them in `Validation` configuration. You can use the configuration in `app\Config\Validation` and specify registration rules like:
+
+```php
+//--------------------------------------------------------------------
+// Rules
+//--------------------------------------------------------------------
+public $registration = [
+    'username'         => 'required|alpha_numeric_space|min_length[3]|is_unique[users.username]',
+    'email'            => 'required|valid_email|is_unique[auth_identities.secret]',
+    'password'         => 'required|strong_password',
+    'password_confirm' => 'required|matches[password]',
+];
+```
+
+Or use `Registrar`. If you are not aware of the Registrar, you can read more about the class [here](https://codeigniter4.github.io/CodeIgniter4/general/configuration.html#registrars).
+
+```php
+/**
+ * Registers the validation.
+ */
+public static function Validation()
+{
+    return [
+        $registration = [
+            'username'         => 'required|alpha_numeric_space|min_length[3]|is_unique[users.username]',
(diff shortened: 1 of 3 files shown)


MGatner (commented):
I think this would make more sense down next to `$validFields` since they influence each other. 

I am fine with this approach but I want @kenjis to give his feedback because he preferred the controller method. I do think this Config file is getting very large, but maybe a better approach would be to split out logical sections into their own Config (like `Passwords`).

kenjis (dismissed):
LGTM!

lonnieezell:
Should this be in Config\Validation, instead of the Auth file? That's the [user guide recommendation](https://codeigniter.com/user_guide/libraries/validation.html#saving-sets-of-validation-rules-to-the-config-file).

jozefrebjak:
@kenjis I agree with lonnie that validation rules needs to be in Config\Validation. Then they can be called directly and we don't need to store them with Auth conf. 

Something like: `Validation.php` ?

```php
<?php

namespace CodeIgniter\Shield\Config;

use Config\Validation as BaseValidation;

class Validation extends BaseValidation
{
    /**
     * Register validation rules
     * Returns the rules that should be used for user register validation.
     */
    public array $register = [
        'username'         => 'required|alpha_numeric_space|min_length[3]|is_unique[users.username]',
        'email'            => 'required|valid_email|is_unique[auth_identities.secret]',
        'password'         => 'required|strong_password',
        'password_confirm' => 'required|matches[password]',
    ];
}
```

kenjis:
Do you mean something like this?

- Add `CodeIgniter\Shield\Config\AuthValidation`
- Add `Config\AuthValidation` (extends `CodeIgniter\Shield\Config\AuthValidation`) during setup

jozefrebjak:
Maybe something like this. Now I'm thinking. Do we need to create another config file ? It's not enough to extend base `Validation.php` like:

- Add `CodeIgniter\Shield\Config\Validation` (extends `Config\Validation`) as services now. 

and when we need to change `register` rules in App we just write own rules for `register` in `App/Config/Validation` and they will be used ? I don't know if something like this will work and if it's good practice.

kenjis:
When `Config\Validation` exists, `config()` can't access `CodeIgniter\Shield\Config\Validation`.
Probably `setting()` also can't.

MGatner:
`Config\Validation` is the ideal place for these rules *but* it makes it hard or impossible for developers to adjust them. This is a deficiency of the framework itself. We have two options:

1. Extend `Config\Validation` with **src/Config/Validation.php**, but this will be ignored if the developer has **app/Config/Validation.php**
2. Use `Registrar` to apply the rules to the existing `Validation` class, but then developers have no way to override them

Putting the rules in `Auth` Config at least let's developers override them.

A third option would be for our controller could check first for `Config\Validation::$registerRules` if we wanted to allow that as a source, then fall back on a default ruleset (which could live in the controller, since it wouldn't need overriding).

MGatner:
Actually after typing that all up I like the third option the best. `getValidationRules()` simply becomes `return setting('Validation.registration') ?? [...` and nothing else had to change.

lonnieezell:
I like the third option a lot. Keeping the logic in the controller for easy overriding there, but also checking the Validation config file should the user prefer that. That's the simple solution I knew had to be there lol. 

We would need docs around that with example of how to change, of course.

docs/7 - customization.md
```diff
@@ -81,3 +81,51 @@ class LoginController extends ShieldLogin
     }
 }
 ```
+
+## Custom validation rules
+
+### Registration
+
+The shield has the following rules for registration:
```
MGatner, line 89:
```suggestion
Shield has the following rules for registration:
```

docs/7 - customization.md
```diff
@@ -81,3 +81,51 @@ class LoginController extends ShieldLogin
...
+```php
+[
+    'username'         => 'required|alpha_numeric_space|min_length[3]|is_unique[users.username]',
+    'email'            => 'required|valid_email|is_unique[auth_identities.secret]',
+    'password'         => 'required|strong_password',
+    'password_confirm' => 'required|matches[password]',
+];
+```
+
+If you need a different set of rules for registration, you can specify them in `Validation` configuration. You can use the configuration in `app\Config\Validation` and specify registration rules like:
```
MGatner, line 100:
```suggestion
If you need a different set of rules for registration, you can specify them in your `Validation` configuration (**app\Config\Validation.php**) like:
```

docs/7 - customization.md
```diff
@@ -81,3 +81,51 @@ class LoginController extends ShieldLogin
...
+public $registration = [
+    'username'         => 'required|alpha_numeric_space|min_length[3]|is_unique[users.username]',
+    'email'            => 'required|valid_email|is_unique[auth_identities.secret]',
+    'password'         => 'required|strong_password',
+    'password_confirm' => 'required|matches[password]',
+];
+```
+
+Or use `Registrar`. If you are not aware of the Registrar, you can read more about the class [here](https://codeigniter4.github.io/CodeIgniter4/general/configuration.html#registrars).
+
```
MGatner, line 115:
```suggestion
For example:

```

docs/7 - customization.md
```diff
@@ -81,3 +81,51 @@ class LoginController extends ShieldLogin
...
+```php
+/**
+ * Registers the validation.
```
MGatner, line 118:
```suggestion
 * Registers validation rules for registration.
```

MGatner (approved):
This looks great! I'm glad we chewed on this for a bit, this feels right. I made some doc suggestions, take them or leave them.

MGatner (approved):

docs/7 - customization.md
```diff
@@ -81,3 +81,54 @@ class LoginController extends ShieldLogin
...
+//--------------------------------------------------------------------
+public $registration = [
+    'username'         => 'required|alpha_numeric_space|min_length[3]|is_unique[users.username]',
+    'email'            => 'required|valid_email|is_unique[auth_identities.secret]',
+    'password'         => 'required|strong_password',
+    'password_confirm' => 'required|matches[password]',
+];
+```
+
+Or use `Registrar`. If you are not aware of the Registrar, you can read more about the class [here](https://codeigniter4.github.io/CodeIgniter4/general/configuration.html#registrars).
```
kenjis, line 114: @MGatner I don't see the use case that devs use Registrar like this to their apps.

Registrar needs existing config properties. So if we use Registrar, we need to do:
1. set `Config\Validation::$registration`
2. add `Registrar` to override the property values

jozefrebjak: I'm using it like this now. My validation rules are defined only in Registrar, so I don't need to touch Validation config at all.

kenjis: Oh really? 
Then the `Registrar` injects properties dynamically even if there are no properties in the config class?

Dynamic properties will be deprecated.
https://wiki.php.net/rfc/deprecate_dynamic_properties

kenjis: @jozefrebjak You are correct. Registrar can add new properties.
https://github.com/codeigniter4/CodeIgniter4/blob/fe1dbc22003fb93d1b57addd3c35fe484cf5aedb/tests/system/Config/BaseConfigTest.php#L252-L255

kenjis: I've confirmed that will not work on PHP 8.2.
https://github.com/codeigniter4/CodeIgniter4/issues/6162

jozefrebjak: @kenjis good to know. So we need to define it in Validation config only. What is a purpose of the Registrar then ? Just to override packages configs ? I liked how it works now.

MGatner: To be clear, I think in this case the property should be added to **app/Config/Validation.php**; using Registrar is an second option for cases where that isn't available (like Module implementation, or Local namespace).

lonnieezell (approved):
I think we're all agreed this can merge?

kenjis:
The implementation is okay.

But I hope to remove the description about Registrars in the docs.
Dynamic Property injection by Registrars is under discussion.
At least most users do not need to use Registrar in this case.

jozefrebjak:
> The implementation is okay.
> 
> 
> 
> But I hope to remove the description about Registrars in the docs.
> 
> Dynamic Property injection by Registrars is under discussion.
> 
> At least most users do not need to use Registrar in this case.

@lonnieezell @kenjis @MGatner So, remove it ?

MGatner:
Yes, please remove the reference to Registrars. Whether Registrars will support dynamic properties in the future is under deliberation, but I think we are close to deciding against it. It also isn't necessary here, since the same thing can be accomplished using **app/Config/Validation.php**. If a developer really needs this feature they can always allow dynamic properties on their own Config files.

kenjis (approved):

kenjis:
@jozefrebjak Thank you!

PR #247, opened 2022-06-20
When using different authenticator

heriniaina:
Should use setting('Auth.defaultAuthenticator') instead of 'session'


heriniaina:
Ah ok. Thank you for the info.

PR #265, opened 2022-06-30
Update Bootstrap core CSS

ThibautPV:
Update with last stable version (5.1.3)


MGatner (changes requested):
Extra white space.

MGatner:
I think this is good to do, but we should probably also talk about non-CDN options at some point.

MGatner (approved):

datamweb:
@ThibautPV Thanks and welcome.

Issue #260, opened 2022-06-28
Validation rules for username


kenjis:
The current validation rules for username permit spaces.

https://github.com/codeigniter4/shield/blob/0e1f3780435500d1ca6c030bcff17cd634cd0521/src/Database/Migrations/2020-12-28-223112_create_auth_tables.php#L14

https://github.com/codeigniter4/shield/blob/0e1f3780435500d1ca6c030bcff17cd634cd0521/src/Controllers/RegisterController.php#L134

https://github.com/codeigniter4/shield/blob/0e1f3780435500d1ca6c030bcff17cd634cd0521/src/Models/UserModel.php#L121

> I'm speaking anecdotally, I don't have an RFC reference or anything, but I would never anticipate allowing a space in a username and I also know many services that allow periods, especially so you can use Gmail-style credentials: `matthew.gatner@gmail.com => matthew.gatner`
https://github.com/codeigniter4/shield/pull/242#discussion_r906804934

kenjis:
If we permit dots and don't allow spaces, the rule should be like these?

```
'required|max_length[30]|regex_match[/\A[a-zA-Z0-9\.]+\z/]|min_length[3]|is_unique[users.username]'
```

MGatner:
Thanks for opening this @kenjis. I will do a little research, as I expect there are some standards around this that would better serve us than my anecdotal preferences. 😊 Maybe @lonnieezell can chime in if spaces were intentionally supported for some reason.

MGatner:
Really great read here, not apparently on character validation but in taking some notes: https://www.b-list.org/weblog/2018/feb/11/usernames/

1. Username uniqueness should be case insensitive

kenjis:
I sent a PR #268 that has  config `$validationRulesUsername`.

kenjis:
We changed the validation rules. See #268

PR #268, opened 2022-07-02
fix: username validation rules

kenjis:
See #260

- change validation rules to `regex_match[/\A[a-zA-Z0-9\.]+\z/]`
- add `Config\AuthSession` and move validation rules to it
- refactor
- fix docs

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -141,7 +141,7 @@ $credentials = [
     'password' => $this->request->getPost('password')
 ];
 
-$validCreds? = auth()->check($credentials);
+$validCreds = auth()->check($credentials);
 
 if (! $validCreds->isOK()) {
     return redirect()->back()->with('error', $loginAttempt->reason());
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -108,7 +108,7 @@ Shield has the following rules for registration:
 ];
 ```
 
-If you need a different set of rules for registration, you can specify them in your `Validation` configuration (**app\Config\Validation.php**) like:
+If you need a different set of rules for registration, you can specify them in your `Validation` configuration (**app/Config/Validation.php**) like:
 
 ```php
 //--------------------------------------------------------------------
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -95,8 +95,14 @@ Shield has the following rules for registration:
 
 ```php
 [
-    'username'         => 'required|alpha_numeric_space|min_length[3]|is_unique[users.username]',
-    'email'            => 'required|valid_email|is_unique[auth_identities.secret]',
+    'username'         => [
+            'required',
+            'max_length[30]',
+            'min_length[3]',
+            'regex_match[/\A[a-zA-Z0-9\.]+\z/]',
+            'is_unique[users.username]',
+        ],
+    'email'            => 'required|max_length[254]|valid_email|is_unique[auth_identities.secret]',
     'password'         => 'required|strong_password',
     'password_confirm' => 'required|matches[password]',
 ];
@@ -109,8 +115,14 @@ If you need a different set of rules for registration, you can specify them in y
 // Rules
 //--------------------------------------------------------------------
 public $registration = [
-    'username'         => 'required|alpha_numeric_space|min_length[3]|is_unique[users.username]',
-    'email'            => 'required|valid_email|is_unique[auth_identities.secret]',
+    'username'         => [
+            'required',
+            'max_length[30]',
+            'min_length[3]',
+            'regex_match[/\A[a-zA-Z0-9\.]+\z/]',
+            'is_unique[users.username]',
+        ],
+    'email'            => 'required|max_length[254]|valid_email|is_unique[auth_identities.secret]',
     'password'         => 'required|strong_password',
     'password_confirm' => 'required|matches[password]',
 ];
--- a/src/Config/Auth.php
+++ b/src/Config/Auth.php
@@ -225,6 +225,20 @@ class Auth extends BaseConfig
 
     /**
      * --------------------------------------------------------------------
+     * Validation Rules for username
+     * --------------------------------------------------------------------
+     *
+     * @var string[]
+     */
+    public array $validationRulesUsername = [
+        'required',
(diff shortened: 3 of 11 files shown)


MGatner (changes requested):
For consistency I would rather see this handled like RegistrationController:
* Move rules to a protected method
* Method checks `config('Validation')` first, falling back to our defaults
* Docs state how developers can change the rules using **app/Config/Validation.php**

I'm open to discussion if there's a reason you went with this approach instead.

kenjis:
> Move rules to a protected method

What do you mean? Which class?

MGatner:
> Which class?

My suggestions is to make `LoginController::getUsernameRules()` which returns what you have in the Config now. Then change this line in `getValidationRules()`:
```diff
-            //'username' => setting('Auth.validationRulesUsername'),
+            //'username' => config('Validation')->username,
```

If another method feels overkill it could be a property or just within the same method. But the ultimate goal is to have these rules overridable in Validation

kenjis:
@MGatner I've change the property name to `$_username` because it is not a normal validation group.

MGatner:
Why isn't it a normal validation group? Did I miss something obvious here...

MGatner:
Ohhh I get it, because it isn't actually a Validation *group*, just a partial rule. Okay, you're going to groan but I think I was wrong about this and the way you had it was better 😅

If we revert, do we need `Config\Auth::$usernameRules` though? Couldn't a developer still override that with `Config\Validation::$login`? I'm just trying to keep customization locations consolidated as much as possible.

kenjis:
A normal validation group is an array that has field name key and validation rules.
See https://codeigniter4.github.io/CodeIgniter4/libraries/validation.html#how-to-save-your-rules
```php
    public $signup = [
        'username'     => 'required',
        'password'     => 'required',
        'pass_confirm' => 'required|matches[password]',
        'email'        => 'required|valid_email',
    ];
```
But this property is:
```php
    public array $_username = [
        'required',
        'max_length[30]',
        'min_length[3]',
        'regex_match[/\A[a-zA-Z0-9\.]+\z/]',
    ];
```

MGatner:
I see we cross-posted there. Let me know what you think of reverting. Sorry for the trouble.

kenjis:
> Couldn't a developer still override that with Config\Validation::$login?

Yes, they can override it with Config\Validation::$login.

If you don't need Config\Validation::$_username, I'm okay with removing it.

MGatner:
Please remove. I think it is more confusing than helpful. My request for this was a mistake.

kenjis:
Removed `Config\Validation::$_username`.

lonnieezell:
While I like the idea of the static methods for this, leaving it in the LoginController seems like a strange place for them. 

We already removed the redirect methods and put them in the Auth config file. This seems like a very similar situation, and we end up using the LoginController as a configuration source for these rules, which is not how we typically want them to be used. 

Would it make sense to move these  static methods to either the Auth config file or the Validation config file?

MGatner:
I don't like using  `Shield\Config\Validation` because the class would be inaccessible via `config()`, so even if the methods were statically usable I fear it would be confusing. I do like the idea of making these methods on `Shield\Config\Auth` that are parallel to the redirect methods, but at that point we have come full circle and have a method returning a simple array - so why not just make it back into a property?? 😅

MGatner:
Let me also say: this Config file is getting huge. I wonder if we could split it out more, like we did with AuthGroups and AuthRoutes? A lot of the fields are specific to the Session handler - perhaps we could have a separate AuthSession file?

datamweb:
> separate AuthSession file?

I agree with this, I believe that this file will get bigger with time, so it is better to divide it into small parts now.

kenjis:
I created `Config\AuthSession` and move validation rules to it.

MGatner:
@datamweb are you suggesting we make those values configurable properties?

MGatner (approved):

datamweb:
> @datamweb are you suggesting we make those values configurable properties?

Yeah,Why?
Currently, 
```php  
public const ID_TYPE_MAGIC_LINK     = 'magic-link';
```
 are defined. I prefer to use `magic_link` so I have to edit the original file([Authenticators/Session.php](https://github.com/codeigniter4/shield/blob/77a047d8f18cc748a2eafc1c6e9135d23d48017a/src/Authentication/Authenticators/Session.php#L32-L37)).
Or for example, let's say I want to change 
```php  
public const ID_TYPE_MAGIC_LINK     = 'magic-link';
``` 
to
```php  
public const ID_TYPE_MAGIC_LINK     = 'magic-link-email';
``` 
and then create 
```php  
public const ID_TYPE_MAGIC_LINK_SMS     = 'magic-link-sms';
``` 
, having values configurable properties in the `Config\AuthSession` file will easily allow me to do that without having to manipulate the original shield file.

MGatner:
Understood! Just clarifying whether you meant:
1. Move these constants to be class constants on the Config class, or
2. Make these configurable properties of the Config class

You want 2. I'm not opposed but I'd like to hear @kenjis' input since he has a better idea of that scope of impact.

kenjis:
@datamweb Technically, of course it is possible.

But I don't know the benefit.
The values are values for ID Type and used in the database,
and tightly coupled with the Session Authenticator implementation.

If a dev changes the value in the Config file, the values in database are not updated automatically,
the system will not work correctly.

I understand the case you want to change the value `magic-link` to `magic-link-email`,
but if you want to add `ID_TYPE_MAGIC_LINK_SMS`, you must change the Session Authenticator implementation.
If you modify Authenticators/Session.php, why don't you add it in it?

datamweb:
> If a dev changes the value in the Config file, the values in database are not updated automatically,
> the system will not work correctly.

In this case, this suggestion will not be useful.

> If you modify `Authenticators/Session.php`, why don't you add it in it?

In general, I am against modify the main files, this will cause problems in the next update.

**Ok, thanks for the explanation.**

kenjis:
I think this PR is ready for merge.
Can I merge?

MGatner:
Yes, please do!

PR #273, opened 2022-07-02
docs: fix definition actions

datamweb:
**Hello friends**, we have already presented the definition of actions in this way:
```php
public array $actions = [
    'login'    => 'CodeIgniter\Shield\Authentication\Actions\Email2FA',
    'register' => 'CodeIgniter\Shield\Authentication\Actions\EmailActivator',
];
```
Or it can be as follows
```php
public array $actions = [
    'login'    => \CodeIgniter\Shield\Authentication\Actions\Email2FA::class,
    'register' => \CodeIgniter\Shield\Authentication\Actions\EmailActivator::class,
];
```
In any case, correction was needed.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -110,7 +110,7 @@ Turned off by default, Shield's Email-based 2FA can be enabled by specifying the
 
 ```php
 public array $actions = [
-    'login'    => \CodeIgniter\Shield\Authentication\Actions\Email2FA,
+    'login'    => 'CodeIgniter\Shield\Authentication\Actions\Email2FA',
     'register' => null,
 ];
 ```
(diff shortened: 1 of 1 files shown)


kenjis (approved):

MGatner (approved):
I prefer the shortname with `::class` to make the class names explicit (also ensures static analysis knows these are classes) but for docs this is probably good.

datamweb:
I also agree with shortname with `::class`, if other members also agree, I can apply changes in this PR based on what @MGatner said.

kenjis:
It is recommended to use `::class` always.

kenjis:
This is a bug in the docs. So I merged.
Welcome another PR to use `::class`.

PR #275, opened 2022-07-03
fix: UserModel::save() can't update User's email/password

### LINKED ISSUE #253: UserModel::save() can't save new User email/password
After calling `parate::save()`, `UserModel::save()` tries to save Email identity.
But when inserting User, `$this->find($this->db->insertID())` returns the User without email/passsword.

https://github.com/codeigniter4/shield/blob/4c8b53c9b6fd7701cba804e86df78bdeef48dc42/src/Models/UserModel.php#L205-L216

### LINKED ISSUE #254: UserModel::save() can't update User's email/password
When updating User's email/password, `parent::save($data)` may returns DataException (no data to update),
in that case, email/password is not updated.

https://github.com/codeigniter4/shield/blob/4c8b53c9b6fd7701cba804e86df78bdeef48dc42/src/Models/UserModel.php#L202-L231

kenjis:
Fixes #254
Fixes #253
- `UserModel::save()`, `update()` and `insert()` save Email Identity with Model Events
- fix `UserModel::findByCredentials()`
- fix `RegisterController::registerAction()`
- fix `User::saveEmailIdentity()`
- add `ValidationException`
- update docs/quickstart.md

### DIFF
--- a/src/Controllers/RegisterController.php
+++ b/src/Controllers/RegisterController.php
@@ -70,9 +70,7 @@ class RegisterController extends BaseController
             $user->username = null;
         }
 
-        if (! $users->save($user)) {
-            return redirect()->back()->withInput()->with('errors', $users->errors());
-        }
+        $users->saveWithEmailIdentity($user);
 
         // To get the complete user object with ID, we need to get from the database
         $user = $users->findById($users->getInsertID());
--- a/src/Controllers/RegisterController.php
+++ b/src/Controllers/RegisterController.php
@@ -57,9 +57,12 @@ class RegisterController extends BaseController
         }
 
         // Save the user
-        $allowedPostFields = array_merge(setting('Auth.validFields'), setting('Auth.personalFields'));
-        $user              = $this->getUserEntity();
-
+        $allowedPostFields = array_merge(
+            setting('Auth.validFields'),
+            setting('Auth.personalFields'),
+            ['password']
+        );
+        $user = $this->getUserEntity();
         $user->fill($this->request->getPost($allowedPostFields));
 
         // Workaround for email only registration/login
@@ -71,11 +74,8 @@ class RegisterController extends BaseController
             return redirect()->back()->withInput()->with('errors', $users->errors());
         }
 
-        // Get the updated user so we have the ID...
+        // To get the complete user object with ID, we need to get from the database
         $user = $users->findById($users->getInsertID());
-
-        // Store the email/password identity for this user.
-        $user->createEmailIdentity($this->request->getPost(['email', 'password']));
 
         // Add to default group
         $users->addToDefaultGroup($user);
--- a/src/Controllers/RegisterController.php
+++ b/src/Controllers/RegisterController.php
@@ -72,7 +72,7 @@ class RegisterController extends BaseController
         }
 
         try {
-            $users->saveWithEmailIdentity($user);
+            $users->save($user);
         } catch (ValidationException $e) {
             return redirect()->back()->withInput()->with('errors', $users->errors());
         }
--- a/src/Controllers/RegisterController.php
+++ b/src/Controllers/RegisterController.php
@@ -7,6 +7,7 @@ use CodeIgniter\Events\Events;
 use CodeIgniter\HTTP\RedirectResponse;
 use CodeIgniter\Shield\Authentication\Authenticators\Session;
 use CodeIgniter\Shield\Entities\User;
+use CodeIgniter\Shield\Exceptions\ValidationException;
 use CodeIgniter\Shield\Models\UserModel;
 
 /**
@@ -70,7 +71,11 @@ class RegisterController extends BaseController
             $user->username = null;
         }
 
-        $users->saveWithEmailIdentity($user);
+        try {
+            $users->saveWithEmailIdentity($user);
+        } catch (ValidationException $e) {
+            return redirect()->back()->withInput()->with('errors', $users->errors());
+        }
 
         // To get the complete user object with ID, we need to get from the database
         $user = $users->findById($users->getInsertID());
--- a/src/Entities/User.php
+++ b/src/Entities/User.php
@@ -2,6 +2,7 @@
 
 namespace CodeIgniter\Shield\Entities;
 
+use CodeIgniter\Database\Exceptions\DataException;
 use CodeIgniter\Entity\Entity;
 use CodeIgniter\Shield\Authentication\Authenticators\Session;
 use CodeIgniter\Shield\Authentication\Traits\HasAccessTokens;
@@ -159,7 +160,21 @@ class User extends Entity
         /** @var UserIdentityModel $identityModel */
         $identityModel = model(UserIdentityModel::class);
 
-        return $identityModel->save($identity);
+        try {
+            /** @throws DataException */
+            $identityModel->save($identity);
+        } catch (DataException $e) {
+            // There may be no data to update.
+            $messages = [
+                lang('Database.emptyDataset', ['insert']),
+                lang('Database.emptyDataset', ['update']),
+            ];
+            if (in_array($e->getMessage(), $messages, true)) {
+                return true;
+            }
+        }
+
+        return true;
     }
 
     /**
--- a/src/Entities/User.php
+++ b/src/Entities/User.php
@@ -172,6 +172,8 @@ class User extends Entity
             if (in_array($e->getMessage(), $messages, true)) {
                 return true;
             }
+
+            throw $e;
         }
 
         return true;
--- a/src/Entities/User.php
+++ b/src/Entities/User.php
@@ -122,7 +122,7 @@ class User extends Entity
     }
 
     /**
-     * If $user, $password, or $password_hash have been updated,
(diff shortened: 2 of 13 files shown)


MGatner (approved):
If this works then let's go with it but I still have plans to add the original method parameters to the event data in the framework, which would make this much easier.

MGatner:
Sorry, I didn't mean to hit "approve" - I'm good with the approach but obviously will need to pass tests 😂

datamweb:
> Sorry, I didn't mean to hit "approve" - I'm good with the approach but obviously will need to pass tests 😂

@MGatner this PR is in position ``Draft``. But you are in a hurry to improve Shield. This is great.

kenjis:
I removed the approval.

kenjis:
> If this works then let's go with it but I still have plans to add the original method parameters to the event data in the framework, which would make this much easier.

The enhancement is good, but we need to wait for v4.3. So I think this PR is needed.

kenjis:
Okay, I think this is ready for reviews.

kenjis:
I think `UserModel::save()` should be renamed.
1. It is confusing only `save()` is overridden and can save email/password. `insert()`/`update()` does not work.
2. Can't change the method signature.

MGatner:
@kenjis I agree. I think the ultimate solution is to modify `CodeIgniter\BaseModel::update()` (and probably `insert()`) to preserve the original `$data` parameter and pass that as an additional key into the `beforeInsert` and `afterInsert` methods. That would give Shield access to the User Entity without needing a separate database query and it wouldn't need to override any model methods.

kenjis:
> something like UserModel::saveWithIdentities(User $user) right?

Yes. I added a commit: https://github.com/codeigniter4/shield/pull/275/commits/057b22107ffeb56a967fa9bac70875245c5c4df6

src/Models/UserModel.php
```diff
@@ -185,48 +184,49 @@ public function activate(User $user): void
...
+     * Save User and its Email Identity (email, password, or password_hash fields)
+     * if they've been modified.
      */
-    public function save($data): bool
+    public function saveWithEmailIdentity(User $data): void
     {
         try {
+            /** @throws DataException */
             $result = parent::save($data);
-
```
lonnieezell, line 206: I'm confused why we are moving this to the `catch` block? We would only end up there if there was no other data on the user to save, right? If the user has also been updated with username, or they have custom fields on the user table, like first/last names, the exception would never be thrown and the email/password would never be saved.

kenjis: The `$user->saveEmailIdentity()` in the catch block is called when there is no data to insert/update to the User table.
If `parent::save()` is successful, `$user->saveEmailIdentity()` will be called after the catch block.

Both `parent::save()` and `$user->saveEmailIdentity()` may throw DataException, and we can't distinguish them. 
Therefore I put only `parent::save()` in the try block.

src/Entities/User.php
```diff
@@ -159,7 +160,23 @@ public function saveEmailIdentity(): bool
         /** @var UserIdentityModel $identityModel */
         $identityModel = model(UserIdentityModel::class);
 
-        return $identityModel->save($identity);
+        try {
+            /** @throws DataException */
+            $identityModel->save($identity);
```
MGatner, line 165: What happened to `allowEmpty()`? Did that go to 4.3?

src/Controllers/RegisterController.php
```diff
@@ -57,26 +57,24 @@ public function registerAction(): RedirectResponse
...
 
         // Workaround for email only registration/login
         if ($user->username === null) {
             $user->username = null;
         }
 
-        if (! $users->save($user)) {
-            return redirect()->back()->withInput()->with('errors', $users->errors());
-        }
+        $users->saveWithEmailIdentity($user);
```
MGatner, line 73: This makes me a little nervous that we rely solely on the Auth validation rules. A developer could provide their own UserModel with separate validation rules that causes the `save()` to fail but we don't catch the errors.

src/Entities/User.php
kenjis: Yes, it is 4.3.

src/Controllers/RegisterController.php
kenjis: Okay, you have a valid point. I'll change to catch.

kenjis: Added ValidationException and catch it.

src/Models/UserModel.php
```diff
@@ -185,48 +184,51 @@ public function activate(User $user): void
...
-     * fields if they've been modified.
-     *
-     * @param array|User $data
-     *
-     * @TODO can't change the return type to void.
+     * Save User and its Email Identity (email, password, or password_hash fields)
+     * if they've been modified.
      */
-    public function save($data): bool
+    public function saveWithEmailIdentity(User $data): void
```
lonnieezell, line 194: Did we need to rename this for a specific reason? If so, or the team majority says we should then ok, but I'm in favor of keeping it as an override of `Model::save`. 

I understand this is a more specific naming, but the whole purpose of the override was to make it as seamless as possible for the developer so they didn't have to think too much about it. This forces them to remember a new method for that single model, which is inconvenient and, as far as I can tell, unnecessary.

lonnieezell: Reading back through I see the discussion between you two so I'm overruled here. Just going on the record to say I disagree :)

MGatner: 👹 Contention noted!

I think this does two good things for us:
1. It lets us still use `save()` without messing with identities.
2. It allows us to force the `User` type parameter which makes this method much easier to handle. 

Also just to say it again: this should just be a temporary measure until we get better Model event parameters in place.

MGatner (approved):
LGTM!

src/Models/UserModel.php
kenjis: I overrode `save()` again.

After all, if save() is going to be used in the future, it is better for users to use save() now.

It is confusing only save() can save Email Identity, but users do not need to use insert() or update().

MGatner: Alternative idea...

The whole reason we override this method is that `afterInsert` and `afterUpdate` receive the data as an array, not the original `User` entity, and that array is filtered by `$allowedFields` so the credentials are stripped. What if we override `insert()` and `update()` to store credentials in a new private property before calling `parent::method()` and then still use the `after____` methods to handle the identities, just using the stored versions instead?

datamweb:
hi @manukieli,
Please see #280.

src/Models/UserModel.php
kenjis: Added `UserModel::saveEmailIdentity()` and the Model Events call it when afterInsert and afterUpdate.

kenjis:
I rewrote UserModel::save() (and update(), insert()) with the Model Events.
Review please.

src/Models/UserModel.php
```diff
@@ -185,48 +192,107 @@ public function activate(User $user): void
...
-
-        $this->checkQueryReturn($return);
+        $this->save($user);
     }
 
     /**
-     * Override the BaseModel's `save()` method to allow
-     * updating of user email, password, or password_hash
-     * fields if they've been modified.
+     * @param User $data
```
MGatner, line 199: I think we should still support arrays, and the callback events would just ignore them.

MGatner (changes requested):
Thanks for taking a pass at this @kenjis! It didn't end up being quite as clean as I hoped but I still think it is a big improvement. I wish we had the `allowEmpty()` method, but what can you do 🤷‍♂️

One of the strengths to this approach is that we can support "normal" model operations instead of relying on special use cases with `save()`. To that end I think that we should still support array parameters, and just have the callback events skip identity updates whenever there is no `$tempUser`. I'm happy to hear an alternate opinion but I think that improves the transparency of these methods, which I believe was @lonnieezell's original intent.

src/Models/UserModel.php
kenjis: Frankly, I am not sure why it is necessary to accept arrays, but I implemented it that way to try it out.

src/Models/UserModel.php
```diff
@@ -185,48 +192,111 @@ public function activate(User $user): void
...
+            $user->password_hash = $this->tempUser->password_hash ?? '';
+
+            $user->saveEmailIdentity();
+
+            return $data;
+        }
+
+        // Update
+        $this->tempUser->saveEmailIdentity();
+
```
MGatner, line 299: I know `$tempUser` is reset on each call but I would feel better if it were also reset after the identity update, unless there is some logical reason to keep it around. 
```suggestion
            $this->tempUser = $user;
        }

        // Update and reset
        $this->tempUser->saveEmailIdentity();
        $this->tempUser = null;
```

MGatner (changes requested):

src/Models/UserModel.php
kenjis: Added `$this->tempUser = null`.

MGatner (approved):
This one has been a long road! Thanks for all your work on it @kenjis. I believe it's ready!

kenjis:
@MGatner Thank you for the review!

PR #278, opened 2022-07-03
docs: update out-of-dated sample code for `syncGroups()`

kenjis:


datamweb:
@kenjis , I had made this mistake once before.
In your opinion, supporting both modes (`string `or `array`) of parameters is not better?

eg: 

```php
$user->syncGroups(['admin', 'beta']);
$user->syncGroups('admin', 'beta');

$user->inGroup(['superadmin', 'admin']);
$user->inGroup('superadmin', 'admin');

$user->addPermission(['users.create', 'users.edit']);
$user->addPermission('users.create', 'users.edit');
```

kenjis:
See https://github.com/codeigniter4/shield/blob/d234619a688f99d1f6cff3b1c08b9d5f63fb6bb9/src/Authorization/Traits/Authorizable.php#L83
Now it is typed as `string`. You don't mistake if you are using IDE.
Screenshot 2022-07-04 13 25 51

kenjis:
In my opinion, supporting both types brings unnecessary complexity in the implementation.

If you really love arrays, you can write like this:
```php
$user->syncGroups(...['admin', 'beta']);
```

datamweb:
Thanks @kenjis , it was very interesting.

jozefrebjak:
@kenjis  Please add real world example with form input data , till the update we synced groups like this:

```php
// Save the user's groups
$user->syncGroups($this->request->getPost('groups') ?? []);
```

What is the best way to handle this now ?

kenjis:
It causes an error when `$this->request->getPost('groups')` returns null.

jozefrebjak:
> It causes an error when `$this->request->getPost('groups')` returns null.

So we need to use it like this:

```php
 $user->syncGroups(...$this->request->getPost('groups') ?? []);
```

MGatner:
> supporting both modes (`string` or `array`) of parameters is not better?

One of the greatest strengths of variadic parameters is that they can be strongly-typed, which is especially appealing since PHP cannot specify collective types like "an array of strings" without PHPDoc `string[]` or `array<string>`. Enforcing the variadic version will help IDEs (as kenjis noted) and static analysis, and catch mistakes that might be overlooked otherwise.

MGatner:
> So we need to use it like this

@jozefrebjak I don't think that will work because the splat operator will try to resolve before the null coalescence. I would just use an intermediate variable, though if you really wanted a single line you might be able to use parentheses to specify the ordering.

MGatner:
Actually I just tried it an it does work! https://3v4l.org/G3qOZ

Who knew 🤭

MGatner (approved):

kenjis:
This works without any problem:
```php
$user->syncGroups(...$this->request->getPost('groups') ?? []);
```
But I feel it is a bit long and difficult to read.
I prefer to use an intermediate variable:
```php
$groups = $this->request->getPost('groups') ?? [];
$user->syncGroups(...groups);
```

PR #286, opened 2022-07-06
fix: Exception structure

kenjis:
This PR changes the hierarchy of Exceptions for Shield as follows:

- BaseException
   - LogicException
      - InvalidArgumentException
   - RuntimeException
      - AuthenticationException
      - AuthorizationException

### DIFF
--- a/src/Exceptions/BaseException.php
+++ b/src/Exceptions/BaseException.php
@@ -1,0 +1,12 @@
+<?php
+
+namespace CodeIgniter\Shield\Exceptions;
+
+use Throwable;
+
+/**
+ * Base Exception Interface for Shield
+ */
+interface BaseException extends Throwable
+{
+}
--- a/src/Authentication/AuthenticationException.php
+++ b/src/Authentication/AuthenticationException.php
@@ -2,9 +2,10 @@
 
 namespace CodeIgniter\Shield\Authentication;
 
-use CodeIgniter\Shield\Exceptions\HTTPException;
+use CodeIgniter\HTTP\Exceptions\HTTPException;
+use CodeIgniter\Shield\Exceptions\RuntimeException;
 
-class AuthenticationException extends HTTPException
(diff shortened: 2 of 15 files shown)


src/Exceptions/BaseException.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace CodeIgniter\Shield\Exceptions;
+
+use Throwable;
+
+/**
+ * Base Exception Interface for Shield
+ */
+interface BaseException extends Throwable
```
MGatner, line 10: I didn't know you could extend `Throwable`, that's rather handy.

MGatner (approved):
Looks good! Your desired hierarchy would work if your PR to the framework for exception interfaces goes through. To that end we should consider making these exception classes `@internal` or `final` or both (which is probably a good idea anyway).

kenjis:
`Shield\Exceptions\HTTPException` was not needed.
Since `CodeIgniter\HTTPExceptionsExceptions\HTTPException` was used in `AuthenticationException`,
I thought we needed an exception to wrap it. However, a closer look at the code showed that it only took the caught `CodeIgniter\HTTPExceptions\HTTPException` and converted it into a new exception.

So I removed `Shield\Exceptions\HTTPException`.

But I found `AuthenticationException` has some `LogicException`.

MGatner:
Just out of curiosity what is the advantage of implementing our own version if `InvalidArgumentException`? Logic errors by definition is code that needs fixing so shouldn't be caught, which makes me think that adding them to `BaseException` means someone might accidentally suppress a logic error by catching all `BaseException`.

kenjis:
The reason for placing all exceptions under BaseException is to make it easier to know where the exception came from.

It is the developer's responsibility to decide how to handle the exception.

If a dev catches `BaseException`, it seems they want to catch all exceptions by Shield.

PR #287, opened 2022-07-06
fix : Bootstrap column on mobile

### LINKED ISSUE #271: Display Form on mobile phones
There is a problem with the display of forms on cell phones. (see attached images)

Indeed, it should be modified : 
- class="card col-5 shadow-sm"

With: 
- class="card col-12 col-md-5 shadow-sm"

The views "login", "register", "magic_link_message" and "magic_link_form" are concerned

Before :
Capture d’écran 2022-07-02 à 18 21 22

After :
Capture d’écran 2022-07-02 à 18 21 52

I also wonder if it would be necessary to add a "class="row"" div after the container?

ThibautPV:
Fixes #271

### DIFF
--- a/src/Views/login.php
+++ b/src/Views/login.php
@@ -5,7 +5,7 @@
 <?= $this->section('main') ?>
 
     <div class="container d-flex justify-content-center p-5">
-        <div class="card col-5 shadow-sm">
+        <div class="card col-12 col-md-5 shadow-sm">
             <div class="card-body">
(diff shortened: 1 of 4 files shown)


ThibautPV:
Hi @MGatner and thank you for the approval

Indeed, I also think that it is possible to reduce this duplicated code, but I don't know what would be the best way to do it 

In the "layout.php" file maybe?

kenjis (approved):
Thank you!
LGTM!

PR #293, opened 2022-07-11
style: run php-cs-fixer

kenjis:
fix coding style.

It seems php-cs-fixer version up causes GA error:
https://github.com/codeigniter4/shield/runs/7279229470?check_suite_focus=true


MGatner (commented):
Smallest PR ever 🤗

PR #298, opened 2022-07-16
Fix: register when logged in

### LINKED ISSUE #294: Registration form still accessible when user is logged in
Let's say I use the registration form to sign up: Once I've registered, I'm automatically logged in.
But then the registration form is still accessible, and I can register again and again (which creates new users and leads to an error every time).

My point is: Once a user is logged in, shouldn't the register form be closed?

I would suggest to add the following code in [src/Controllers/RegisterController.php ](https://github.com/codeigniter4/shield/blob/develop/src/Controllers/RegisterController.php#L29), at the beginning of the `registerView()` method:

```php
    ...
    public function registerView()
    {
        if (auth()->loggedIn()) {
            return redirect()->to(config('Auth')->registerRedirect());
        }

        // Check if registration is allowed
        ...
```
I can make a PR if you agree with that (actually I don't know how to proceed: should I just create an issue or directly make a PR?).

parisiam:
Fixes #294

- Prevents a logged in user to register
- A corresponding unit test was added

### DIFF
--- a/src/Controllers/RegisterController.php
+++ b/src/Controllers/RegisterController.php
@@ -27,6 +27,10 @@ class RegisterController extends BaseController
      */
     public function registerView()
     {
+        if (auth()->loggedIn()) {
+            return redirect()->to(config('Auth')->registerRedirect());
+        }
+
         // Check if registration is allowed
         if (! setting('Auth.allowRegistration')) {
             return redirect()->back()->withInput()
--- a/src/Controllers/RegisterController.php
+++ b/src/Controllers/RegisterController.php
@@ -45,6 +45,10 @@ class RegisterController extends BaseController
      */
     public function registerAction(): RedirectResponse
     {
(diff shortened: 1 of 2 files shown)


MGatner (changes requested):
Can we get the same on `registerAction()`?

parisiam:
Yes, sure, happy to help ;)
I've submitted the changes for registerAction() and for the tests.
Please check the tests (I may have be redundant or incomplete).

kenjis:
What if a user is pending login state (in the process of 2FA)?
See  #132

MGatner:
I think that's fine. They may have mistyped their email address, or want multiple accounts. In my opinion this is mostly about preventing confusion, rather than a security or behavior issue.

MGatner (approved):

kenjis (approved):
Okay, thanks!

PR #308, opened 2022-07-21
fix: UserModel::findById() does not work with withIdentities()

### LINKED ISSUE #307: UserModel::findById() shows Error
When trying to find user by id  with identities return error "Attempt to read property \"id\" on int" in "vendor/codeigniter4/shield/src/Models/UserModel.php" on line 92

`$profile = (new \CodeIgniter\Shield\Models\UserModel())->withIdentities()->findById($Id);`

But if I use :
`$profile = (new \CodeIgniter\Shield\Models\UserModel())->where('id',$Id)->withIdentities()->findAll($Id);` 

It return the user record with the identity record duplicated. 

Am I missing something?
Regards

kenjis:
Fixes #307


MGatner (approved):
It's a little hard to see what's happening on mobile but the test case looks good and reporter confirms the fix.

PR #315, opened 2022-07-25
phpDocs: modify the action type

datamweb:


MGatner (approved):
Good catch!

src/Controllers/RegisterController.php
```diff
@@ -98,7 +98,7 @@ public function registerAction(): RedirectResponse
 
         $authenticator->startLogin($user);
 
-        // If an action has been defined for login, start it up.
+        // If an action has been defined for register, start it up.
```
kenjis, line 101: @datamweb It is not phpDoc. Just a comment.
PHPDocs must start with `/**` and end with `*/`.
See https://www.php.net/manual/en/reflectionclass.getdoccomment.php

datamweb: @kenjis , I am an electrical engineer. I became interested in PHP many years ago and tried to learn it.
During this time, your presence made me move from a static state to a forward state in programming.
I thank you very much.
And I'm sorry that I usually surprise you.

kenjis: This is not particularly surprising!
It is hard to imagine that there is a special comment style called Doc comment.

PR #321, opened 2022-07-27
Add german translation

sba:
For German-speaking people: In der Übersetzung sieze ich den Benutzer. Wäre duzen besser? Oder ohne direkte Anrede, also im Sinn von "Bitte E-Mail Adresse bestätigen" anstelle von "Bestätigen Sie Ihre E-Mail Adresse" oder "Bestätige deine E-Mail Adresse'?


sba:
> Please add test code.

ok, I added the test-file: https://github.com/codeigniter4/shield/pull/322

datamweb:
One thing at a time: **a pull request should only contain one change. That does not mean only one commit,** but one change - however many commits it took. The reason for this is that if you change X and Y but send a single pull request for both at the same time, we might really want X but disagree with Y, meaning we cannot merge the request. Using the Git-Flow branching model you can create new branches for both of these features and send two requests.
more info [see ](https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/workflow.md#branching)

sba:
@[datamweb](https://github.com/datamweb) thanks for your comment - I added the test changes to this commit, I hope it's ok now.

kenjis:
@sba 
Please fix the coding style.
Run `composer cs-fix`.

See https://github.com/codeigniter4/shield/runs/7555124912?check_suite_focus=true

MGatner (approved):
LGTM! Danke

kenjis (approved):
Thank you!

PR #322, opened 2022-07-27
Create GermanTranslationTest.php

sba:


datamweb:
https://github.com/codeigniter4/shield/blob/00023eee70d65c17cb9bd840a174885feb15bebd/tests/Language/AbstractTranslationTestCase.php#L51 **Line 51** is comment or uncomment?(more info [see](https://www.php.net/manual/en/language.basic-syntax.comments.php))

kenjis:
We can't see the test result, because there is no translations.
This commit should be included in  #321

PR #324, opened 2022-07-27
feat: add allow to cancel actions for custom cases

datamweb:
Hello friends.
The purpose of these PR,
Add a feature to cancel actions if there are custom cases. 
This PR allows the shield to cancel actions by checking the user `groups`,`permissions `or `userID`.

Task checklist:
- [x] Add to docs
- [x] Edit `Session::class`
- [ ] Add test

### DIFF
--- a/docs/auth_actions.md
+++ b/docs/auth_actions.md
@@ -2,7 +2,9 @@
 
 - [Authentication Actions](#authentication-actions)
   - [Configuring Actions](#configuring-actions)
+  - [Cancel Actions For Custom Cases](#cancel-actions-for-custom-cases)
   - [Defining New Actions](#defining-new-actions)
+
 
 Authentication Actions are a way to group actions that can happen after login or registration.
 Shield ships with two actions you can use, and makes it simple for you to define your own.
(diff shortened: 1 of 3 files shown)


datamweb:
Guys, if you agree with these **PR** in general, please let me know so I can complete the test as well.

kenjis:
Why do you need this feature?
Use cases?

datamweb:
- The ability to manage resources (management of server resources for sending emails, managing financial resources for sending SMS fees)
- Non-engagement of unfamiliar users with new topics (2fa)

Two-step login for users with low access level is **not necessary** in many cases and it causes **financial cost or filling the ceiling of email sending limits**, but it **is very important** for **users with high access level even** though it is expensive.
Therefore, by canceling actions for groups with low access level, it has helped a lot in this regard.

I believe that the actions in the shield are very useful, but they need to be changed. for example
Why don't the codes sent by actions expire?
Why is it not possible to define multiple actions for one type of action?
Why is it not possible to order the length of the sent code (usually 4 digits are used in SMS)
Why don't we have an action called `Google2FA`, even though it is very popular?
All these things can be implemented, but why does the shield not support them by default?

datamweb:
I will close this issue, if a user needed this PR, we should have received a feedback by now.

kenjis:
I think if an action is to be disabled only for specific users, it is better to make it configurable for that action, rather than disabling all actions.

datamweb:
That's right, so each user can write their own [custom action](https://github.com/datamweb/shield/blob/develop/docs/auth_actions.md#defining-new-actions) if needed.

PR #340, opened 2022-08-05
chore: update composer cs-fix command

kenjis:
- Use cache. No cache is slow.
- Make `cs-fix` default. CodeIgniter4 repository does not have `style` command.


MGatner (approved):
If these are changes worth sharing then please submit to DevKit as well.

Issue #314, opened 2022-07-25
RememberMe: Unknown column 'users.remember' in 'where clause'


sammyskills:
Hi all,

I just tried to login with the remember-me checkbox 'checked', but it returned an error message:

```
mysqli_sql_exception #1054
Unknown column 'users.remember' in 'where clause'
```

kenjis:
Do you use the latest `develop` branch?
Can you show back trace?

There is no such column `users.remember`.
https://github.com/codeigniter4/shield/blob/121ac4a9c44bd67fcd1f9130e7119ab9b5477634/src/Database/Migrations/2020-12-28-223112_create_auth_tables.php#L13-L21

sammyskills:
Yes, I'm using the latest `develop` branch, and I checked the db also, that column doesn't exist. 
Here's the back trace:
![image](https://user-images.githubusercontent.com/8720569/180736290-bba84458-4184-45b6-8576-7cc799c3b78b.png)

kenjis:
What's your `$validFields` value in Config\Auth ?
https://github.com/codeigniter4/shield/blob/121ac4a9c44bd67fcd1f9130e7119ab9b5477634/src/Config/Auth.php#L222-L225

sammyskills:
![image](https://user-images.githubusercontent.com/8720569/180754248-68eccd19-b59d-4e79-8af4-bcc05eb118c0.png)

kenjis:
It seems you customized `LoginController::loginAction()`.
It seems the cause of the error.

kenjis:
Please `unset($credentials['remember'])`  before the following line.

```php
$result = auth('session')->remember($remember)->attempt($credentials);
```

sammyskills:
Yes, I customized the `loginAction()` method, but I don't expect it to cause any error. 
I tried to make the login form accept either username or email address. Here's my code:
```
public function loginAction(): RedirectResponse
{
    // The post request
    $post_request = $this->request->getPost();
        
    // Determine the credential type
    $login = $post_request['login'];
    $type = filter_var($login, FILTER_VALIDATE_EMAIL) ? 'email' : 'username';
    $type == 'email' ? $post_request['email'] = $login : $post_request['username'] = $login;
    // Unset the login array
    unset($post_request['login']);

    // Run validation as it isn't covered in the model
    if ( ! $this->validateData($post_request, 'login_'.$type) ) return redirect()->back()->withInput()->with('errors', $this->validator->getErrors());
        
[...]

PR #344, opened 2022-08-05
chore: add issue templete

kenjis:
Copied from CodeIgniter4 repository, and tweaked.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -18,16 +18,12 @@ body:
 
         ---
 
-  - type: dropdown
+  - type: input
     id: php-version
     attributes:
       label: PHP Version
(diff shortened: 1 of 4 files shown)


datamweb:
Considering that Shield allows customizable in different parts, it seems that you have made an option similar to which part of customizable? It can be useful.
See [comment](https://github.com/codeigniter4/shield/issues/314#issuecomment-1193868437) . 
In the PHP version section, it can be helpful to enter the exact version, for example, 7.4.3, see #319.

kenjis:
Thanks for the good comment. I updated.

MGatner (changes requested):
You have a few changes that appear to be from other branches. The templates look great though, thanks for setting this up!

kenjis:
Removed irrelevant commits.

MGatner (approved):

PR #345, opened 2022-08-05
Fix email deliverability error log

### LINKED ISSUE #330: Emails do not deliver
Hi all,

Has anyone tried to use any of the auth actions (Email2FA or EmailActivator) and got an email delivered? 

I've tried severally using gmail's smpt, cpanel email, etc, none delivered. I only got the message:
```
Cannot send email for user: kuqib@mailinator.com
```

First, the code does not help in knowing the exact error as one can not see the actual problem with `printDebugger()`: https://github.com/codeigniter4/shield/blob/835bec6a6bfde617f5f5c3f14935e8b61282c20b/src/Authentication/Actions/EmailActivator.php#L50-L52 and when I tried to edit the code to show the exact error message, I got this `EXIM` error:
```
We do not authorize the use of this system to transport unsolicited,  and/or bulk e-mail
```
Secondly, I recently tried using phpmailer (by extending the default activator class) with the same email settings and mails delivered successfully. But then, I couldn't get the code to be sent as it is called from a private method:
https://github.com/codeigniter4/shield/blob/835bec6a6bfde617f5f5c3f14935e8b61282c20b/src/Authentication/Actions/EmailActivator.php#L104-L120

So, here's what I'm proposing:
* Can we have the default auth actions display or log the exact email error if it fails, for proper debugging?
* Can we also make the `createIdentity()` method a protected method instead of a private method so it can be used by child classes?

What do you think guys? I can make a PR if that's okay.

sammyskills:
Fixes #330

In this commit, 
* I updated the code to show a more detailed error if an email is not sent

### DIFF
--- a/src/Authentication/Actions/Email2FA.php
+++ b/src/Authentication/Actions/Email2FA.php
@@ -80,6 +80,9 @@ class Email2FA implements ActionInterface
             throw new RuntimeException('Cannot send email for user: ' . $user->email . "\n" . $email->printDebugger(['headers']));
         }
 
+        // Clear the email
+        $email->clear();
+
         return view(setting('Auth.views')['action_email_2fa_verify']);
     }
 
--- a/src/Authentication/Actions/Email2FA.php
+++ b/src/Authentication/Actions/Email2FA.php
@@ -77,7 +77,7 @@ class Email2FA implements ActionInterface
         $email->setMessage(view(setting('Auth.views')['action_email_2fa_email'], ['code' => $identity->secret]));
 
         if ($email->send(false) === false) {
-            throw new RuntimeException('Cannot send email for user: ' . $user->email . ' ' . $email->printDebugger(['headers']));
+            throw new RuntimeException('Cannot send email for user: ' . $user->email . '\n' . $email->printDebugger(['headers']));
         }
 
         return view(setting('Auth.views')['action_email_2fa_verify']);
@@ -114,7 +114,7 @@ class Email2FA implements ActionInterface
         $this->createIdentity($user);
     }
 
-    protected function createIdentity(User $user): void
+    private function createIdentity(User $user): void
     {
         /** @var UserIdentityModel $identityModel */
         $identityModel = model(UserIdentityModel::class);
--- a/src/Authentication/Actions/Email2FA.php
+++ b/src/Authentication/Actions/Email2FA.php
@@ -71,14 +71,13 @@ class Email2FA implements ActionInterface
 
         // Send the user an email with the code
         helper('email');
-        $return = emailer()->setFrom(setting('Email.fromEmail'), setting('Email.fromName') ?? '')
-            ->setTo($user->email)
-            ->setSubject(lang('Auth.email2FASubject'))
-            ->setMessage(view(setting('Auth.views')['action_email_2fa_email'], ['code' => $identity->secret]))
-            ->send();
+        $email = emailer()->setFrom(setting('Email.fromEmail'), setting('Email.fromName') ?? '');
+        $email->setTo($user->email);
+        $email->setSubject(lang('Auth.email2FASubject'));
+        $email->setMessage(view(setting('Auth.views')['action_email_2fa_email'], ['code' => $identity->secret]));
 
-        if ($return === false) {
-            throw new RuntimeException('Cannot send email for user: ' . $user->email);
+        if ($email->send(false) === false) {
+            throw new RuntimeException('Cannot send email for user: ' . $user->email . ' '. $email->printDebugger(['headers']));
         }
 
         return view(setting('Auth.views')['action_email_2fa_verify']);
@@ -115,7 +114,7 @@ class Email2FA implements ActionInterface
         $this->createIdentity($user);
     }
 
-    private function createIdentity(User $user): void
+    protected function createIdentity(User $user): void
     {
         /** @var UserIdentityModel $identityModel */
         $identityModel = model(UserIdentityModel::class);
--- a/src/Authentication/Actions/Email2FA.php
+++ b/src/Authentication/Actions/Email2FA.php
@@ -77,7 +77,7 @@ class Email2FA implements ActionInterface
         $email->setMessage(view(setting('Auth.views')['action_email_2fa_email'], ['code' => $identity->secret]));
 
         if ($email->send(false) === false) {
-            throw new RuntimeException('Cannot send email for user: ' . $user->email . ' '. $email->printDebugger(['headers']));
+            throw new RuntimeException('Cannot send email for user: ' . $user->email . ' ' . $email->printDebugger(['headers']));
         }
 
         return view(setting('Auth.views')['action_email_2fa_verify']);
--- a/src/Authentication/Actions/EmailActivator.php
+++ b/src/Authentication/Actions/EmailActivator.php
@@ -50,6 +50,9 @@ class EmailActivator implements ActionInterface
             throw new RuntimeException('Cannot send email for user: ' . $user->email . "\n" . $email->printDebugger(['headers']));
         }
 
+        // Clear the email
+        $email->clear();
+
         // Display the info page
         return view(setting('Auth.views')['action_email_activate_show'], ['user' => $user]);
     }
--- a/src/Authentication/Actions/EmailActivator.php
+++ b/src/Authentication/Actions/EmailActivator.php
@@ -41,14 +41,13 @@ class EmailActivator implements ActionInterface
 
         // Send the email
         helper('email');
-        $return = emailer()->setFrom(setting('Email.fromEmail'), setting('Email.fromName') ?? '')
-            ->setTo($userEmail)
-            ->setSubject(lang('Auth.emailActivateSubject'))
-            ->setMessage(view(setting('Auth.views')['action_email_activate_email'], ['code' => $code]))
-            ->send();
+        $email = emailer()->setFrom(setting('Email.fromEmail'), setting('Email.fromName') ?? '');
+        $email->setTo($userEmail);
+        $email->setSubject(lang('Auth.emailActivateSubject'));
+        $email->setMessage(view(setting('Auth.views')['action_email_activate_email'], ['code' => $code]));
 
-        if ($return === false) {
-            throw new RuntimeException('Cannot send email for user: ' . $user->email);
+        if ($email->send(false) === false) {
+            throw new RuntimeException('Cannot send email for user: ' . $user->email . ' '. $email->printDebugger(['headers']));
         }
 
         // Display the info page
@@ -101,7 +100,7 @@ class EmailActivator implements ActionInterface
         $this->createIdentity($user);
     }
 
-    private function createIdentity(User $user): string
+    protected function createIdentity(User $user): string
     {
         /** @var UserIdentityModel $identityModel */
         $identityModel = model(UserIdentityModel::class);
--- a/src/Authentication/Actions/EmailActivator.php
+++ b/src/Authentication/Actions/EmailActivator.php
@@ -47,7 +47,7 @@ class EmailActivator implements ActionInterface
         $email->setMessage(view(setting('Auth.views')['action_email_activate_email'], ['code' => $code]));
 
         if ($email->send(false) === false) {
-            throw new RuntimeException('Cannot send email for user: ' . $user->email . ' ' . $email->printDebugger(['headers']));
+            throw new RuntimeException('Cannot send email for user: ' . $user->email . '\n' . $email->printDebugger(['headers']));
         }
 
         // Display the info page
@@ -100,7 +100,7 @@ class EmailActivator implements ActionInterface
         $this->createIdentity($user);
     }
 
-    protected function createIdentity(User $user): string
+    private function createIdentity(User $user): string
     {
         /** @var UserIdentityModel $identityModel */
         $identityModel = model(UserIdentityModel::class);
--- a/src/Authentication/Actions/EmailActivator.php
+++ b/src/Authentication/Actions/EmailActivator.php
@@ -47,7 +47,7 @@ class EmailActivator implements ActionInterface
         $email->setMessage(view(setting('Auth.views')['action_email_activate_email'], ['code' => $code]));
 
         if ($email->send(false) === false) {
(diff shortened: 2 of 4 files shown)


kenjis:
Thank you for sending PR!

Can you show a real example of the new error message?

sammyskills:
Here's a sample error message:

```
CRITICAL - 2022-08-05 16:28:03 --> Cannot send email for user: kuqib@mailinator.com, 220-business67.web-hosting.com ESMTP Exim 4.95 #2 Fri, 05 Aug 2022 11:27:56 -0400 
220-We do not authorize the use of this system to transport unsolicited, 
220 and/or bulk e-mail.
<br /><pre>hello: 250-business67.web-hosting.com Hello localhost [197.242.98.221]
250-SIZE 52428800
250-8BITMIME
250-PIPELINING
250-PIPE_CONNECT
250-AUTH PLAIN LOGIN
250 HELP
</pre><pre>from: 250 OK
</pre><pre>to: 550 "Your IP: 197.242.98.221 : Your domain  is not allowed in header From"
</pre>The following SMTP error was encountered: 550 "Your IP: 197.242.98.221 : Your domain  is not allowed in header From"
<br /><pre>quit: 221 business67.web-hosting.com closing connection
</pre>Unable to send email using PHP SMTP. Your server might not be configured to send mail using this method.<br />220-business67.web-hosting.com ESMTP Exim 4.95 #2 Fri, 05 Aug 2022 11:27:59 -0400 
220-We do not authorize the use of this system to transport unsolicited, 
220 and/or bulk e-mail.
<br /><pre>hello: 250-business67.web-hosting.com Hello localhost [197.242.98.221]
250-SIZE 52428800
250-8BITMIME
250-PIPELINING
250-PIPE_CONNECT
250-AUTH PLAIN LOGIN
250 HELP
</pre><pre>from: 250 OK
</pre><pre>to: 550 "Your IP: 197.242.98.221 : Your domain  is not allowed in header From"
</pre>The following SMTP error was encountered: 550 "Your IP: 197.242.98.221 : Your domain  is not allowed in header From"
<br /><pre>quit: 221 business67.web-hosting.com closing connection
</pre>Unable to send email using PHP SMTP. Your server might not be configured to send mail using this method.<br />220-business67.web-hosting.com ESMTP Exim 4.95 #2 Fri, 05 Aug 11:28:02 -0400 
220-We do not authorize the use of this system to transport unsolicited, 
220 and/or bulk e-mail.
<br /><pre>hello: 250-business67.web-hosting.com Hello localhost [197.242.98.221]
250-SIZE 52428800
250-8BITMIME
250-PIPELINING
250-PIPE_CONNECT
250-AUTH PLAIN LOGIN
250 HELP
</pre><pre>from: 250 OK
</pre><pre>to: 550 "Your IP: 197.242.98.221 : Your domain  is not allowed in header From"
</pre>The following SMTP error was encountered: 550 "Your IP: 197.242.98.221 : Your domain  is not allowed in header From"
<br /><pre>quit: 221 business67.web-hosting.com closing connection
</pre>Unable to send email using PHP SMTP. Your server might not be configured to send mail using this method.<br /><pre>Date: Fri, 05 Aug 16:27:53 +0100
From: &quot;PBO&quot; &lt;&gt;
Return-Path: &lt;&gt;
To: kuqib@mailinator.com
Subject: =?UTF-8?Q?Your=20activation=20code?=
Reply-To: &lt;&gt;
User-Agent: CodeIgniter
X-Mailer: CodeIgniter
X-Priority: 3 (Normal)
Message-ID: &lt;62e2aaffbfcce1.83381459&gt;
Mime-Version: 1.0

</pre>
in VENDORPATH\codeigniter4\shield\src\Authentication\Actions\EmailActivator.php on line 60.
 1 VENDORPATH\codeigniter4\shield\src\Controllers\ActionController.php(52): CodeIgniter\Shield\Authentication\Actions\EmailActivator->show()
 2 VENDORPATH\codeigniter4\shield\src\Controllers\ActionController.php(40): CodeIgniter\Shield\Controllers\ActionController->show()
 3 SYSTEMPATH\CodeIgniter.php(895): CodeIgniter\Shield\Controllers\ActionController->_remap('show')
 4 SYSTEMPATH\CodeIgniter.php(466): CodeIgniter\CodeIgniter->runController(Object(CodeIgniter\Shield\Controllers\ActionController))
 5 SYSTEMPATH\CodeIgniter.php(349): CodeIgniter\CodeIgniter->handleRequest(null, Object(Config\Cache), false)
 6 FCPATH\index.php(55): CodeIgniter\CodeIgniter->run()
 7 SYSTEMPATH\Commands\Server\rewrite.php(43): require_once('FCPATH\\index.php')
```

kenjis:
Thank you!

How about `MagicLinkController`?

src/Authentication/Actions/Email2FA.php
```diff
@@ -71,14 +71,13 @@ public function handle(IncomingRequest $request)
...
-        if ($return === false) {
-            throw new RuntimeException('Cannot send email for user: ' . $user->email);
+        $email = emailer()->setFrom(setting('Email.fromEmail'), setting('Email.fromName') ?? '');
+        $email->setTo($user->email);
+        $email->setSubject(lang('Auth.email2FASubject'));
+        $email->setMessage(view(setting('Auth.views')['action_email_2fa_email'], ['code' => $identity->secret]));
+
+        if ($email->send(false) === false) {
+            throw new RuntimeException('Cannot send email for user: ' . $user->email . ' ' . $email->printDebugger(['headers']));
         }
```
kenjis, line 81: Is it better to call `$email->clear()` here?

src/Authentication/Actions/Email2FA.php
```diff
@@ -71,14 +71,13 @@ public function handle(IncomingRequest $request)
...
-
-        if ($return === false) {
-            throw new RuntimeException('Cannot send email for user: ' . $user->email);
+        $email = emailer()->setFrom(setting('Email.fromEmail'), setting('Email.fromName') ?? '');
+        $email->setTo($user->email);
+        $email->setSubject(lang('Auth.email2FASubject'));
+        $email->setMessage(view(setting('Auth.views')['action_email_2fa_email'], ['code' => $identity->secret]));
+
+        if ($email->send(false) === false) {
+            throw new RuntimeException('Cannot send email for user: ' . $user->email . ' ' . $email->printDebugger(['headers']));
```
kenjis, line 80: How about this?

```suggestion
            throw new RuntimeException('Cannot send email for user: ' . $user->email . "\" . $email->printDebugger(['headers']));
```

MGatner:
Why is the scope change on `createIdentity()` needed here? Can that be a separate PR?

src/Authentication/Actions/Email2FA.php
sammyskills: I don't think so, since it is not a loop, and the `$email` variables are not being reinitialized. See [docs](https://codeigniter4.github.io/CodeIgniter4/libraries/email.html#clear).

src/Authentication/Actions/Email2FA.php
sammyskills: This is good. 
Thank you.

sammyskills:
> How about `MagicLinkController`?

 I think I totally missed that. I'll update the PR now.


> Why is the scope change on `createIdentity()` needed here? Can that be a separate PR?

Well, I felt since it is just a minor change, and since the issues are related, it can also be in the same PR. But, thanks for pointing it out. I'll create a separate PR for that.

src/Authentication/Actions/Email2FA.php
kenjis: The Email service is shared.
https://github.com/codeigniter4/shield/blob/44e65eff3043530e21337ffdb2d2f6c991a0e548/src/Helpers/email_helper.php#L44
If devs use the shared Email (`service('email')`) in their app, the state is not clear.

sammyskills: Does 'clearing' not mean the same as trying to reinitialise the email variables?

src/Authentication/Actions/Email2FA.php
sammyskills: Just tried this, but it is returning syntax error. Is this what you were trying to explain:
```
throw new RuntimeException('Cannot send email for user: ' . $user->email . '"\"' . $email->printDebugger(['headers']));
```

kenjis: Oh, sorry, I missed `n`. I mean `"\n"`.

sammyskills: Thank you.

src/Authentication/Actions/Email2FA.php
MGatner: If necessary, probably a call to `clear()` should happen in the `emailer()` helper function.

sammyskills: Done, thank you.

kenjis: No, it is not.
The issue is that a dev might use the Email object in their app after Shield send email.

`$email->send()` calls `clear()`. But you changed it to `$email->send(false)` that does not call `clear()`.
`emailer()` is a helper for Shield. It is not a framework helper, so a dev does not use it.

If a dev calls `service('email')`, the `clear()` you added in the `emailer()` is not called.
To begin with, `emailer()` calls `$email->initialize()` so you don't need to add `clear()`.

kenjis: `service('email')` or `Services::email()` returns a shared instance by default.
It might be not good design.

Getting a new instance in `emailer()` may be better.
```diff
--- a/src/Helpers/email_helper.php
+++ b/src/Helpers/email_helper.php
@@ -41,7 +41,7 @@ function emailer(array $overrides = []): Email
         }
 
         /** @var Email $email */
-        $email = service('email');
+        $email = service('email', false);
 
         return $email->initialize($config);
     }
```

sammyskills: Oh, thank you for the explanation.

kenjis (approved):
LGTM!

src/Authentication/Actions/Email2FA.php
MGatner: Using a single service makes testing harder, just need to be sure to inject the right class ahead of time. I think calling `clear()` before the final return is fine, but I also expect developers to handle this themselves. Any time `email` service is used there is a possibility that it is not in a "clean" state so should be initialized/cleaned. This is of course terrible design but will be fixed when `Mailer` is released.

kenjis: Calling `clear()` before the final return does not make sense, becaue `initialize()` calls `clear()`.

Then, calling `clear()` after `throw new RuntimeException()` is the only way.
Otherwise this PR is a breaking change.

src/Authentication/Actions/Email2FA.php
```diff
@@ -71,14 +71,13 @@ public function handle(IncomingRequest $request)
...
-
-        if ($return === false) {
-            throw new RuntimeException('Cannot send email for user: ' . $user->email);
+        $email = emailer()->setFrom(setting('Email.fromEmail'), setting('Email.fromName') ?? '');
+        $email->setTo($user->email);
+        $email->setSubject(lang('Auth.email2FASubject'));
+        $email->setMessage(view(setting('Auth.views')['action_email_2fa_email'], ['code' => $identity->secret]));
+
+        if ($email->send(false) === false) {
+            throw new RuntimeException('Cannot send email for user: ' . $user->email . '\n' . $email->printDebugger(['headers']));
```
kenjis, line 80: `'\n'` is literal `'\n'`. I mean a line break.

```suggestion
            throw new RuntimeException('Cannot send email for user: ' . $user->email . "\n" . $email->printDebugger(['headers']));
```

src/Authentication/Actions/EmailActivator.php
```diff
@@ -41,14 +41,13 @@ public function show(): string
...
-
-        if ($return === false) {
-            throw new RuntimeException('Cannot send email for user: ' . $user->email);
+        $email = emailer()->setFrom(setting('Email.fromEmail'), setting('Email.fromName') ?? '');
+        $email->setTo($userEmail);
+        $email->setSubject(lang('Auth.emailActivateSubject'));
+        $email->setMessage(view(setting('Auth.views')['action_email_activate_email'], ['code' => $code]));
+
+        if ($email->send(false) === false) {
+            throw new RuntimeException('Cannot send email for user: ' . $user->email . '\n' . $email->printDebugger(['headers']));
```
kenjis, line 50:
```suggestion
            throw new RuntimeException('Cannot send email for user: ' . $user->email . "\n" . $email->printDebugger(['headers']));
```

kenjis (changes requested):
See my two comments.

src/Authentication/Actions/Email2FA.php
sammyskills: Done, thanks.

src/Authentication/Actions/EmailActivator.php
sammyskills: Done, thank you.

src/Helpers/email_helper.php
```diff
@@ -41,7 +41,7 @@ function emailer(array $overrides = []): Email
         }
 
         /** @var Email $email */
-        $email = service('email');
+        $email = service('email', false);
```
kenjis, line 44: Let's follow MGater's opinion and undo this here.

```suggestion
        $email = service('email');
```

kenjis:
To keep the same behavior, please add ` $email->clear();`:
```php
        if ($email->send(false) === false) {
            throw new RuntimeException('Cannot send email for user: ' . $user->email . "\n" . $email->printDebugger(['headers']));
        }

        $email->clear();
```

src/Helpers/email_helper.php
sammyskills: So, is the email still going to be `clear()`ed ?

kenjis: See my comment below.

sammyskills: Ok. 
Done that now.

kenjis (approved):
Thank you!

MGatner (approved):
Looks good! Thanks everyone.

PR #346, opened 2022-08-06
Change the scope of `createIdentity()` method

sammyskills:
Sometimes developers might prefer to just extend the default Auth actions classes instead of creating a new one, and since the creation/generation of tokens is done in the `createIdentity()` methods, it is impossible to use it inside child classes as it is `private`. See:
https://github.com/codeigniter4/shield/blob/44e65eff3043530e21337ffdb2d2f6c991a0e548/src/Authentication/Actions/Email2FA.php#L118
https://github.com/codeigniter4/shield/blob/44e65eff3043530e21337ffdb2d2f6c991a0e548/src/Authentication/Actions/EmailActivator.php#L104

Ref #330

### DIFF
--- a/src/Authentication/Actions/Email2FA.php
+++ b/src/Authentication/Actions/Email2FA.php
@@ -115,7 +115,7 @@ class Email2FA implements ActionInterface
         $this->createIdentity($user);
     }
 
-    private function createIdentity(User $user): void
+    protected function createIdentity(User $user): void
     {
         /** @var UserIdentityModel $identityModel */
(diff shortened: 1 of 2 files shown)


MGatner (approved):
Fine by me! Maybe should consider making these `final` - see what @kenjis and @lonnieezell think.

kenjis:
Fine by me. Making these `final` might be better.

MGatner (approved):
Thanks! Looks good.

PR #347, opened 2022-08-07
Change the second param in Forge's `createTable()` method

sammyskills:
According to the [docs](https://codeigniter4.github.io/CodeIgniter4/installation/upgrade_422.html#others), 
* The method `Forge::createTable()` no longer executes a `CREATE TABLE IF NOT EXISTS`. If table is not found in `$db->tableExists($table)` then `CREATE TABLE` is executed.
* The second parameter `$ifNotExists` of `Forge::_createTable()` is deprecated. It is no longer used and will be removed in a future release.

And since shield has been updated to `^4.2.3`, I think this should be followed.

### DIFF
--- a/src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
+++ b/src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
@@ -22,7 +22,7 @@ class CreateAuthTables extends Migration
         ]);
         $this->forge->addPrimaryKey('id');
         $this->forge->addUniqueKey('username');
-        $this->forge->createTable('users', true);
+        $this->forge->createTable('users');
 
         /*
          * Auth Identities Table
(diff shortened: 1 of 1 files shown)


MGatner (approved):
I'm in favor of this both due to the changes and for general practice. I don't like `IF [NOT] EXISTS` in migrations because I believe that if a migration conflict occurs then some other problem has happened and "hiding" it is a mistake. 

But not everyone shares my opinion so let's see what others say.

kenjis:
> The method Forge::createTable() no longer executes a CREATE TABLE IF NOT EXISTS. If table is not found in $db->tableExists($table) then CREATE TABLE is executed.

Ah, this may be misleading.

> The method Forge::createTable() no longer executes a CREATE TABLE IF NOT EXISTS.

True.

> If table is not found in $db->tableExists($table) then CREATE TABLE is executed.

This is when `$ifNotExists` is true.
When `$ifNotExists` is false (by default), CREATE TABLE is always executed.

kenjis (approved):

datamweb (approved):

MGatner:
Thanks @sammyskills!

PR #350, opened 2022-08-08
fix: field name translation in validation errors

datamweb:
Before:
befor

After:
After

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -95,35 +95,47 @@ Shield has the following rules for registration:
 
 ```php
 [
-    'username'         => [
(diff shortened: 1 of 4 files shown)


datamweb:
We have a bug in `MagicLinkController`, I prefer to fix it in a separate PR and write the test there.
https://github.com/codeigniter4/shield/blob/342a368536678621998c3c41d276480cd14ec6c6/src/Controllers/MagicLinkController.php#L63

kenjis (approved):
LGTM!

MGatner (approved):

Issue #264, opened 2022-06-30
Security : Translation


ThibautPV:
Hello,

I think that the translations should be modified to improve the security of Shield.

Indeed, the current translations can give information to malicious people and also be compromising for personal data. 

For example, with this translation: "Unable to log you in. Please check your password", we know that the email address exists in the database.

This is also the case with the sentence "Unable to verify the email address matches the email on record."

For example, I suggest replacing with the valid sentence "Check your email! We just sent you an email with a Login link inside. It is only valid for 60 minutes."

What do you think?

InstantT

kenjis:
This is a usability and security issue.

It may be better to change the default message to something less informative, 
since more secure is preferred as the default.

lonnieezell:
I’m on my phone currently and can’t find the right doc, but the latest advice from NOST, IIRC, is that you should focus on usability and help people with the messages instead of secure it. Combined with encouraging better passwords they found the obscure messaging wasn’t needed and tended to hurt the experience with no benefit to security.

kenjis:
Authentication and Error Messages
https://cheatsheetseries.owasp.org/cheatsheets/Authentication_Cheat_Sheet.html#authentication-and-error-messages
> Incorrectly implemented error messages in the case of authentication functionality can be used for the purposes of user ID and password enumeration. An application should respond (both HTTP and HTML) in a generic manner.

> The problem with returning a generic error message for the user is a User Experience (UX) matter. A legitimate user might feel confused with the generic messages, thus making it hard for them to use the application, and might after several retries, leave the application because of its complexity. The decision to return a generic error message can be determined based on the criticality of the application and its data. For example, for critical applications, the team can decide that under the failure scenario, a user will always be redirected to the support page and a generic error message will be returned.

kenjis:
Probably Lonnie said about NIST Guidelines https://pages.nist.gov/800-63-3/
But I didn't immediately know where exactly to read.

lonnieezell:
I was just reading that part from OWASP about usability. I’ll try to find the NIST recommends later tonight. However, I’m tempted to say we leave it as is, and let people that need more security change their lang file. This is similar to how we treat remember me - since more secure sites are recommended not to have that functionality anyway.

MGatner:
Maybe we need a docs section on "hardening security" with recommendations on how to set this package to be maximally secure.

ThibautPV:
I think you should still modify the translations or indicate it in the documentation as suggested by @MGatner 

[...]

PR #351, opened 2022-08-09
fix: show validate errors in `magic-link`

datamweb:
Fixes: https://github.com/codeigniter4/shield/pull/350#issuecomment-1207656514

### DIFF
--- a/src/Controllers/MagicLinkController.php
+++ b/src/Controllers/MagicLinkController.php
@@ -60,7 +60,7 @@ class MagicLinkController extends BaseController
         // Validate email format
         $rules = $this->getValidationRules();
         if (! $this->validate($rules)) {
-            return redirect()->route('magic-link')->with('error', lang('Auth.invalidEmail'));
+            return redirect()->route('magic-link')->with('errors', $this->validator->getErrors());
         }
 
         // Check if the user exists
--- a/src/Views/magic_link_form.php
+++ b/src/Views/magic_link_form.php
@@ -9,9 +9,20 @@
         <div class="card-body">
             <h5 class="card-title mb-5"><?= lang('Auth.useMagicLink') ?></h5>
 
-            <?php if (session('error') !== null) : ?>
-            <div class="alert alert-danger" role="alert"><?= session('error') ?></div>
-            <?php endif ?>
+                <?php if (session('error') !== null) : ?>
+                    <div class="alert alert-danger" role="alert"><?= session('error') ?></div>
+                <?php elseif (session('errors') !== null) : ?>
+                    <div class="alert alert-danger" role="alert">
+                        <?php if (is_array(session('errors'))) : ?>
+                            <?php foreach (session('errors') as $error) : ?>
+                                <?= $error ?>
+                                <br>
+                            <?php endforeach ?>
+                        <?php else : ?>
+                            <?= session('errors') ?>
+                        <?php endif ?>
+                    </div>
+                <?php endif ?>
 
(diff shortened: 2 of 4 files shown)


datamweb:
Hi @kenjis, I had to edit file `tests/Authentication/MagicLinkTest.php`.
We have two test files for `MagicLink` :
 `tests/Authentication/MagicLinkTest.php` and `tests/Controllers/MagicLinkTest.php`.

**Please check more carefully.**

kenjis:
Why do you think this is bug fix?
Why do you need to show the validation error?

datamweb:
Currently, the shield is displaying message `Unable to verify the email address matches the email on record`.  if the email is empty!
Or if it is email format wrong, the email form will display message `Unable to verify the email address matches the email on record`!
Maybe there is no problem in the performance of the Shield, **but is actually not displaying the correct message**.
Block `if ` is related to validation, so it is reasonable to display the messages related to validation and not `Auth.invalidEmail`.

Can you explain to me why message `Auth.invalidEmail` should be displayed instead of message `The Email Address field is required.` when the form is empty?

MGatner (approved):
Very nice! Thank you @datamweb

MGatner:
@datamweb I tried emailing you on 7/31 (@yahoo.com) but heard no response. Is there a better way to get in touch with you? Are you on the CodeIgniter Slack?

datamweb:
> datamweb I tried emailing you on 7/31 (@yahoo.com) but heard no response. Is there a better way to get in touch with you? Are you on the CodeIgniter Slack?

@MGatner Yes , I received your email.
I replied to your email on **1 August**.
It seems that you had a problem receiving my email.
I sent you the email again today.

Also, I have registered in slack after receiving your email.
**datamweb.slack.com** and my email (@yahoo.com).

In any case, if you do not succeed in contacting me, please let me know.

kenjis:
I'm not sure what error message is better.

> why the message Auth.invalidEmail should be displayed when the length of the entered email is more than 254 characters?

It is kind of an attack. We don't need to response kindly.
There is also Security vs Usability issue.

datamweb:
> It is kind of an attack. We don't need to response kindly. There is also Security vs Usability issue.

I believe the code should be secure enough that I don't need to hide it.
On the other hand, I believe that the failure to display the messages correctly will destroy user-friendliness.
However, your experience in this matter is higher, and if you think that these PR will harm the security of the shield, I respect it.
Please let me know so I can close.

kenjis:
The security concern about the magic-link page is that attackers know that the email address exists in the database.
Ref #264

In the current implementation, attackers easily know it from the error message.
Screenshot 2022-08-13 17 10 07
Screenshot 2022-08-13 17 10 23

On the other hand, this PR just gives validation error messages.
An invalid email address is never be the address in the database.
So attackers do not send such a email address.

The conclusion is this PR does not increase the security risk.

kenjis (approved):

datamweb:
> In the current implementation, attackers easily know it from the error message.

@kenjis, encouraging users to use captcha on magic-link, can it be useful?
If yes, we can create a note in the documents.

kenjis:
@datamweb To increase the cost of posting the form is a mitigation.
Overall risk can be lowered. Captcha or rate limit can be used.

However, it may not be very useful if an attacker wants to find out if one particular person's e-mail address is registered.

datamweb:
So we have this issue in case  `$registrationEmailRules` as well.

PR #356, opened 2022-08-10
docs: remove unnecessary details of `CI4.1.x`

datamweb:
Hello friends,
CodeIgniter Shield requires Codeigniter v4.2.3 or later.
Do we need a description of the lower version?


kenjis (approved):
Thank you!

MGatner (approved):
Good catch! That security issue forced us to bump minimum version, but I'm actually glad we can implement the newer features.

PR #366, opened 2022-08-12
fix: bootstrap column on mobile device

MitkoIT:

PR #367, opened 2022-08-12
Update bootstrap col on mobile device

MitkoIT:


datamweb:
Hi @MitkoIT ,
I prefer this commit to be done in PR #366 .
So, please do it in #366.
Thanks

PR #368, opened 2022-08-12
feat: autoload helpers by Composer

kenjis:


docs/authentication.md
```diff
@@ -65,6 +63,10 @@ auth()->id();
 user_id();
 ```
 
+> **Note**
+> The `auth_helper` is autoloaded by Composer. If you want to *override* the functions,
+> you need to define them in `app/Common.php`.
```
MGatner, line 68: I was wrong, Common.php does come first! That is good.

MGatner (approved):

docs/authentication.md
datamweb: > you need to define them in `app/Common.php`

This was unclear to me too!

datamweb:
@kenjis، in this PR, something else is going to be done?
If not please merge.

kenjis:
Nothing. But only MGater approved.

datamweb (changes requested):

datamweb:
I didn't like it!! Why repeat twice!

Screenshot 2022-08-14 114856

kenjis:
Because it checks the file twice.
`$this->helpers = array_merge($this->helpers, ['auth', 'setting']);` and
`$this->helpers = array_merge($this->helpers, ['setting']);`

It is not easy to show the message once.

docs/install.md
```diff
@@ -117,12 +117,12 @@ class Auth extends ShieldAuth
 }
 ```
 
-2. **Helper Setup** The `auth` and `setting` helpers need to be included in almost every page. The simplest way to do this is to add them to the `BaseController::initController` method:
+2. **Helper Setup** The `setting` helpers need to be included in almost every page. The simplest way to do this is to add them to the `BaseController::initController` method:
```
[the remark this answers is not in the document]
kenjis: Thank you. I've fixed!

datamweb (approved):

kenjis:
@datamweb I could fix the repeated Skip messages. See https://github.com/codeigniter4/shield/pull/371

PR #371, opened 2022-08-14
Improve setup

kenjis:
Follow-up #368

- show Skip message once
- small fix for the docs

---

Clean installation and
When `$this->helpers = array_merge($this->helpers, ['auth', 'setting']);` exists: 
screenshot 2022-08-14 19 32 06

---

When `$this->helpers = array_merge($this->helpers, ['setting']);` exists:
screenshot 2022-08-14 19 33 36

### DIFF
--- a/src/Commands/Setup.php
+++ b/src/Commands/Setup.php
@@ -234,8 +234,7 @@ class Setup extends BaseCommand
         $replaces = [
             '$this->helpers = array_merge($this->helpers, [\'auth\', \'setting\']);' => $check,
         ];
-        $return = $this->replace($file, $replaces);
-        if ($return) {
+        if ($this->replace($file, $replaces)) {
             return;
         }
 
--- a/src/Commands/Setup.php
+++ b/src/Commands/Setup.php
@@ -196,10 +196,12 @@ class Setup extends BaseCommand
     }
 
     /**
+     * Replace for setupHelper()
+     *
      * @param string $file     Relative file path like 'Controllers/BaseController.php'.
      * @param array  $replaces [search => replace]
      */
-    protected function replace(string $file, array $replaces): void
+    private function replace(string $file, array $replaces): bool
     {
         $path      = $this->distPath . $file;
         $cleanPath = clean_path($path);
@@ -209,16 +211,18 @@ class Setup extends BaseCommand
         $output = $this->replacer->replace($content, $replaces);
 
         if ($output === $content) {
-            CLI::error("  Skipped {$cleanPath}. It has already been updated.");
-
-            return;
+            return false;
         }
 
         if (write_file($path, $output)) {
             CLI::write(CLI::color('  Updated: ', 'green') . $cleanPath);
-        } else {
-            CLI::error("  Error updating {$cleanPath}.");
+
+            return true;
         }
+
+        CLI::error("  Error updating {$cleanPath}.");
+
+        return false;
     }
 
     private function setupHelper(): void
@@ -230,7 +234,10 @@ class Setup extends BaseCommand
         $replaces = [
             '$this->helpers = array_merge($this->helpers, [\'auth\', \'setting\']);' => $check,
         ];
-        $this->replace($file, $replaces);
+        $return = $this->replace($file, $replaces);
+        if ($return) {
+            return;
+        }
 
         // Add helper setup
         $pattern = '/(' . preg_quote('// Do Not Edit This Line', '/') . ')/u';
--- a/docs/install.md
+++ b/docs/install.md
@@ -117,7 +117,7 @@ class Auth extends ShieldAuth
 }
 ```
 
-2. **Helper Setup** The `setting` helper needs to be included in almost every page. The simplest way to do this is to add them to the `BaseController::initController` method:
+2. **Helper Setup** The `setting` helper needs to be included in almost every page. The simplest way to do this is to add it to the `BaseController::initController` method:
 
 ```php
 public function initController(RequestInterface $request, ResponseInterface $response, LoggerInterface $logger)


datamweb (approved):
Great, I confirm that the problem is fixed.
Screenshot 2022-08-14 142303
Thanks!

src/Commands/Setup.php
```diff
@@ -230,7 +234,10 @@ private function setupHelper(): void
         $replaces = [
             '$this->helpers = array_merge($this->helpers, [\'auth\', \'setting\']);' => $check,
         ];
-        $this->replace($file, $replaces);
+        $return = $this->replace($file, $replaces);
+        if ($return) {
```
MGatner, line 238: Since it is always a boolean can we skip the intermediary?
```suggestion
        if ($this->replace($file, $replaces)) {
```

MGatner (changes requested):

datamweb:
Wait please!

kenjis:
@datamweb What do you mean? Any problem?

datamweb:
> @datamweb What do you mean? Any problem?

@kenjis, `APPPATH\Config/Auth.php` Or `APPPATH\Config\Auth.php`?

kenjis:
No problem. The code works. It is just only the output is a bit strange.
It is because Windows uses `\` as a path separator.

On Unix system:
```
CodeIgniter v4.2.4 Command Line Tool - Server Time: 2022-08-15 02:57:04 UTC-05:00

  Created: APPPATH/Config/Auth.php
  Created: APPPATH/Config/AuthGroups.php
  Updated: APPPATH/Controllers/BaseController.php
  Updated: APPPATH/Config/Routes.php
  Run `spark migrate --all` now? [y, n]: n
```

datamweb:
Okey @Kenjis, my question is why when there are `app\Config\Auth.php` and `app\Config\AuthGroups.php` files.
Is the correct output shown?
```
  File 'APPPATH\Config\Auth.php' already exists in destination. Overwrite? [n, y]: y
  Created: APPPATH\Config\Auth.php
  File 'APPPATH\Config\AuthGroups.php' already exists in destination. Overwrite? [n, y]: y
  Created: APPPATH\Config\AuthGroups.php
  Skipped APPPATH\Controllers\BaseController.php. It has already been updated.
  Skipped APPPATH\Config\Routes.php. It has already been updated.
  Run `spark migrate --all` now? [y, n]: n
```

datamweb:
@kenjis , I found a way, but I don't know if it's the right one.
If you change `/` to` \` below, the correct output will be shown.
https://github.com/codeigniter4/shield/blob/c48cb81bd80762e0133d1ba53f50e16ab99c1f6e/src/Commands/Setup.php#L108

https://github.com/codeigniter4/shield/blob/c48cb81bd80762e0133d1ba53f50e16ab99c1f6e/src/Commands/Setup.php#L120

output:
```
P:\WorkApp\CI423\cI423>php spark shield:setup

CodeIgniter v4.2.4 Command Line Tool - Server Time: 2022-08-15 04:03:36 UTC-05:00

  Created: APPPATH\Config\Auth.php
  Created: APPPATH\Config\AuthGroups.php
  Skipped APPPATH\Controllers\BaseController.php. It has already been updated.
  Skipped APPPATH\Config\Routes.php. It has already been updated.
  Run `spark migrate --all` now? [y, n]:
```

kenjis:
>  I found a way, but I don't know if it's the right one.

It works only on Windows.

kenjis:
> why when there are app\Config\Auth.php and app\Config\AuthGroups.php files.
Is the correct output shown?

I don't know. It is a mystery!
Search "Created:" in the `Setup` class. There is only one place.

MGatner (approved):
Looks good to me! I don't really care about the path separator issue as I think the message conveys its intent both ways, but I'm happy for you two to keep discussing 🤗

kenjis:
@datamweb The solution that works perfectly is to use `DIRECTORY_SEPARATOR` for the path in PHP code.
```php
$file     = 'Config' . DIRECTORY_SEPARATOR . 'Auth.php'
```
But I don't like it, because the code will be a bit difficult to read and longer.

datamweb:
@kenjis, Even if we define a private function, for example
```php
$file     = 'Config/Auth.php';
$file = $this->modifyDisplayPath($file);
```
Do you believe the code is unreadable?
If yes, I have no problem with the merge.

kenjis:
@datamweb Yes, it is a way to handle this. Where should we do the convertion?

datamweb:
@kenjis, that's what I meant.

```php
    private function publishConfigAuth(): void
    {
        
         $file = 'Config/Auth.php';
         $file = $this->modifyDisplayPath($file);
        //$file     = 'Config' . DIRECTORY_SEPARATOR . 'Auth.php';
        $replaces = [
            'namespace CodeIgniter\Shield\Config'  => 'namespace Config',
            'use CodeIgniter\\Config\\BaseConfig;' => 'use CodeIgniter\\Shield\\Config\\Auth as ShieldAuth;',
            'extends BaseConfig'                   => 'extends ShieldAuth',
        ];

        $this->copyAndReplace($file, $replaces);
    }
    
    private function modifyDisplayPath(string $file): string
    {
        return str_replace('/',DIRECTORY_SEPARATOR,$file);
    }
```

kenjis:
There is another way, to convert just right before output.

datamweb:
```diff
    protected function writeFile(string $file, string $content): void
    {
        $path      = $this->distPath . $file;
--- $cleanPath = clean_path($path);
+++  $cleanPath = $this->modifyDisplayPath(clean_path($path));

        $directory = dirname($path);
```
```diff
+++private function modifyDisplayPath(string $file): string
+++{
+++return str_replace('/',DIRECTORY_SEPARATOR,$file);
+++}
```

MGatner:
Would it be too heavy-handed to modify `clean_path()` to do this natively? Convert between OS-specific slashes.

kenjis:
I think that is possible.

PR #377, opened 2022-08-14
refactor: add return types

kenjis:
- add `void_return` rule for php-cs-fixer
- add missing return types
- add doc comments

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -17,6 +17,7 @@ $finder = Finder::create()
 
 $overrides = [
     'declare_strict_types' => true,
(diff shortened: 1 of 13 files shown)


paulbalandan (commented):
To automate this for void returns, you can use the `void_return` rule of php-cs-fixer. However, that rule can break public API.

kenjis:
@paulbalandan Shield is under pre-release. We don't care small breaking changes so much.

MGatner (approved):
I usually run this manually with Psalm's CLI arguments, but maybe we should think about making this change in DevKit.

datamweb (approved):

PR #378, opened 2022-08-15
Remove superfluous `--using-cache` option

paulbalandan:
When using `nexusphp/cs-config`, the `--using-cache` is default `true` if not provided.
https://github.com/NexusPHP/cs-config/blob/e9092a9ab3a59c982bf0c79c89df36a7b2c0b86e/src/Factory.php#L122

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -75,7 +75,7 @@
             "phpstan analyze",
             "psalm"
         ],
-        "cs-fix": "php-cs-fixer fix --ansi --verbose --diff --using-cache=yes",
+        "cs-fix": "php-cs-fixer fix --ansi --verbose --diff",
         "style": "@cs-fix",
         "test": "phpunit"
     }


kenjis:
I don't know why, but the original setting was `--using-cache=no`.
And the option does not show in the help `vendor/bin/php-cs-fixer --help`.
So I leave the option to easily know the option name.

I'm fine with this PR.

datamweb:
Related : #340

MGatner:
It came from my default toolkit:
https://github.com/tattersoftware/codeigniter4-tools/blob/develop/src/composer.php#L121

These are primarily for my library development, since maintaining the development tools across 40 repos is a nightmare. These libraries are mostly pretty small so I prefer to run without cache as my default. I'm fine with speeding things up for Shield, which is a larger scope. Also good with removing the argument here after @paulbalandan's explanation.

paulbalandan:
Are there blocking issues/concerns preventing the merge?

datamweb (approved):

MGatner (approved):
No problems! There wasn't any activity after my comment so it didn't come back in my feed. Sorry!

PR #379, opened 2022-08-15
refactor: add `declare(strict_types=1)`

kenjis:
- add `declare_strict_types` rule to php-cs-fixer
- add `declare(strict_types=1)`

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 use CodeIgniter\CodingStandard\CodeIgniter4;
 use Nexus\CsConfig\Factory;
 use PhpCsFixer\Finder;
@@ -13,7 +15,9 @@ $finder = Finder::create()
     ->exclude('build')
     ->append([__FILE__]);
 
-$overrides = [];
+$overrides = [
+    'declare_strict_types' => true,
+];
 
 $options = [
     'finder'    => $finder,
--- a/src/Auth.php
+++ b/src/Auth.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace CodeIgniter\Shield;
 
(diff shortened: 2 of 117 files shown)


MGatner (approved):
Nice to see everything was already working as strict! I think this is a good change.  @lonnieezell has been hesitant in the past so let's make sure he has a chance to weigh in.

paulbalandan (approved):

paulbalandan:
When will this ship to the main repo?? 🤔😂

lonnieezell:
If everyone else on the team feels strongly then we should do this then do it. My hesitation has always been based on an incorrect understanding I think. IIRC, this will still cast numeric strings, for example, into ints when called from my code, just not when called within the same file. If I'm remembering that correct then I'm fine. If I'm not, and it forces you to explicitly cast in your own code I find that onerous. I stopped programming in C/C++ years ago and don't have an interest in using Java :)

kenjis:
See https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.strict

> Strict typing applies to function calls made from within the file with strict typing enabled, not to the functions declared within that file. If a file without strict typing enabled makes a call to a function that was defined in a file with strict typing, the caller's preference (coercive typing) will be respected, and the value will be coerced.

datamweb:
There is one thing that you need to keep in mind when using the strict_type declaration. It only works for the file that the declaration is placed in. Meaning that when you use a class, which has the strict_type declaration, it can still result in PHP casting primitive value. So keep that in mind when you want to start using this!

lonnieezell:
Perfect. Then I'm fine.

datamweb (approved):

MGatner:
Yes, these are correct. The good news is that declaring strict has grown massively during the static analysis era, since most SA tools actually end up enforcing proper "strict" handling already. I expect a larger percent of developers than ever to understand the benefits and limitations. 

Merge complete. On to the main repo! ====€

PR #383, opened 2022-08-15
fix: redirect()->withInput() causes ValidationException

kenjis:
Fixes https://github.com/codeigniter4/shield/discussions/381

`redirect()->withInput()` or `withErrors()` saves validation errors in the Session.
And the check of the validation errors in Model (`checkQueryReturn()`),
`$this->validation->getErrors()` returns the errors in the Session. 

**How to Test**
```diff
--- a/app/Config/Filters.php
+++ b/app/Config/Filters.php
@@ -36,6 +36,7 @@ class Filters extends BaseConfig
             // 'honeypot',
             // 'csrf',
             // 'invalidchars',
+            'session' => ['except' => 'login*'],
         ],
         'after' => [
             'toolbar',
```
```diff
--- a/app/Controllers/Home.php
+++ b/app/Controllers/Home.php
@@ -2,10 +2,22 @@
 
 namespace App\Controllers;
 
+use Config\Services;
+
 class Home extends BaseController
 {
     public function index()
     {
-        return view('welcome_message');
+        $this->validation = Services::validation();
+
+        $this->validation->setRules([
+            'email' => 'required|valid_email|unique_email[{user_id}]',
+        ]);
+
+        if (! $this->validation->withRequest($this->request)->run()) {
+            return redirect()
+                ->to('logout')
+                ->withInput();
+        }
     }
 }
```

### DIFF
--- a/src/Models/CheckQueryReturnTrait.php
+++ b/src/Models/CheckQueryReturnTrait.php
@@ -52,8 +52,8 @@ trait CheckQueryReturnTrait
      */
     private function getValidationErrors(): array
     {
-        // @TODO When CI v4.3 is released, you can use `$this->validation->getErrors(false)`.
-        // See https://github.com/codeigniter4/CodeIgniter4/pull/6381
+        // @TODO When CI v4.3 is released, you don't need this hack.
+        //       See https://github.com/codeigniter4/CodeIgniter4/pull/6384
         return $this->getValidationPropertyErrors();
     }
 
--- a/src/Models/CheckQueryReturnTrait.php
+++ b/src/Models/CheckQueryReturnTrait.php
@@ -32,7 +32,7 @@ trait CheckQueryReturnTrait
 
     private function checkValidationError(): void
     {
-        $validationErrors = $this->validation->getErrors();
+        $validationErrors = $this->getValidationErrors();
 
         if ($validationErrors !== []) {
             $message = 'Validation error:';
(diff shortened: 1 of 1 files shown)


datamweb:
@kenjis, I think we should merge this PR.
The number of people involved in PR seems to be high.
according to what you did recently for **hotfix CI**, we can proceed to write the test after the merge.

jozefrebjak:
+1 for merge now, It's breaking app at all.

kenjis:
To merge, we need reviews.

jozefrebjak:
@kenjis For me this PR solved validation in forms.

MGatner:
I don't like the production use of reflection. It's hard to trace this back in mobile, but couldn't this be solved by handling our validations from a clean instance?
```php
// Validate the registration
if (! single_service('validation')->setRules($this->registrationRules)->run($event)) {
...
```

kenjis:
> couldn't this be solved by handling our validations from a clean instance?

No, because the session data ($_SESSION) is global.

See https://github.com/codeigniter4/CodeIgniter4/pull/6381
It helps your understanding.

MGatner:
That helps! Okay I am on board with the Reflection solution. It's a hack but we have a fix in `framework` at least. Maybe add a note that if the minimum CI4 version is ever increased past 4.2.4 then it can be refactored. Or alternatively, check the actual version and have Reflection be the fallback solution.

MGatner (changes requested):
_See comment above_

datamweb (approved):
After applying comment **MGatner** , I don't see any problem.

Screenshot 2022-08-16 145031

MGatner (approved):
👍

PR #384, opened 2022-08-16
feat: update `shield:setup` for Security Setup

datamweb:
Hello friends,
This PR gives command `shield:setup` the ability to automatically implement the security proposal issued in [SecurityAdvisories](https://github.com/codeigniter4/shield/security/advisories/GHSA-5hm8-vh6r-2cjq).

### DIFF
--- a/src/Commands/Setup.php
+++ b/src/Commands/Setup.php
@@ -87,9 +87,9 @@ class Setup extends BaseCommand
         $this->setupHelper();
         $this->setupRoutes();
 
-        $this->runMigrations();
-
         $this->setSecurityCSRF();
+
+        $this->runMigrations();
     }
 
     /**
@@ -260,18 +260,6 @@ class Setup extends BaseCommand
         $this->add($file, $check, $pattern, $replace);
     }
 
-    private function runMigrations(): void
-    {
-        if (
-            $this->cliPrompt('  Run `spark migrate --all` now?', ['y', 'n']) === 'n'
-        ) {
-            return;
-        }
-
-        $command = new Migrate(Services::logger(), Services::commands());
-        $command->run(['all' => null]);
-    }
-
     /**
      * @see https://github.com/codeigniter4/shield/security/advisories/GHSA-5hm8-vh6r-2cjq
      */
@@ -302,10 +290,22 @@ class Setup extends BaseCommand
         }
 
         if (write_file($path, $output)) {
-            CLI::write(CLI::color('  Update: ', 'green') . "We have updated file '{$cleanPath}' for security reasons.");
+            CLI::write(CLI::color('  Updated: ', 'green') . "We have updated file '{$cleanPath}' for security reasons.");
         } else {
             CLI::error("  Error updating file '{$cleanPath}'.");
         }
+    }
+
+    private function runMigrations(): void
+    {
+        if (
+            $this->cliPrompt('  Run `spark migrate --all` now?', ['y', 'n']) === 'n'
+        ) {
+            return;
+        }
+
+        $command = new Migrate(Services::logger(), Services::commands());
+        $command->run(['all' => null]);
     }
 
     /**
--- a/src/Commands/Setup.php
+++ b/src/Commands/Setup.php
@@ -88,6 +88,8 @@ class Setup extends BaseCommand
         $this->setupRoutes();
 
         $this->runMigrations();
+
+        $this->setSecurityItem();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Commands/Setup.php
```diff
@@ -98,6 +100,10 @@ protected function copyAndReplace(string $file, array $replaces): void
     {
         $path = "{$this->sourcePath}/{$file}";
 
+        if ($file === 'Config/Security.php') {
+            $path = "{$this->distPath}/{$file}";
+        }
```
kenjis, line 105: This `if` is not good. 

The `copyAndReplace()` method does:
1. read a source file
2. replace it
3. write it to dest path

But this change will break the behavior.
It is unpredictable for devs if they read the code carefully.

Conclusion: you should not use this method for `Config/Security.php`.

src/Commands/Setup.php
```diff
@@ -146,7 +152,7 @@ protected function writeFile(string $file, string $content): void
             mkdir($directory, 0777, true);
         }
 
-        if (file_exists($path)) {
+        if (file_exists($path) && $file !== 'Config/Security.php') {
```
MGatner, line 155: I know this code predates this PR, but I'm just noticing it... `file_exists()` should really only be used to check for collisions. `is_file()` has better performance and won't return `true` for directories (which can cause a lot of confusion, given how the first method is named).

MGatner (commented):
This is a cool change! @kenjis has a better handle on the replacement system so I will leave those details, but I'm grateful for the work.

src/Commands/Setup.php
datamweb: @kenjis for the step by step explanation, thank you, I like this type of explanation.

I changed the code as follows:

```php
    private function setSecurityItem(): void
    {
        $file     = 'Config/Security.php';
        $replaces = [
            'public $csrfProtection = \'cookie\';' => 'public $csrfProtection = \'session\';',
        ];

        $path      = $this->distPath . $file;
        $cleanPath = clean_path($path);

        if (!is_file($path)){
            CLI::write(CLI::color('  Not found: ', 'red') . $cleanPath);

            return;
        }
        
        $content = file_get_contents($path);
        $output = $this->replacer->replace($content, $replaces);

        // check $csrfProtection = 'session' 
        if ($output === $content) {
            CLI::write(CLI::color('  Security Setup: ', 'green') . 'Everything is fine.');

            return;
        }

        if (write_file($path, $output)) {
            CLI::write(CLI::color('  Update: ', 'green') . "We have updated file '{$cleanPath}' for security reasons.");
        } else {
            CLI::error("  Error updating file {$cleanPath}.");
        }
    }
```

Is it understandable and readable enough?

kenjis: Okay, it is the good first step. Now we have only one Config file that will be updated directly, it is fine.

Can you change the method name to `setSecurityCSRF`?

Isn't it error?
```php
CLI::write(CLI::color('  Not found: ', 'red') . $cleanPath);
```
If there is no `Config/Security.php`, it cannot set `session` for CSRF protection,
and Shield will not work.

datamweb: > Can you change the method name to `setSecurityCSRF`?

**Done.**

> Isn't it error?

Yes! In fact, checking the existence of file `Config/Security.php` is a bit exaggerated. However, I considered the worst possible situation.
**Done.**

src/Commands/Setup.php
datamweb: I agree with this.
But I think it is more logical for us to solve this issue in a separate PR under title **refactor**.
I think file `Setup.php` needs **refactor**. This refactor can include fixing problem `APPPATH\Config/Auth.php`, fixing the not very interesting naming of methods(`writeFile `and `write_file `| `replace `and `$this->replacer->replace`) and fixing the use of `file_exists()`.
However, if you think that this should be fixed in public relations. please inform.

MGatner: I like your approach; hold for later.

MGatner (approved):
Looks good! Please await @kenjis' final review.

src/Commands/Setup.php
```diff
@@ -88,6 +88,8 @@ private function publishConfig(): void
         $this->setupRoutes();
 
         $this->runMigrations();
+
+        $this->setSecurityCSRF();
```
kenjis, line 92: Can you move this line above the `$this->runMigrations()`?
It seems to me that a good order would be to set up the configuration files and then run the migration last.

datamweb: **Done**, and I made some changes to the test.
@kenjis please check again.

kenjis (approved):
LGTM! Thank you.

Issue #385, opened 2022-08-16
Bug: NothingPersonalValidator strip_explode too sensible


tomatlscomm:
### PHP Version

7.4

### CodeIgniter4 Version

4.2.4

### Shield Version

1.0.0-beta.2

### Which operating systems have you tested for this bug?

macOS

### Which server did you use?

apache

### Database

MySQL 5.6

### Did you customize Shield?

No

### What happened?

The function NothingPersonalValidator->strip_explode() is too sensible for exploding the user password.
Indeed, it explode any part separated for example with an underscore or a "-".
This can produce the search of only one letter in the username and/or email adresse as part of a personal information.
The user is then prompted wrongly that his password cannot contain any personal information contained is his username or email.

### Steps to Reproduce

Try to register with : 
email : xxx@gmail.com
password : G-test#1234

In this case, NothingPersonalValidator will fail cause "G" is considered as part of "xxx@gmail.com"

### Expected Output

I think the NothingPersonalValidator->strip_explode() function should check if the searched string is not too short and then try to make a match only if the searching string contain 3 or more caracters.

datamweb:
Hello @tomatlscomm ,
Thank you very much for your report.
I confirm this issue.

![Screenshot 2022-08-16 180439](https://user-images.githubusercontent.com/9530214/184906817-bfd0b4cb-cc60-48df-87e0-209a959220cb.jpg)

tomatlscomm:
What i suggest to fix : 

File /src/Authentication/Passwords/NothingPersonalValidator.php, line 115
Replace 
`if (empty($haystack) || in_array($haystack, $trivial, true)) {`
by
`if (empty($haystack) || in_array($haystack, $trivial, true) || strlen($haystack)<3) {`

MGatner:
I think that is a fine solution @tomatlscomm. We will want to make sure it works with existing tests and has its own new test. Would you be willing to open a Pull Request?

tomatlscomm:
@MGatner Sure

PR #389, opened 2022-08-18
fix: NothingPersonalValidator strip_explode too sensible

tomatlscomm:
Fixes https://github.com/codeigniter4/shield/issues/385

What happened?
The function NothingPersonalValidator->strip_explode() is too sensible for exploding the user password.
Indeed, it explode any part separated for example with an underscore or a "-".
This can produce the search of only one letter in the username and/or email adresse as part of a personal information.
The user is then prompted wrongly that his password cannot contain any personal information contained is his username or email.

Steps to Reproduce
Try to register with :
email : xxx@gmail.com
password : G-test#1234

In this case, NothingPersonalValidator will fail cause "G" is considered as part of "xxx@gmail.com"

Expected Output
I think the NothingPersonalValidator->strip_explode() function should check if the searched string is not too short and then try to make a match only if the searching string contain 3 or more caracters.

### DIFF
--- a/src/Authentication/Passwords/NothingPersonalValidator.php
+++ b/src/Authentication/Passwords/NothingPersonalValidator.php
@@ -114,8 +114,8 @@ class NothingPersonalValidator extends BaseValidator implements ValidatorInterfa
             $haystacks = $this->strip_explode($password);
 
             foreach ($haystacks as $haystack) {
-                if (empty($haystack) || in_array($haystack, $trivial, true)) {
-                    continue;  //ignore trivial words
+                if (empty($haystack) || in_array($haystack, $trivial, true) || strlen($haystack)<3) {
+                    continue;  //ignore trivial words, or little words of less than 3 characters
                 }
 
                 foreach ($needles as $needle) {


datamweb (commented):
Hi, 
Thank you for sending PR.
This PR needs a unit test, can you write a unit test for it?

kenjis:
Thank you for the PR.

Please sign your git commits.

MGatner:
It looks like most of our Actions did not run. Is anyone else only seeing PHPCPD?

datamweb:
> It looks like most of our Actions did not run. Is anyone else only seeing PHPCPD?

Yes, they did not run.

kenjis:
Ran GA jobs manually.

kenjis:
Please fix coding style.
See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#php-style

> Note, we expect all code changes or bug-fixes to be accompanied by one or more tests added to our test suite to prove the code works. If pull requests are not accompanied by relevant tests, they will likely be closed. 
https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#contributions

If you are not familiar with PHPUnit testing, See [Resources](https://github.com/codeigniter4/CodeIgniter4/tree/develop/tests#resources)

MGatner:
Thank you @tomatlscomm. Your suggestion was incorporated into another PR, and the fix is now merged.

PR #391, opened 2022-08-19
fix: Session auth action checks

### LINKED ISSUE #349: Bug: LogicException occurs when an unactivated user tries to log in
### PHP Version

8.0.21

### CodeIgniter4 Version

4.2.3

### Shield Version

1.0.0-beta.2

macOS

### Which server did you use?

cli-server (PHP built-in webserver)

### Database

MySQL 5.7

### Did you customize Shield?

No.

### What happened?

CodeIgniter\Shield\Exceptions\LogicException

The user has User Info in Session, so already logged in or in pending login state. If a logged in user logs in again with other account, the session data of the previous user will be used as the new user. Fix your code to prevent users from logging in without logging out or delete the session data. user_id: 2 

### Steps to Reproduce

`Config\Auth.php`:
```php
    public array $actions = [
        'login'    => null,
        'register' => 'CodeIgniter\Shield\Authentication\Actions\EmailActivator',
    ];
```

1. Register a user.
2. Go to the login page, and try to login.

### Expected Output

At least, no Exception occurs.

### Anything else?

Related #343

kenjis:
- Fixes #349
  - if the user goes to the login page, redirected to the Email Activation page and the code will be sent again
- After registration, when the user do not send activation code and the session expires
  - if the user logs in, the Email Activation code will be sent again

### DIFF
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -413,11 +413,21 @@ class Session implements AuthenticatorInterface
      */
     private function setAuthAction(): void
     {
-        // Get identities for action
-        $identities = $this->getIdentitiesForAction();
+        // First, check ID_TYPE_EMAIL_ACTIVATE identity
+        $this->setAuthActionEmailActivate();
 
-        // Having an action?
-        foreach ($identities as $identity) {
+        // Next, check ID_TYPE_EMAIL_2FA identity
+        $this->setAuthActionEmail2FA();
+    }
+
+    private function setAuthActionEmailActivate(): bool
+    {
+        $identity = $this->userIdentityModel->getIdentityByType(
+            $this->user,
+            self::ID_TYPE_EMAIL_ACTIVATE
+        );
+
+        if ($identity) {
             $actionClass = setting('Auth.actions')[$identity->name];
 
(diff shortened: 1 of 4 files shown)


MGatner:
This is looking good! I will wait for a review request.

src/Authentication/Authenticators/Session.php
```diff
@@ -149,6 +152,9 @@ public function attempt(array $credentials): Result
         // Update the user's last used date on their password identity.
         $user->touchIdentity($user->getEmailIdentity());
 
+        // Check ID_TYPE_EMAIL_ACTIVATE identity
+        $hasEmailActivate = $this->setAuthActionEmailActivate();
```
datamweb, line 156: @kenjis , let's say the user uses his own custom actions, for example `sms_activate` , `sms_2fa` , have you considered this?
Or am I wrong?

kenjis: This PR only fixes the bugs. Refactoring will be done in the next PR #397.

pjsde (approved):

MGatner (approved):
Looks good as an intermediary step. Very clean and descriptive tests, easy to follow! Thank you.

PR #392, opened 2022-08-19
Refactor record active date

pjsde:
Refactor recordActiveDate() to avoid updating the 'updated_at' column.
Fix https://github.com/codeigniter4/shield/discussions/388

### DIFF
--- a/src/Models/UserModel.php
+++ b/src/Models/UserModel.php
@@ -320,4 +320,14 @@ class UserModel extends Model
 
         return $data;
     }
+
+    /**
+     * Updates the user's last active date.
+     */
+    public function updateActiveDate(User $user): void
+    {
+        $this->builder->set('last_active', $user->last_active)
+            ->where('id', $user->id)
+            ->update();
+    }
 }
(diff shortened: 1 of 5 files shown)


MGatner (approved):
CS Fixer error is unrelated to this PR - probably changes due to a new minor version of the tool. Thanks @pjsde, this looks good to me! See if @kenjis has any final thoughts.

src/Models/UserModel.php
```diff
@@ -320,4 +320,14 @@ protected function saveEmailIdentity(array $data): array
 
         return $data;
     }
+
+    /**
+     * Updates the user's last active date.
+     */
+    public function saveActiveDate(User $user): void
```
kenjis, line 327: `updateActiveDate()` is better?
Because it is just updating only.
*save* implies insert or update.

pjsde: I agree. It's done. I also added one more test that was missing for the `AccessTokens` authenticator

kenjis (approved):

src/Models/UserModel.php
kenjis: Thank you!

MGatner:
Thanks all; glad to see this one in.

Issue #131, opened 2022-05-17
Session Authenticator Event and Logging specifications


kenjis:
The following is a list of the current implementation for Session Authenticator.
Are these okay?
- default login
  - Post email/password
    - OK → event `login` / table `auth_logins`
    - NG → event `failedLogin` / table `auth_logins`
- email2fa login
  1. Post email/password
      - OK → no event / table `auth_logins`
      - NG → event `failedLogin` / table `auth_logins`
  2. Post token
      - OK → event `login`
      - NG → no event
- remember-me login
  - Send remember-me cookie w/o session cookie
    - OK → no event
    - NG → no event
- magic-link login
  1. Post email 
      - OK → no event
      - NG → no event
  2. Send request with token
      - OK → event `login` / table `auth_logins`
      - NG → event `failedLogin` / table `auth_logins`
- default register
  - Post email/username/password
    - OK → event `register` and `login`
    - NG → no event
- register with email activation
  1. Post email/username/password
      - OK → event `register`
      - NG → no event
  2. Post token
      - OK → event `login`
      - NG → no event

kenjis:
MGatner says:
- the term "login" should denote once authentication is complete, i.e. the user has no further Auth steps to take.
- anything that results in an authenticated user should trigger login.

If we agree with these, remember-me and magic-link are also logins, and should fire `login` event?

lonnieezell:
Magic Link is definitely a login. Remember isn't, though, since it's an automatic thing and not something directly initialized by the user. I don't think that whatever someone might want to do after a login is the same thing they would want to do every time the load the page. 

- email2fa login / Post email/password  - I am probably confused about when exactly this is fired, but I thought we already added an event here? If this is after they complete the 2FA then yes a login event should be triggered. 
- Magic links should have login and failedLogin events

Other than that I'm fine with the rest. If we want a failedRegister event to make things symmetrical, that's fine. I can't really think of a use for it, though.

MGatner:
Wouldn't remember-me login only fire after the Session expired, and the long-term cookie was used to "refresh"? I don't have strong opinions on that either way, but it might be useful to have login records when that happens.

lonnieezell:
> Wouldn't remember-me login only fire after the Session expired, and the long-term cookie was used to "refresh"? I don't have strong opinions on that either way, but it might be useful to have login records when that happens.

[...]

PR #397, opened 2022-08-20
refactor: Auth Actions (action classes are executed in the order set in `Config/Auth::$actions`)

kenjis:
~~Need to rebase after merging #391~~

Change the behavior of auth actions based on the following assumptions:
- All actions must be processed one at a time in sequence
- Action classes are executed in the order set in `Config/Auth::$actions`
- All actions have identities


MGatner:
I'll need to come back to these two another day.

MGatner:
Awaiting rebase.

MGatner (approved):
Looks good to me. I'm still not actually using Shield in any projects so let's be sure to get another review from someone. @datamweb and/or @lonnieezell?

datamweb (approved):
@kenjis code works fine, and I didn't see any problems.
**Thanks**
During the review of this PR, I noticed an issue, which I thought was better to raise here.
1. Register a new user.By default, every user login the system after registration.
2 . Check the table with command `php spark db:table test__auth_logins`.
Why is the user's login not recorded in the table?!!
It seems that we need PR to cover this issue.

kenjis:
> Why is the user's login not recorded in the table?!!

It is intentional.
Because it is not a login to record. It is a registration.
A login to record is a user action to login.
I think that a user does not think they are logged in when they register.

The current specification is stated in #131

datamweb:
> It is intentional. 

Thanks Kenjis, I will check #131.

kenjis:
Updated the docs.

kenjis:
I refactored action classes to be more extensible. 

Now you can probably extend the action class to use time limited multiple codes:
https://github.com/codeigniter4/shield/discussions/343#discussioncomment-3416568

MGatner (approved):
Looks good! I'm glad to see some of the ambiguity cleared up. I can't keep track of what has been released... Do we need to start a changelog for some of these behavior changes?

datamweb (changes requested):
I think the following change will help the readability of the documents(`auth_actions.md`)
1. **Email-based Account Activation** 
2. **Email-based Two Factor Authentication** 

I prefer to check this PR locally more carefully so I need more time.
If you are not in a hurry to merge, wait a little.

tests/Unit/EmailActivatorTest.php
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * @internal
+ */
+final class EmailActivatorTest extends TestCase
+{
+    public function testShowCannotGetPendingLoginUser(): void
+    {
+        $this->expectException(RuntimeException::class);
+        $this->expectExceptionMessage('Cannot get the pending login User');
```
datamweb, line 20:
```suggestion
        $this->expectExceptionMessage('Cannot get the pending login User.');
```

tests/Unit/EmailActivatorTest.php
```diff
@@ -0,0 +1,37 @@
...
+
+        $activator = new EmailActivator();
+
+        $activator->show();
+    }
+
+    public function testVerifyCannotGetPendingLoginUser(): void
+    {
+        $this->expectException(RuntimeException::class);
+        $this->expectExceptionMessage('Cannot get the pending login User');
```
datamweb, line 30:
```suggestion
        $this->expectExceptionMessage('Cannot get the pending login User.');
```

datamweb (changes requested):

tests/Unit/EmailActivatorTest.php
kenjis: Done.

kenjis:
@datamweb 
> I prefer to check this PR locally more carefully so I need more time.

Okay. I don't hurry.

datamweb (approved):
Well! 
After hours of playing with the code, I found no problem.

kenjis:
@datamweb Thank you for your testing!

PR #398, opened 2022-08-21
refactor: TokenAuth

kenjis:
Cannot find this code from the `recordActiveDate()` method definition on IDE.


MGatner (approved):

MGatner:
(I'm fine with you YOLO-merging tweaks like this if you're ever comfortable with that.)

kenjis:
It is important to me that the IDE be able to find all the usage points of a method.

If there is one usage point that the IDE doesn't find, I probably forget to change it when refactoring, which could lead to a bug.

PR #401, opened 2022-08-24
docs: improve install.md

kenjis:
- fix Markdown format
- move the Migration step

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -15,7 +15,7 @@ These instructions assume that you have already [installed the CodeIgniter 4 app
 ## Requirements
 
 - [Composer](https://getcomposer.org)
-- Codeigniter v4.2.3 or later
+- Codeigniter **v4.2.3** or later
 - A created database that you can access via the Spark CLI script
 
 ## Composer Installation
--- a/docs/install.md
+++ b/docs/install.md
@@ -40,21 +40,21 @@ If you get the following error:
 
 1. Add the following to change your [minimum-stability](https://getcomposer.org/doc/articles/versions.md#minimum-stability) in your project `composer.json`:
 
-    ```
+    ```json
         "minimum-stability": "dev",
         "prefer-stable": true,
     ```
 
 2. Or specify an explicit version:
 
-    ```
+    ```console
     > composer require codeigniter4/shield:dev-develop
     ```
 
     The above specifies `develop` branch.
     See https://getcomposer.org/doc/articles/versions.md#branches
 
-    ```
+    ```console
     > composer require codeigniter4/shield:^1.0.0-beta
     ```
 
@@ -67,7 +67,7 @@ If you get the following error:
 
 1. Run the following command. This command handles steps 1-5 of *Manual Setup* and runs the migrations.
 
-    ```
+    ```console
     > php spark shield:setup
     ```
 
@@ -117,7 +117,7 @@ your project.
 
 5. **Migration** Run the migrations.
 
-    ```
+    ```console
     > php spark migrate --all
     ```
 
--- a/docs/install.md
+++ b/docs/install.md
@@ -1,9 +1,9 @@
 # Installation
 
 - [Installation](#installation)
+  - [Requirements](#requirements)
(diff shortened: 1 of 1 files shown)


datamweb (approved):
Suggestion, you can do it or pass it:

Codeigniter **v4.2.3 or later**

docs/install.md
```diff
@@ -37,60 +40,36 @@ If you get the following error:
 
 1. Add the following to change your [minimum-stability](https://getcomposer.org/doc/articles/versions.md#minimum-stability) in your project `composer.json`:
 
-```
-    "minimum-stability": "dev",
-    "prefer-stable": true,
-```
+    ```
```
MGatner, line 43: Is it worth highlighting partials?
```suggestion
    ```json
```

docs/install.md
```diff
@@ -37,60 +40,36 @@ If you get the following error:
...
+    > composer require codeigniter4/shield:dev-develop
+    ```
 
-When you run `spark migrate --all`, if you get `Class "SQLite3" not found` error:
+    The above specifies `develop` branch.
+    See https://getcomposer.org/doc/articles/versions.md#branches
 
-1. Remove sample migration files in `tests/_support/Database/Migrations/`
-2. Or install `sqlite3` php extension
+    ```
```
MGatner, line 57:
```suggestion
    ```shell
```

docs/install.md
```diff
@@ -37,60 +40,36 @@ If you get the following error:
...
 
 ### Command Setup
 
-1. Run the following command. This command handles steps 1-4 of *Manual Setup* and runs the migrations.
+1. Run the following command. This command handles steps 1-5 of *Manual Setup* and runs the migrations.
 
-```
-> php spark shield:setup
-```
+    ```
```
MGatner, line 70:
```suggestion
    ```shell
```

docs/install.md
```diff
@@ -99,43 +78,60 @@ your project.
...
+    public function initController(RequestInterface $request, ResponseInterface $response, LoggerInterface $logger)
+    {
+        $this->helpers = array_merge($this->helpers, ['setting']);
 
-    // Do Not Edit This Line
-    parent::initController($request, $response, $logger);
-}
-```
+        // Do Not Edit This Line
+        parent::initController($request, $response, $logger);
```
MGatner, line 104: Side note... I've always hated that we have this there. No idea why there wasn't just a dedicated initialization function.

docs/install.md
```diff
@@ -99,43 +78,60 @@ your project.
...
-```
+    ```php
+    service('auth')->routes($routes);
+    ```
 
 4. **Security Setup** Set `Config\Security::$csrfProtection` to `'session'` (or set `security.csrfProtection = session` in your `.env` file) for security reasons, if you use Session Authenticator.
 
+5. **Migration** Run the migrations.
+
+    ```
```
MGatner, line 120:
```suggestion
    ```shell
```

MGatner (changes requested):
Generally, what does indenting the triple-backtick blocks do? I thought those were supposed to be fully left-aligned, but it looks like they work either way.

docs/install.md
kenjis: Done.

docs/install.md
kenjis: What do you mean?
The Controller should have another initialization method?

MGatner: Right. There's absolutely no benefit to having those parameters exposed and adding Aline that developers shouldn't touch. Call `initController()` (and make it `final`) and then add and call a blank `initialize()` method to `CodeIgniter\Controller` and extend that in `BaseController` instead. Bam: no more weird code in userland. 

Something for v5. 🤷‍♂️

MGatner (approved):

MGatner:
Still curious if there is an advantage to indenting triple backticks. 👆

kenjis:
Before:
Screenshot 2022-08-25 19 00 26

After:
Screenshot 2022-08-25 19 00 36

PR #331, opened 2022-08-02
chore: add multiple db platforms in GA workflow PHPUnit

kenjis:
Probably now Shield works with only ~~MySQL or~~ SQLite3.

- [x] MySQL #336
- [x] Postgre #339
- [x] SQLSRV
- [x] OCI8 #402

PR #402, opened 2022-08-25
feat: support other databases

kenjis:
Ref #331

- fixes for other databases.

Support:
- OCi8
- MySQLi
- Postgre
- SQLSRV

### DIFF
--- a/src/Models/LoginModel.php
+++ b/src/Models/LoginModel.php
@@ -58,6 +58,10 @@ class LoginModel extends Model
     ): void {
         $this->disableDBDebug();
 
+        if ($this->db->getPlatform() === 'OCI8' && $identifier === '') {
+            $identifier = ' ';
+        }
+
         $return = $this->insert([
             'ip_address' => $ipAddress,
             'user_agent' => $userAgent,
--- a/src/Models/LoginModel.php
+++ b/src/Models/LoginModel.php
@@ -56,6 +56,8 @@ class LoginModel extends Model
         ?string $userAgent = null,
         $userId = null
     ): void {
+        $this->disableDBDebug();
+
         $return = $this->insert([
             'ip_address' => $ipAddress,
             'user_agent' => $userAgent,
--- a/src/Entities/AccessToken.php
+++ b/src/Entities/AccessToken.php
@@ -20,6 +20,7 @@ class AccessToken extends Entity
      * @var array<string, string>
      */
     protected $casts = [
+        'id'           => '?integer',
         'last_used_at' => 'datetime',
         'extra'        => 'array',
         'expires'      => 'datetime',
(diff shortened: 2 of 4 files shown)


src/Models/LoginModel.php
```diff
@@ -56,6 +56,12 @@ public function recordLoginAttempt(
         ?string $userAgent = null,
         $userId = null
     ): void {
+        $this->disableDBDebug();
```
MGatner, line 59: Can you explain the need for this?

src/Models/LoginModel.php
```diff
@@ -56,6 +56,12 @@ public function recordLoginAttempt(
...
+
+        if ($this->db->getPlatform() === 'OCI8' && $identifier === '') {
+            $identifier = ' ';
```
MGatner, line 62: ... and this?

MGatner: Ah I see the whole commit message now, that explains this one. Seems like this is probably a bug in the framework driver itself, right?

src/Models/LoginModel.php
kenjis: `ErrorException` was thrown on Oracle database because of the SQL error.

src/Models/LoginModel.php
kenjis: I'm not sure.
Oracle sees `''` as `NULL`.

datamweb: Oracle Database currently treats a character value with a length of zero as null. However, this may not continue to be true in future releases, and Oracle recommends that you do not treat empty strings the same as nulls.

More [see](https://docs.oracle.com/cd/B19306_01/server.102/b14200/sql_elements005.htm).

kenjis: @datamweb I knew it, but I don't know what should we do in this case.
We try to insert `''`, but Oracle treats it as `NULL` and says "cannot insert NULL"

datamweb: @kenjis  I'm not sure if it's helpful, but take a look if you have time.
https://stackoverflow.com/questions/2185743/is-it-possible-to-store-empty-string-as-a-non-null-value-in-the-database/73396113#73396113

kenjis: > Use an alternate representation, such as a single space character, for an empty string.

It is what I did here.

MGatner: @ytetsuro have any recommendations? Or @sclubricants?

kenjis:
I would like to merge this. Can someone approve?

If a true Oracle user knows better solution, anytime they can send a PR.

datamweb (approved):
LGTM! 
See what @MGatner  finally says.

MGatner (approved):
I pinged the only Oracle users I know (at least that I know that I know). This seems like a bigger question than this PR, but this address the problem at hand so let's move forward.

MGatner:
🥳

PR #409, opened 2022-08-29
fix: add logic to skip filters if not in HTTP

kenjis:
All filters do not make sense in CLI.

Fixes the PHPStan error:

```
Error: Call to an undefined method CodeIgniter\HTTP\RequestInterface::getHeaderLine().
 ------ ----------------------------------------------------- 
  Line   src/Filters/TokenAuth.php                            
 ------ ----------------------------------------------------- 
  43     Call to an undefined method                          
         CodeIgniter\HTTP\RequestInterface::getHeaderLine().  
 ------ ----------------------------------------------------- 
```
https://github.com/codeigniter4/shield/runs/8072975213?check_suite_focus=true

### DIFF
--- a/src/Filters/AuthRates.php
+++ b/src/Filters/AuthRates.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace CodeIgniter\Shield\Filters;
 
 use CodeIgniter\Filters\FilterInterface;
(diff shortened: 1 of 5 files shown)


datamweb (approved):
Everything is new🫣, 
Thanks!

PR #413, opened 2022-08-31
feat: notify devs when user has used magic link login.

lonnieezell:
Provides a temp session variable to allow developers to respond as they choose whenever a user has completed a magic link login.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -131,7 +131,7 @@ public array $actions = [
 
 ### Responding to Magic Link Logins
 
-Magic Link logins allow a user that has forgotten their password that have an email sent with a unique login link that will provide a one-time login for them. Once they've logged in you can decide how to respond. In some cases, you might want to redirect them to a special page where the must choose a new password. In other cases, you might simply want to display a one-time message prompting them to go to their account page and choose a new password there.
+Magic Link logins allow a user that has forgotten their password to have an email sent with a unique, one-time login link. Once they've logged in you can decide how to respond. In some cases, you might want to redirect them to a special page where they must choose a new password. In other cases, you might simply want to display a one-time message prompting them to go to their account page and choose a new password.
 
 #### Session Notification
 
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -146,7 +146,7 @@ if (session('magicLogin')) {
 This value sticks around in the session for 5 minutes. Once you no longer need to take any actions, you might want to delete the value from the session.
 
 ```php
-session()->removeTempData('magicLogin');
+session()->removeTempdata('magicLogin');
 ```
 
 #### Event
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -14,6 +14,8 @@ NOTE: The examples assume that you have run the setup script and that you have c
     - [Enable Account Activation via Email](#enable-account-activation-via-email)
     - [Enable Two-Factor Authentication](#enable-two-factor-authentication)
     - [Responding to Magic Link Logins](#responding-to-magic-link-logins)
+      - [Session Notification](#session-notification)
+      - [Event](#event)
   - [Authorization Flow](#authorization-flow)
     - [Change Available Groups](#change-available-groups)
     - [Set the Default Group](#set-the-default-group)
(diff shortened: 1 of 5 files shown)


docs/quickstart.md
```diff
@@ -13,6 +13,7 @@ NOTE: The examples assume that you have run the setup script and that you have c
     - [Change Access Token Lifetime](#change-access-token-lifetime)
     - [Enable Account Activation via Email](#enable-account-activation-via-email)
     - [Enable Two-Factor Authentication](#enable-two-factor-authentication)
+    - [Responding to Magic Link Logins](#responding-to-magic-link-logins)
```
kenjis, line 16: How do you update the TOC?
We need tool and/or documentation for contributors.

datamweb: I ask you to pay attention to #21 again.
The fact that we do not know about _vuepress static site generater_ is not a good reason for not paying attention to the good performance of #21 
Also, we have already had a request in #373.

kenjis: I just asked how to update the TOC in the markdown file.

I don't need HTML documentation personally.
But if someone wants it, I'm fine.

docs/quickstart.md
```diff
@@ -126,6 +127,24 @@ public array $actions = [
 ];
 ```
 
+### Responding to Magic Link Logins
+
+Magic Link logins allow a user that has forgotten their password that have an email sent with a unique login link that will provide a one-time login for them. Once they've logged in you can decide how to respond. In some cases, you might want to redirect them to a special page where the must choose a new password. In other cases, you might simply want to display a one-time message prompting them to go to their account page and choose a new password there.
```
MGatner, line 132:
```suggestion
Magic Link logins allow a user that has forgotten their password to have an email sent with a unique, one-time login link. Once they've logged in you can decide how to respond. In some cases, you might want to redirect them to a special page where they must choose a new password. In other cases, you might simply want to display a one-time message prompting them to go to their account page and choose a new password.
```

tests/Authentication/MagicLinkTest.php
```diff
@@ -114,6 +114,9 @@ public function testMagicLinkVerifyExpired(): void
 
         $result->assertRedirectTo(route_to('magic-link'));
         $result->assertSessionHas('error', lang('Auth.magicLinkExpired'));
+
+        // It should have set temp session var
```
MGatner, line 118:
```suggestion
        // It should not have set temp session var
```

MGatner (changes requested):
I'm not opposed to the session variable, but why not just trigger an Event? Or do both?

docs/quickstart.md
lonnieezell: @kenjis I just use the Markdown All-in-One plugin for VS Code.

tests/Authentication/MagicLinkTest.php
lonnieezell: @MGatner I'm fine with both if we want that. My initial instinct was an event, but you can't redirect from an event so you would have to set a session variable or something anyway to know you needed to do something so this saves that step. I couldn't really think of too many actions that would need to be done from an event at the time.

MGatner: I don't think the Event hurts anything. A dev could actually throw a `RedirectException` from the Event trigger to force a redirect - not the best, but it might actually be an appropriate use here.

lonnieezell: @MGatner added the event.

kenjis:
Can you use `git rebase` to update your branch?
See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/workflow.md#updating-your-branch

src/Controllers/MagicLinkController.php
```diff
@@ -163,6 +163,12 @@ public function verify(): RedirectResponse
 
         $this->recordLoginAttempt($identifier, true, $user->id);
 
+        // Give the developer a way to know the user
+        // logged in via a magic link.
+        session()->setTempdata('magic_login', true);
+
+        Events::trigger('magic_login');
```
kenjis, line 170: `failedLogin` and `magic_login`, not consistent. `magicLogin`?

Please update the docs:
https://github.com/codeigniter4/shield/blob/develop/docs/events.md#event-list
https://github.com/codeigniter4/shield/blob/develop/docs/session_auth_event_and_logging.md#login

lonnieezell:
Doc changes pushed.

kenjis:
Rebased and force pushed.

MGatner (approved):
Looks good! I think this addressed all the comments. @kenjis or @datamweb Feel free to merge if you approve.

datamweb (approved):
The test did not cover event `magicLogin`!
Other items are good.

docs/quickstart.md
```diff
@@ -126,6 +129,37 @@ public array $actions = [
...
+```php
+if (session('magicLogin')) {
+    return redirect()->route('set_password');
+}
+```
+
+This value sticks around in the session for 5 minutes. Once you no longer need to take any actions, you might want to delete the value from the session.
+
+```php
+session()->removeTempData('magicLogin');
```
datamweb, line 149:
```suggestion
session()->removeTempdata('magicLogin');
```
Sorry guys, I just saw this.🫣

datamweb (changes requested):

datamweb (approved):
Thanks! 🥀

PR #414, opened 2022-08-31
docs: add Session Authenticator Event and Logging

kenjis:
From #131

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -65,3 +65,9 @@ When the magic link login fails, the following array will be provided:
 #### logout
 
 Fired immediately after a successful logout. The only argument is the `User` entity.
+
+### Event Timing
+
+To learn more about Event timing, please see the list below.
+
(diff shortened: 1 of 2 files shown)


MGatner (approved):
Very helpful!

datamweb:
Just one suggestion, it will be more beautiful with [Markdown Mermaid](https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/)

kenjis:
@datamweb I'm not sure Mermaid can improve this document. 
What diagram?

datamweb:
Ok @kenjis, I will do it if necessary.

Only we have listed all the files in `index.md`, do we need to add this file?

kenjis:
> Only we have listed all the files in index.md, do we need to add this file?

I don't think so. If we does so, when the number of files increases too much, it becomes difficult to find them.

PR #419, opened 2022-09-05
simplify docs in quickstart - managing user

arif-rh:
remove repeated explanation.

PR #421, opened 2022-09-05
Minor fix for docs

arif-rh:
This should close PR #419  and close PR #420


MGatner (approved):
Looks good. Thank you!

datamweb (approved):
LGTM! Thanks :+1:

We have very interesting statistics of Shield installation and visits.
But I always wonder why these visitors and users don't give us stars.
@arif-rh You have contributed to the improvement of the Shield, may I ask why you did not give the Shield a star?

kenjis (approved):

PR #417, opened 2022-09-02
ApiToken docs and filter update.

lonnieezell:
Added the first of a handful of guides stepping through setup and use Access Tokens to handle various situations. This one covers generic API requests. 

Along the way I discovered that the `TokenAuth` filter didn't support checking against a scope, so I updated the filter to handle a single permission added when used as a route filter.

```php
$routes->get('users', 'UserController::list', ['filter' => 'tokens:users-read']);
```

PR #422, opened 2022-09-06
docs: mobile token guides

lonnieezell:
Adds a guide describing one way of using Access Tokens to authenticate a mobile application. 

Requires #417 to be merged first.


MGatner:
Awaiting the other PR and rebase before review.

kenjis:
Rebased.

kenjis:
Added small fixes.

MGatner (approved):
This is rather opinionated but I like it as a practical example of a real and common scenario in webdev.

datamweb:
Is this explanation comprehensive, what about tablets, smart watches and ...?
Wouldn't it be better to use **smart devices** instead of mobile?

lonnieezell:
Thanks for the additional work here @kenjis.

@MGatner The code is mostly taken from the existing login method. 

I still want to look into how we do SPA authentication and provide a guide for that. I feel that might need a new method added, but unsure yet. If we use the session filter for SPAs I think we need to provide a way for an API to get a CSRF token. If we go with tokens for SPA's there needs to be some careful consideration there as that would leave the token exposed to anyone who wanted to look. So would probably have to put the token in a cookie or something to hide it? Unsure at the moment.

kenjis (approved):

PR #423, opened 2022-09-07
refactor: fix static analysis errors

kenjis:


src/Authentication/Authenticators/AccessTokens.php
```diff
@@ -115,6 +115,8 @@ public function check(array $credentials): Result
             ]);
         }
 
+        assert($token->last_used_at instanceof Time || $token->last_used_at === null);
```
kenjis, line 118: I don't know why but without this line, phpstan reports errors.

MGatner: How about adding `@property Time|null $last_used` to the entity?

MGatner (approved):

src/Authentication/Authenticators/AccessTokens.php
datamweb: Although I don't like doing anything without understanding the reason, #417 seems to be waiting for this PR.

datamweb (approved):

PR #427, opened 2022-09-10
docs: add some important shield features to `README.md`

datamweb:


MGatner (changes requested):
This is a great overview, but it feels too long for the top of the docs "landing page". I would recommend trying to cut it in half and make it bullet points instead of numbers; we can still find a place for the full list on another page.

datamweb:
> This is a great overview, but it feels too long for the top of the docs "landing page". I would recommend trying to cut it in half and make it bullet points instead of numbers; we can still find a place for the full list on another page.

I did it, but I think it's still not good! What do you think?

MGatner (approved):
Yes I think that's much better! It highlights some of our coolest features but still fits comfortably on the page. Thank you!

kenjis (approved):

PR #432, opened 2022-09-10
docs: Missing apostrophe

iRedds:
Missing apostrophe in getVar() method parameter

```php
$routes->get('/access/token', static function() {
    $token = auth()->user()->generateAccessToken(service('request')->getVar('token_name));

    return json_encode(['token' => $token->raw_token]);
});
```


datamweb (approved):
Welcome! Thanks👍

kenjis (approved):

PR #434, opened 2022-09-13
chore: add support links

datamweb:
This PR adds `support ` ~~and `funding `~~ links.

You all work hard for Codeigniter community. 
:green_heart: **Happy Programmers Day!** :green_heart:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -79,5 +79,13 @@
         "mutate": "infection --threads=2 --skip-initial-tests --coverage=build/phpunit",
         "style": "@cs-fix",
         "test": "phpunit"
+    },
+    "support": {
+        "forum": "https://forum.codeigniter.com",
+        "chat": "https://github.com/codeigniter4/shield/discussions",
+        "slack": "https://codeigniterchat.slack.com",
+        "source": "https://github.com/codeigniter4/shield",
+        "issues": "https://github.com/codeigniter4/shield/issues",
+        "docs": "https://github.com/codeigniter4/shield/blob/develop/docs/index.md"
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -81,8 +81,7 @@
         "test": "phpunit"
     },
     "support": {
-        "forum": "https://forum.codeigniter.com",
-        "chat": "https://github.com/codeigniter4/shield/discussions",
+        "forum": "https://github.com/codeigniter4/shield/discussions",
         "slack": "https://codeigniterchat.slack.com",
         "source": "https://github.com/codeigniter4/shield",
         "issues": "https://github.com/codeigniter4/shield/issues",


composer.json
```diff
@@ -79,5 +79,23 @@
...
+    },
+    "support": {
+        "forum": "https://forum.codeigniter.com",
+        "chat": "https://github.com/codeigniter4/shield/discussions",
+        "slack": "https://codeigniterchat.slack.com",
+        "source": "https://github.com/codeigniter4/shield",
+        "issues": "https://github.com/codeigniter4/shield/issues",
+        "docs": "https://github.com/codeigniter4/shield/blob/develop/docs/index.md"
+    },
+    "funding": [
```
lonnieezell, line 91: While I developed the vast majority of this, I'm not comfortable having my info on there for funding now that it is an official project. Unfortunately we don't currently have any funding options setup for the framework as a whole. Please remove this section.

datamweb: Sorry @lonnieezell , this performance was due to my lack of knowledge.
Done.

composer.json
```diff
@@ -79,5 +79,13 @@
         "mutate": "infection --threads=2 --skip-initial-tests --coverage=build/phpunit",
         "style": "@cs-fix",
         "test": "phpunit"
+    },
+    "support": {
+        "forum": "https://forum.codeigniter.com",
+        "chat": "https://github.com/codeigniter4/shield/discussions",
```
kenjis, line 85: GitHub Discussions do not seem to be *chat*. It is *forum* ?

datamweb: Actually I don't know how to add two forum.
It is not explained here:
https://getcomposer.org/doc/04-schema.md#support

kenjis: You can't. `support.forum` should be a string. If you set an array, it raises a validation error.

datamweb: So which one do you choose between `https://forum.codeigniter.com` and `https://github.com/codeigniter4/shield/discussions`?

MGatner: I'm in favor of Discussions. The forums are kind of a mess of support issues and anything we can do to split them out a bit more might be good.

MGatner: @kenjis answers more support forum threads than me though, so he may disagree 😉

datamweb: Maybe this conversation will be useful for you in making a decision. Some people still have problems with the forum.

Screenshot_2022-09-14-15-23-22-467_com miui gallery

kenjis: I'm fine with GitHub discussions.
To be honest, I like GitHub because it is easier to read the source code than the forum.

datamweb: Good feedback.
Done!

kenjis (approved):

composer.json
MGatner: Our forums are hard to use, on many fronts. It's pretty old software, and even though it has been kept up it has that "patched legacy churn" feel. The team keeps throwing around the idea of building a CI4 forum software to replace it, as a group project, but everyone is already very busy and the framework takes lots of time so I don't know if that will happen.

MGatner (approved):

PR #439, opened 2022-09-16
fix: strong_password check

### LINKED ISSUE #438: Bug: ValidationRules strong_password is not working when called using tokens in before filters API routes.
### PHP Version

8.1.0

### CodeIgniter4 Version

4.2.6

### Shield Version

dev-develop

Windows

### Which server did you use?

apache

### Database

MySQL 8.0.27

### Did you customize Shield?

No

### What happened?

I develop API for registration. To secure an API, I implement tokens on before-filters.  When calling registration API, I sent my developer tokens in the Authorization header.

But when I post password contains personal information, it can be passed. `strong_password` valdiation is not working.

### Steps to Reproduce

1. make simple API for registration
2. implements tokens on before-filter to secure the API routes
3. Set the token in the Authorization header
4. post any password contains personal information

### Expected Output

Api should return error that "Passwords cannot contain re-hashed personal information."

### Anything else?

I am new in this library, so this report could be false, maybe I do wrong setup or implementation of this library. But for now, I am thinking, the cause could be this [line](https://github.com/codeigniter4/shield/blob/94dc77354b9cf52ebbf9c986ec236bf58d705ee6/src/Authentication/Passwords/ValidationRules.php#L39):

````php        
    if (function_exists('auth') && auth()->user()) {  
            $user = auth()->user();
    }
````
when I change those line to:
````php        
    if (function_exists('auth') && auth('session')->user()) {  
            $user = auth('session')->user();
    }
````
then `strong_password` validation is working as expected.

### PR DESCRIPTION
**Pull request by arif-rh**
Fixed #438

### DIFF
--- a/src/Authentication/Passwords/ValidationRules.php
+++ b/src/Authentication/Passwords/ValidationRules.php
@@ -36,8 +36,8 @@ class ValidationRules
         /** @var Passwords $checker */
         $checker = service('passwords');
 
-        if (function_exists('auth') && auth()->user()) {
-            $user = auth()->user();
+        if (function_exists('auth') && auth('session')->user()) {
+            $user = auth('session')->user();
         } else {
             $user = empty($data) ? $this->buildUserFromRequest() : $this->buildUserFromData($data);
         }
--- a/tests/Controllers/Api/Users.php
+++ b/tests/Controllers/Api/Users.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /**
  * This file is part of CodeIgniter 4 framework.
  *
@@ -8,6 +10,7 @@
  * For the full copyright and license information, please view
  * the LICENSE file that was distributed with this source code.
  */
+
 namespace Tests\Controllers\Api;
 
 use CodeIgniter\RESTful\ResourceController;
--- a/tests/Controllers/Api/Users.php
+++ b/tests/Controllers/Api/Users.php
@@ -3,12 +3,7 @@
 declare(strict_types=1);
 
 /**
- * This file is part of CodeIgniter 4 framework.
- *
- * (c) CodeIgniter Foundation <admin@codeigniter.com>
- *
- * For the full copyright and license information, please view
- * the LICENSE file that was distributed with this source code.
+ * This file is part of StrongPasswordValidationTest
  */
 
 namespace Tests\Controllers\Api;
--- a/tests/Controllers/Api/Users.php
+++ b/tests/Controllers/Api/Users.php
@@ -1,0 +1,66 @@
+<?php
+
+/**
+ * This file is part of CodeIgniter 4 framework.
+ *
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+namespace Tests\Controllers\Api;
+
+use CodeIgniter\RESTful\ResourceController;
+
+/**
+ * An extendable controller to provide a RESTful API for a resource.
+ */
+class Users extends ResourceController
+{
+    public function signup()
+    {
+        $rules = $this->getValidationRules();
+
+        if (! $this->validate($rules)) {
+            return $this->respond(['errors' => $this->validator->getErrors()]);
+        }
+
+        return $this->respond(['errors' => '']);
+    }
+
+    /**
+     * Returns the rules that should be used for validation.
+     *
+     * @return string[]
+     */
+    protected function getValidationRules(): array
+    {
+        $registrationUsernameRules = array_merge(
+            config('AuthSession')->usernameValidationRules,
+            ['is_unique[users.username]']
+        );
+        $registrationEmailRules = array_merge(
+            config('AuthSession')->emailValidationRules,
+            ['is_unique[auth_identities.secret]']
+        );
+
+        return setting('Validation.registration') ?? [
+            'username' => [
+                'label' => 'Auth.username',
+                'rules' => $registrationUsernameRules,
+            ],
+            'email' => [
+                'label' => 'Auth.email',
+                'rules' => $registrationEmailRules,
+            ],
+            'password' => [
+                'label' => 'Auth.password',
+                'rules' => 'required|strong_password',
+            ],
+            'password_confirm' => [
+                'label' => 'Auth.passwordConfirm',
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by kenjis** [comment]
Thank you for sending a PR.

> We expect all contributions to conform to our [style guide](https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/styleguide.md), be commented (inside the PHP source files), be documented (in the [user guide](https://codeigniter4.github.io/userguide/)), and **unit tested** (in the [test folder](https://github.com/codeigniter4/CodeIgniter4/tree/develop/tests)).
https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md

Please add test code to prove the bug is fixed.

File: src/Authentication/Passwords/ValidationRules.php [hunk 1]
```diff
@@ -36,8 +36,8 @@ public function strong_password(string $value, ?string &$error1 = null, array $d
         /** @var Passwords $checker */
         $checker = service('passwords');
 
-        if (function_exists('auth') && auth()->user()) {
-            $user = auth()->user();
+        if (function_exists('auth') && auth('session')->user()) {
+            $user = auth('session')->user();
```
kenjis, line 40: I think this change is wrong. We don't know what authenticator is used.

And checking `auth()->user()` does not make sense when you register another user.
We need to fix the logic to create `$user` instance.

arif-rh: > And checking and getting auth()->user() does not make sense when you register another user.

I agree. That why I am thinking, maybe we completely remove those `auth()->user()` check? Because when we need to validate strong password, that must be come from post incoming request,  So this line maybe enough for it?

````php
$user = empty($data) ? $this->buildUserFromRequest() : $this->buildUserFromData($data);
````

kenjis: It must work even when only the password is to be changed, such as on the Change Password page.
So it seems to me that we need to retrieve the data of a registered user and overwrite them with the data sent to.

However, for example. If the page is for an administrator to change a user's password, getting the authenticated user's data will not work because the administrator and the user being changed are different accounts.

Probably the correct user ID is needed for `strong_password` to work perfectly.

arif-rh: I dont understand this part
> Probably the correct user ID is needed for strong_password to work perfectly.

Because the scenario of using `strong_password` are:

- Existing user change his own password. Yes, the correct user ID is needed.
- Administrator change other user account password. Yes, the correct user account ID which will be changed is needed.
- Register new user. In this case, I think there is _no correct user ID_ because even the `user` is not exist yet.

So the only thing in my mind is only to compare the posted password with posted personal fields, so `$this->buildUserFromRequest()` and `$this->buildUserFromData($data)` maybe enough for create `user` object to be passed on `NothingPersonalValidation`.

kenjis: > Register new user. In this case, I think there is no correct user ID because even the user is not exist yet.

Yes. The use id does not exist yet.

The current implementation does not seem to anticipate the use case of setting or changing someone else's password. Either the user does not exist (registration) or an authenticated user changes their own password.

> So the only thing in my mind is only to compare the posted password with posted personal fields

Yes, but now there is no way to know the posted data is not the data of the authenticated user.
In other words, `strong_password` does not know the data is for a new user.

MGatner: Consideration: `strong_password` might be doing too much for a validation rule. Should we make a dedicate password service instead and let the rule be more about formatting?

kenjis: That's it!
`strong_password` is already doing too much for a validation rule.
So it is better not to do more.

`strong_password` is for checking the authenticated user's password (or anonymous user's password when registering). It covers most of use cases. And it is not good practice to set another user's password.

But if we need to check another user's password, we can use password service directly in the controller.

@arif-rh Why don't you use `service('passwords')` in  your controller?

arif-rh: @kenjis  using `service('passwords')` directly to check the password is fine.

I create an issue, for people who use the shield and implement `ValdiationRules` to validate their controller and using tokens in before filters. 

If it is considered as a bug, we should find a way to fix it. If we have not fond the solution yet, we may need to inform the user in the docs about the situation and give them workaround, like after `validate` the controller, then check again the password using `service('passwords')`.

kenjis: My opinion is like this:
- `strong_password` only supports use cases to check the authenticated user's password, or anonymous user's password (when registering).
- It is not good practice to set another user's password. Because a passwords should be known by a user only. So `strong_password` does not support such a case. In that case, use `service('passwords')` directly.
- Add these in the docs.

arif-rh: > - strong_password only supports use cases to check the authenticated user's password, or anonymous user's password when registering.

and please understand, that an issue I posted is related to above conditions. Let me summarize again the case here:

1. build an API for registering user
2. add `before` filters using `tokens`
3. validate the request from API using `ValidationRules`

and `strong_password` validation is failed, because when using `tokens` in before filters, `auth()->user()` get the user object from the `AccessToken` class, while the user should be _anonymous user_ as you mention. That; is the issue.

Authentications available on shields are either `tokens` or `sessions`. But for validation `strong_password`, it will not working when using `tokens` authentication, instead using `sessions` is working, because:

1. even when using `tokens' in `before` filters, the user object will never be build based on `token` header.
2. when using `sessions` authentication, `strong_password` also still working.

See the unit test I push, it test both for using `sessions` and `tokens` in `before` filter.

kenjis: I mean an anonymous user is a visitor who is not authenticated.
I mean the user with valid AccessToken is not an anonymous user.

kenjis: > even when using tokens in before filters, the user object will never be build based on token header.

? 
I understand when using tokens in before filters, the user object will be build based on token header.

kenjis: > See the unit test I push, it test both for using sessions and tokens in before filter.

I can see only the test (`testUsingTokenBeforeFilter`) with tokens in before filter.

arif-rh:
I will close this since #438 is considered not as a bug.

PR #442, opened 2022-09-17
fix: correct display of the error in case of change `authenticatorHeader`

datamweb:
```php    
public array $authenticatorHeader = [
        'tokens' => 'MyToken',
    ];
```
**Befor:**
Every request must have a bearer token in the **Authorization** header.

**After:**
Every request must have a bearer token in the **MyToken** header.


kenjis (approved):

MGatner (approved):

datamweb:
Add a note in the documentation and change the test commit message.

docs/guides/mobile_apps.md
```diff
@@ -58,3 +58,9 @@ class LoginController extends BaseController
 ```
 
 When making all future requests to the API, the mobile client should return the raw token in the `Authorization` header as a `Bearer` token.
+
+> **Note**
+>
+> By default, `$authenticatorHeader['tokens']` is set to `Authorization`. You can change this value by setting the `$authenticatorHeader['tokens']` value in the `Auth.php` config file.
```
MGatner, line 64:
```suggestion
> By default, `$authenticatorHeader['tokens']` is set to `Authorization`. You can change the header name by setting the `$authenticatorHeader['tokens']` value in the `Auth.php` config file.
```

docs/guides/mobile_apps.md
```diff
@@ -58,3 +58,9 @@ class LoginController extends BaseController
...
+>
+> e.g. if `$authenticatorHeader['tokens']` is set to `PersonalAccessCodes`. the mobile client should return the raw token in the `PersonalAccessCodes` header as a `Bearer` token.
```
MGatner, line 66:
```suggestion
> e.g. if `$authenticatorHeader['tokens']` is set to `PersonalAccessCodes` then the mobile client should return the raw token in the `PersonalAccessCodes` header as a `Bearer` token.
```

MGatner (changes requested):
Looks good! Small proofreading

kenjis (approved):

MGatner (approved):

PR #429, opened 2022-09-10
docs: explanation of how to use the `session` filter to protect pages

datamweb:
Hello friends.
We have answered how to protect routes by several users in several discussions.
However, I recently received a similar message about how to protect the site panel.
I have covered this topic in this PR.

PR #443, opened 2022-09-18
docs: fix protect all url with `$globals`

datamweb:
ref: https://github.com/codeigniter4/shield/discussions/441#discussioncomment-3672007
It seems that in https://github.com/codeigniter4/shield/pull/429#issuecomment-1242676793 I meant to _rewrite_ filter `session`, but I didn't, that was my carelessness, sorry.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -149,8 +149,18 @@ These can be used in any of the [normal filter config settings](https://codeigni
 If you want to limit all routes (e.g. `localhost:8080/admin`, `localhost:8080/panel` and ...), you need to add the following code in the `app\Config\Filters.php` file.
 
 ```php
-public $filters = [
-    'session' => ['except' => ['login*', 'register*']],
+public $globals = [
+    'before' => [
+        // 'honeypot',
+        // 'csrf',
+        // 'invalidchars',
+        'session' => ['except' => ['login*', 'register*']],
+    ],
+    'after' => [
+        'toolbar',
+        // 'honeypot',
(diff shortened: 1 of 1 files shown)


docs/install.md
```diff
@@ -149,8 +149,18 @@ These can be used in any of the [normal filter config settings](https://codeigni
...
+        // 'honeypot',
+        // 'csrf',
+        // 'invalidchars',
+        'session' => ['except' => ['login*', 'register*']],
+    ],
+    'after' => [
+        'toolbar',
+        // 'honeypot',
+        // 'secureheaders',
+    ],
```
kenjis, line 163:
```suggestion
    'before' => [
        // ...
        'session' => ['except' => ['login*', 'register*']],
    ],
    // ...
```

docs/install.md
```diff
@@ -149,8 +149,18 @@ These can be used in any of the [normal filter config settings](https://codeigni
 If you want to limit all routes (e.g. `localhost:8080/admin`, `localhost:8080/panel` and ...), you need to add the following code in the `app\Config\Filters.php` file.
```
kenjis, line 149:
```suggestion
If you want to limit all routes (e.g. `localhost:8080/admin`, `localhost:8080/panel` and ...), you need to add the following code in the `app/Config/Filters.php` file.
```

kenjis (approved):
LGTM!

MGatner (approved):
Thanks for the quick fix!

PR #449, opened 2022-09-20
docs: add note about strong_password

kenjis:
See https://github.com/codeigniter4/shield/pull/439#discussion_r974787536 and https://github.com/codeigniter4/shield/issues/438


MGatner (approved):
A little "best practice" thrown in there, why not?

datamweb (approved):
Good explanation. 
Thanks🥀

kenjis:
> A little "best practice" thrown in there, why not?

What do you mean?
What is the best practice?

PR #452, opened 2022-09-26
fix: bug that when UserIdentity is changed, the correct user identities are not returned afterwards

### LINKED ISSUE #450: Bug: $user->getEmailIdentity() returns null after saving new user
### PHP Version

7.4

### CodeIgniter4 Version

4.2.6

### Shield Version

dev-develop

macOS

### Which server did you use?

apache

### Database

MySQL 5.7

### Did you customize Shield?

No

### What happened?

I think something is wrong here :) Am i right?  
src/Entities/User.php

```php
/**
     * @var UserIdentity[]|null
     */
    private ?array $identities = null; **<- parameter can be either empty array**

............

/**
     * ensures that all of the user's identities are loaded
     * into the instance for faster access later.
     */
    private function populateIdentities(): void
    {
        if ($this->identities === null) { **<- check for null only, if this parameter is empty array does not get identity**
            $identityModel = model(UserIdentityModel::class);

            $this->identities = $identityModel->getIdentities($this);
        }
    }
```

### Steps to Reproduce

n/a

### Expected Output

`$user->getEmailIdentity()` to return Identity.

### Anything else?

_No response_

kenjis:
Fixes #450

- Fix bug that when `UserIdentity` is changed (including when saving User object), the correct user identities does not be returned afterwards.
- Now `UserModel` does not alter the passed `User` object state when saving, ~~but when it is an insertion, add the user id to it.~~

### DIFF
--- a/src/Models/UserIdentityModel.php
+++ b/src/Models/UserIdentityModel.php
@@ -12,6 +12,7 @@ use CodeIgniter\Shield\Authentication\Passwords;
 use CodeIgniter\Shield\Entities\AccessToken;
 use CodeIgniter\Shield\Entities\User;
 use CodeIgniter\Shield\Entities\UserIdentity;
+use CodeIgniter\Shield\Exceptions\LogicException;
 use Faker\Generator;
 
 class UserIdentityModel extends Model
@@ -59,7 +60,7 @@ class UserIdentityModel extends Model
      */
     public function createEmailIdentity(User $user, array $credentials): void
     {
-        assert($user->id !== null, '"$user->id" must not be null.');
+        $this->checkUserId($user);
 
         /** @var Passwords $passwords */
         $passwords = service('passwords');
@@ -72,6 +73,15 @@ class UserIdentityModel extends Model
         ]);
 
         $this->checkQueryReturn($return);
+    }
+
+    private function checkUserId(User $user): void
+    {
+        if ($user->id === null) {
+            throw new LogicException(
+                '"$user->id" is null. You should not use the incomplete User object.'
+            );
+        }
     }
 
     /**
@@ -87,7 +97,7 @@ class UserIdentityModel extends Model
         array $data,
         callable $codeGenerator
     ): string {
-        assert($user->id !== null, '"$user->id" must not be null.');
+        $this->checkUserId($user);
 
         helper('text');
 
@@ -122,7 +132,7 @@ class UserIdentityModel extends Model
      */
     public function generateAccessToken(User $user, string $name, array $scopes = ['*']): AccessToken
     {
-        assert($user->id !== null, '"$user->id" must not be null.');
+        $this->checkUserId($user);
 
         helper('text');
 
@@ -157,7 +167,7 @@ class UserIdentityModel extends Model
 
     public function getAccessToken(User $user, string $rawToken): ?AccessToken
     {
-        assert($user->id !== null, '"$user->id" must not be null.');
+        $this->checkUserId($user);
 
         return $this->where('user_id', $user->id)
             ->where('type', AccessTokens::ID_TYPE_ACCESS_TOKEN)
@@ -173,7 +183,7 @@ class UserIdentityModel extends Model
      */
     public function getAccessTokenById($id, User $user): ?AccessToken
     {
-        assert($user->id !== null, '"$user->id" must not be null.');
+        $this->checkUserId($user);
 
         return $this->where('user_id', $user->id)
             ->where('type', AccessTokens::ID_TYPE_ACCESS_TOKEN)
@@ -187,7 +197,7 @@ class UserIdentityModel extends Model
      */
     public function getAllAccessTokens(User $user): array
     {
-        assert($user->id !== null, '"$user->id" must not be null.');
+        $this->checkUserId($user);
 
         return $this
             ->where('user_id', $user->id)
@@ -218,7 +228,7 @@ class UserIdentityModel extends Model
      */
     public function getIdentities(User $user): array
     {
-        assert($user->id !== null, '"$user->id" must not be null.');
+        $this->checkUserId($user);
 
         return $this->where('user_id', $user->id)->orderBy($this->primaryKey)->findAll();
     }
@@ -238,7 +248,7 @@ class UserIdentityModel extends Model
      */
     public function getIdentityByType(User $user, string $type): ?UserIdentity
     {
-        assert($user->id !== null, '"$user->id" must not be null.');
+        $this->checkUserId($user);
 
         return $this->where('user_id', $user->id)
             ->where('type', $type)
@@ -255,7 +265,7 @@ class UserIdentityModel extends Model
      */
     public function getIdentitiesByTypes(User $user, array $types): array
     {
-        assert($user->id !== null, '"$user->id" must not be null.');
+        $this->checkUserId($user);
 
         if ($types === []) {
             return [];
@@ -281,7 +291,7 @@ class UserIdentityModel extends Model
 
     public function deleteIdentitiesByType(User $user, string $type): void
     {
-        assert($user->id !== null, '"$user->id" must not be null.');
+        $this->checkUserId($user);
 
         $return = $this->where('user_id', $user->id)
             ->where('type', $type)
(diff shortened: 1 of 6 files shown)


src/Models/UserIdentityModel.php
```diff
@@ -59,6 +59,8 @@ public function create($data): void
      */
     public function createEmailIdentity(User $user, array $credentials): void
     {
+        assert($user->id !== null, '"$user->id" must not be null.');
```
MGatner, line 62: Any reason not to use an Assert library to make these production-level assertions?

src/Models/UserModel.php
```diff
@@ -225,12 +225,18 @@ public function activate(User $user): void
...
 
         $result = parent::insert($data, $returnID);
 
         $this->checkQueryReturn($result);
 
+        // Set user id to the passed User object.
+        if ($data instanceof User) {
+            $data->id = $this->insertID;
+        }
+
```
MGatner, line 239: I'm not a fan of modifying the object at all; it gives the false impression of database equivalency. We have no created/updated timestamps, or other fields coming from the database, or effects of stored protocols and field defaults. If we really need the ID on the object instead of just in the return would should reload the whole thing.

MGatner (changes requested):

src/Models/UserIdentityModel.php
kenjis: What is an Assert library?

src/Models/UserModel.php
kenjis: > We have no created/updated timestamps, or other fields coming from the database, or effects of stored protocols and field defaults.

Exactly.
We should discourage the use of User objects that are not fetched from the database?
We should think the following code is just a misuse?

```php
        $user           = new User();
        $user->username = 'foo';
        $user->email    = 'foo@bar.com';
        $user->password = 'password';
        $user->active   = false;

        $users->save($user);

        $identity = $user->getEmailIdentity();
```

I added the set use id logic because if I remove it, the test fails:
```
There was 1 failure:

1) Tests\Unit\UserModelTest::testSaveNewUserAndGetEmailIdentity
"$user->id" must not be null. in /Users/kenji/work/codeigniter/official/codeigniter-shield/src/Models/UserIdentityModel.php:221

Caused by
AssertionError: "$user->id" must not be null. in /Users/kenji/work/codeigniter/official/codeigniter-shield/src/Models/UserIdentityModel.php:221
Stack trace:
#0 /Users/kenji/work/codeigniter/official/codeigniter-shield/src/Models/UserIdentityModel.php(221): assert(false, '"$user->id" mus...')
#1 /Users/kenji/work/codeigniter/official/codeigniter-shield/src/Entities/User.php(81): CodeIgniter\Shield\Models\UserIdentityModel->getIdentities(Object(CodeIgniter\Shield\Entities\User))
#2 /Users/kenji/work/codeigniter/official/codeigniter-shield/src/Entities/User.php(95): CodeIgniter\Shield\Entities\User->populateIdentities()
#3 /Users/kenji/work/codeigniter/official/codeigniter-shield/src/Entities/User.php(66): CodeIgniter\Shield\Entities\User->getIdentities('email_password')
#4 /Users/kenji/work/codeigniter/official/codeigniter-shield/src/Entities/User.php(134): CodeIgniter\Shield\Entities\User->getIdentity('email_password')
#5 /Users/kenji/work/codeigniter/official/codeigniter-shield/tests/Unit/UserModelTest.php(75): CodeIgniter\Shield\Entities\User->getEmailIdentity()
#6 /Users/kenji/work/codeigniter/official/codeigniter-shield/vendor/phpunit/phpunit/src/Framework/TestCase.php(1545): Tests\Unit\UserModelTest->testSaveNewUserAndGetEmailIdentity()
#7 /Users/kenji/work/codeigniter/official/codeigniter-shield/vendor/phpunit/phpunit/src/Framework/TestCase.php(1151): PHPUnit\Framework\TestCase->runTest()
#8 /Users/kenji/work/codeigniter/official/codeigniter-shield/vendor/phpunit/phpunit/src/Framework/TestResult.php(726): PHPUnit\Framework\TestCase->runBare()
#9 /Users/kenji/work/codeigniter/official/codeigniter-shield/vendor/phpunit/phpunit/src/Framework/TestCase.php(903): PHPUnit\Framework\TestResult->run(Object(Tests\Unit\UserModelTest))
#10 /Users/kenji/work/codeigniter/official/codeigniter-shield/vendor/phpunit/phpunit/src/Framework/TestSuite.php(672): PHPUnit\Framework\TestCase->run(Object(PHPUnit\Framework\TestResult))
#11 /Users/kenji/work/codeigniter/official/codeigniter-shield/vendor/phpunit/phpunit/src/Framework/TestSuite.php(672): PHPUnit\Framework\TestSuite->run(Object(PHPUnit\Framework\TestResult))
#12 /Users/kenji/work/codeigniter/official/codeigniter-shield/vendor/phpunit/phpunit/src/Framework/TestSuite.php(672): PHPUnit\Framework\TestSuite->run(Object(PHPUnit\Framework\TestResult))
#13 /Users/kenji/work/codeigniter/official/codeigniter-shield/vendor/phpunit/phpunit/src/TextUI/TestRunner.php(673): PHPUnit\Framework\TestSuite->run(Object(PHPUnit\Framework\TestResult))
#14 /Users/kenji/work/codeigniter/official/codeigniter-shield/vendor/phpunit/phpunit/src/TextUI/Command.php(143): PHPUnit\TextUI\TestRunner->run(Object(PHPUnit\Framework\TestSuite), Array, Array, true)
#15 /Users/kenji/work/codeigniter/official/codeigniter-shield/vendor/phpunit/phpunit/src/TextUI/Command.php(96): PHPUnit\TextUI\Command->run(Array, true)
#16 /Users/kenji/work/codeigniter/official/codeigniter-shield/vendor/phpunit/phpunit/phpunit(98): PHPUnit\TextUI\Command::main()
#17 /Users/kenji/work/codeigniter/official/codeigniter-shield/vendor/bin/phpunit(123): include('/Users/kenji/wo...')
#18 {main}
```

kenjis: I removed the logic.

kenjis:
I've changed my mind.

The following code is a misuse. Now it causes an exception.

```php
$user           = new User();
$user->username = 'foo';
$user->email    = 'foo@bar.com';
$user->password = 'password';
$user->active   = false;

$users->save($user);

// $user is not fetched from the database.
$identity = $user->getEmailIdentity();
```

Because CI4's `Model::save()` does not alter the passed object state.

We should not use incomplete User objects.
If we don't get User objects from the database,
these have no created/updated timestamps, or other fields coming from the database,
or effects of stored procedures and field defaults.

So if you do `new User()`, you should save it and get again.

```php
// To get the complete user object, we need to get from the database
$user = $users->findById($users->getInsertID());
```

kpeu3u (approved):

src/Models/UserIdentityModel.php
MGatner: It's basically the same as `assert()` but with same nice built-ins and (mostly importantly) it will throw exceptions in production. They also play nice with static analysis. These are the two ones I know about: Webmozart is so popular I would guess that it is probably already in most Composer vendor directories due to dependencies. 
* https://packagist.org/packages/webmozart/assert
* https://github.com/beberlei/assert

MGatner (approved):
Yes I like this approach much better! Entities without identity are not _technically_ entities and shouldn't be used other than passing to persistence. [When to assign the ID](https://matthiasnoback.nl/2018/05/when-and-where-to-determine-the-id-of-an-entity/) is always tricky for auto-incrementing tables.

src/Models/UserIdentityModel.php
kenjis: Thank you. I have seen webmozart/assert.

> Any reason not to use an Assert library to make these production-level assertions?

- If the only goal is to have the error occur even in a production environment, this can be accomplished with the PHP configuration. Also, if an error should be raised in production environment, an exception should be thrown, not `assert()`.
- One should be cautious about increasing the number of dependent packages in a CI project. We cannot easily add more dependent libraries.
- The current code base does not make heavy use of assertions that an assertion library is necessary, and there are currently no plans to do so.

At the moment, I do not see the need to introduce an Assertion library.

PR #463, opened 2022-10-03
Add alerts dismiss

treborin:
Added missing 'pagestcript' section (for bootstrap.bundle.min.js, in this case, but you can add more .js files there). already inside layout.php. This alaw you to dismiss (close) alerts msgs. with a "X" buttom.
 **FILES**: login.php and register.php.

### DIFF
--- a/src/Views/login.php
+++ b/src/Views/login.php
@@ -10,9 +10,12 @@
                 <h5 class="card-title mb-5"><?= lang('Auth.login') ?></h5>
 
                 <?php if (session('error') !== null) : ?>
-                    <div class="alert alert-danger" role="alert"><?= session('error') ?></div>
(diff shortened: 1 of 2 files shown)


MGatner (changes requested):
I don't like bundling Bootstrap explicitly. People may be using any other frontend solution, or may be adding Bootstrap themselves and this would conflict. It also creates an inline version dependency that we need to keep up without package management. 

This should be a very easy thing to write in native JS, if it is a desired feature.

datamweb:
Hi @treborin 
Thank you for your excellent PR.
Shield allows the implementation of customized views for any developer. 
Therefore, I assume that in a real project, all views are definitely customized.

On the other hand, there is a nice package [here ](https://github.com/tattersoftware/codeigniter4-alerts) that any developer can use without much time.

We will wait for you in another PR.
So thank you.

PR #467, opened 2022-10-05
docs: update badge links in README

kenjis:


datamweb (approved):
I was thinking about where the problem comes from😄.
Thanks👍👍

PR #489, opened 2022-10-30
feat: add `shield:publish` CLI Command

datamweb:
Hello friends,
This PR helps to create custom views and adjust it. (related to: #480 )
It also helps to create custom UserModel. (related to: #476)

- `php spark shield:publish view`
- `php spark shield:publish model`

I'm not sure if merging this command is really necessary, however if you think it might be useful please give it a first look.

### DIFF
--- a/src/Commands/Publish.php
+++ b/src/Commands/Publish.php
@@ -1,0 +1,202 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CodeIgniter\Shield\Commands;
+
+use CodeIgniter\CLI\BaseCommand;
+use CodeIgniter\CLI\CLI;
+use CodeIgniter\Shield\Commands\Setup\ContentReplacer;
+
+class Publish extends BaseCommand
+{
+    /**
+     * The group the command is lumped under
+     * when listing commands.
+     *
+     * @var string
+     */
+    protected $group = 'Shield';
+
+    /**
+     * The Command's name
+     *
+     * @var string
+     */
+    protected $name = 'shield:publish';
+
+    /**
+     * the Command's short description
+     *
+     * @var string
+     */
+    protected $description = 'It helps you to order views or model.';
+
+    /**
+     * the Command's usage
+     *
+     * @var string
+     */
+    protected $usage = 'shield:publish [<name>]';
+
+    /**
+     * the Command's Arguments
+     *
+     * @var array<string, string>
+     */
+    protected $arguments = [
+        'view'  => 'If you plan to customize the Shield views.',
+        'model' => 'If you plan to customize the UserModel.',
+    ];
+
+    protected $defaultViewsPath = VENDORPATH . 'codeigniter4/shield/src/Views';
+    private ContentReplacer $replacer;
+
+    /**
+     * Displays the help for the spark cli script itself.
+     */
+    public function run(array $params): void
+    {
+        $this->replacer = new ContentReplacer();
+
+        if (current($params) === 'view') {
+            $this->publishViews();
+
+            exit;
+        }
+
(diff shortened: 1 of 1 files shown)


kenjis:
Generator? not publisher. If so, `shield:make` is better?
https://codeigniter4.github.io/CodeIgniter4/cli/cli_generators.html
https://codeigniter4.github.io/CodeIgniter4/libraries/publisher.html

datamweb:
**Kenjis**, you are right. I don't agree with the name `shield:publish` either, to be honest, I already chose `shield:custom`.
However, `shield:make` is good, but you have to be careful, in `make:...` , the adjustment work is never done, here after the copy views, the adjustment of the views is done.

I was going to use `GeneratorTrait` to write command `shield:publish model`, but `GeneratorTrait` didn't have the necessary flexibility, it seems that some changes must be made in `GeneratorTrait`, which I don't intend to go into.

kenjis:
Ah, the view command is a bit different from `make` command.
`make` generates files from templates.
But the view command copies view files and fixes the Config file.

kenjis:
This is the output of Myth Auth `auth:publish` command.
It seems the view command is the same as the part of `Publish Views?`.

```console
$ php spark auth:publish

CodeIgniter v4.1.5 Command Line Tool - Server Time: 2021-12-09 16:36:40 UTC+09:00

Publish Migration? [y, n]: y
  Created: Database/Migrations/2017-11-20-223112_create_auth_tables.php
  Remember to run `spark migrate -all` to migrate the database.
Publish Models? [y, n]: y
  Created: Models/LoginModel.php
  Created: Models/UserModel.php
Publish Entities? [y, n]: y
  Created: Entities/User.php
Publish Controller? [y, n]: y
  Created: Controllers/AuthController.php
Publish Views? [y, n]: y
  Created: Views/Auth/_footer.php
  Created: Views/Auth/_header.php
  Created: Views/Auth/_message_block.php
  Created: Views/Auth/_navbar.php
  Created: Views/Auth/emails/activation.php
  Created: Views/Auth/emails/forgot.php
  Created: Views/Auth/forgot.php
  Created: Views/Auth/layout.php
  Created: Views/Auth/login.php
  Created: Views/Auth/register.php
  Created: Views/Auth/reset.php
Publish Filters? [y, n]: y
  Created: Filters/LoginFilter.php
  Created: Filters/PermissionFilter.php
  Created: Filters/RoleFilter.php
Publish Config file? [y, n]: y
  Created: Config/Auth.php
Publish Language file? [y, n]: y
  Created: Language/en/Auth.php
```

datamweb:
> But the view command copies view files and fixes the Config file.

Exactly.

> `make` generates files from templates.

Yes, if you want I can write it with `GeneratorTrait`.

datamweb:
> It seems the view command is the same as the part of `Publish Views?`.


I haven't used `auth:publish` before, but that's what I see.
I don't think we need to cover all of this, some of it is in `shield:setup`. However, I want to see what you mean?
If you think there is a need for changes or something should be added, I will wait for you.

kenjis:
I think the following commands are better:

-    `php spark shield:publish view`
-    `php spark shield:make model`

kenjis:
`shield:model` was implemented in #558

datamweb:
@kenjis We have now documented the [customization details for views](https://datamweb.github.io/shield/customization/views/#change-views).
I want to know if we need `php spark shield:publish view` now?

kenjis:
It is assumed that all view files will be customized by developers.

Because of the large number of files, it would be convenient if they can be prepared by one command. 
So I would be happy to have it.
However, I don't mind if it is not in v1.0.0.

PR #498, opened 2022-11-01
fix: remove `username` in $validFields by default

kenjis:
The default configuration does not support `username` as login id.
There is no form input for `username`, and no validation rules for it.

But Shield accepts `username` in the login form (attackers can send any field), 
and sends it to the database without validation.

This is not an exploitable vulnerability, but it is undesirable. Unvalidated user input should not be sent to the database.

Screenshot 2022-11-01 17 28 07

### DIFF
--- a/src/Config/Auth.php
+++ b/src/Config/Auth.php
@@ -223,7 +223,7 @@ class Auth extends BaseConfig
      */
     public array $validFields = [
         'email',
-        'username',
+        // 'username',
     ];
 
     /**
--- a/src/Controllers/RegisterController.php
+++ b/src/Controllers/RegisterController.php
@@ -79,7 +79,7 @@ class RegisterController extends BaseController
         $allowedPostFields = array_merge(
             setting('Auth.validFields'),
             setting('Auth.personalFields'),
-            ['password']
+            array_keys($rules),
         );
         $user = $this->getUserEntity();
         $user->fill($this->request->getPost($allowedPostFields));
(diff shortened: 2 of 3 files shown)


MGatner (approved):
Seems logical. Please wait for a review from @datamweb

datamweb (commented):
@Kenji I already saw someone looking for a method to support both **username** and **email** for login.
By default, Shield accepts only one field for login.
But my idea was that with changes in the front end, for example, with the help of `Vue.js`, the user can identify the input whether it is an **email or a username**.
**Of course, this requires changes in the default validations rules** (`required_without` / `required_with` and `permit_empty`.

I have not done this in practice.

Now my question is, did I think right or not?

If it is true, doesn't this PR create a limitation?

kenjis:
> Now my question is, did I think right or not?

You think right. If you configure correctly, you can send either email or username as a login id.

> If it is true, doesn't this PR create a limitation?

No, this PR just changes the default configuration.
If you set `$validFields` and proper validation rules, you can use  email or username.

datamweb (approved):
Thanks!
I did not see any problem in the registration process.

And login with `username` is Ok.
```console
Data of Table "auth_logins":

+----+------------+--------------------+----------+------------+---------+--------------------+---------+
| id | ip_address | user_agent         | id_type  | identifier | user_id | date               | success |
+----+------------+--------------------+----------+------------+---------+--------------------+---------+
| 26 | ::1        | Mozilla/5.0 (Wi... | username | Pooya      | 3       | 2022-11-01 17:1... | 1       |
+----+------------+--------------------+----------+------------+---------+--------------------+---------+
```

PR #501, opened 2022-11-02
fix: set only validated fields to User Entity in RegisterController

kenjis:
All user input should be validated.


MGatner (approved):

datamweb (approved):
Thanks!

PR #503, opened 2022-11-02
fix: Error exception occurs when user with pending activation tries magic-link login

### LINKED ISSUE #465: Bug: Error exception occurs when a user with pending registration tries to log in.
### PHP Version

7.4

### CodeIgniter4 Version

4.2.6

### Shield Version

develop

Windows

### Which server did you use?

apache

### Database

MariaDB 10.4.18

### Did you customize Shield?

No.

### What happened?

I tried to log in without entering the registration confirmation code from the email, and I got the error message: 

```
The user has identities for action, so cannot complete login. If you want to start to login with auth action, use startLogin() instead. Or delete identities for action in database. user_id: 23680
```

### Steps to Reproduce

1. Activate the register action
2. Enable magic link logins
3. Register, but do not get the confirmation code/token from your email
4. Open a new browser, click on magic link to get a one-time-login link
5. Go to your email address and click on the login link

### Expected Output

I expect that the user should be redirected to the auth token confirmation page with a more-descriptive message stating that:

```
You must complete your registration by confirming the tokens sent to your email address.
```

### Anything else?

_No response_

kenjis:
Fixes #465


src/Language/sk/Auth.php
```diff
@@ -81,6 +81,7 @@
     'emailActivateSubject'  => 'Váš aktivačný kód',
     'emailActivateMailBody' => 'Pomocou nižšie uvedeného kódu aktivujte svoj účet a môžete začať používať stránku.',
     'invalidActivateToken'  => 'Kód bol nesprávny',
+    'needActivate'          => '(To be translated) You must complete your registration by confirming the code sent to your email address.',
```
jozefrebjak, line 84: Registráciu musíte dokončiť potvrdením kódu zaslaného na vašu e-mailovú adresu.

datamweb (commented):
@kenjis, why do we send email to a user who is not active yet through magic link?
Wouldn't it be better to display the error on the **Magic Link page before sending the email**?
_I always have a problem with sending unnecessary emails._

kenjis:
The user has not activated their account, maybe because they have not received the email for activation.

datamweb:
Well, this is not a reason for us to allow the magic link to be sent to it email.
If he did not receive the activation email, he should apply again from 'auth/a/show', not through the **magic link**.

kenjis:
If the session is expired, no way to get `auth/a/show` when they forget the password.
If they know their password, they don't need to go magic link page.
They can go login page.

tests/Controllers/MagicLinkTest.php
```diff
@@ -57,4 +64,51 @@ public function testShowValidateErrorsInMagicLink(): void
 
         $result->assertSessionHas('errors', $expected);
     }
+
```
datamweb, line 67: Maybe it will help in the future `https://github.com/codeigniter4/shield/issues/465`

datamweb (changes requested):

src/Language/sk/Auth.php
kenjis: Done.

tests/Controllers/MagicLinkTest.php
kenjis: Done.

datamweb (approved):

src/Language/fr/Auth.php
```diff
@@ -81,6 +81,7 @@
     'emailActivateSubject'  => 'Votre code d\'activation',
     'emailActivateMailBody' => 'Veuillez utiliser le code suivant pour activer votre compte et commencer à utiliser le site.',
     'invalidActivateToken'  => 'Le code était incorrect.',
+    'needActivate'          => '(To be translated) You must complete your registration by confirming the code sent to your email address.',
```
MGatner, line 84:
```suggestion
    'needActivate'          => 'Complétez votre inscription en confirmant le code envoyé à votre email.',
```

MGatner (changes requested):

MGatner (approved):

PR #504, opened 2022-11-02
fix: UserModel cannot save an array

### LINKED ISSUE #471: Bug: UserModel cannot save an array
### PHP Version

8.1.2

### CodeIgniter4 Version

4.2.7

### Shield Version

`develop`

Linux

### Which server did you use?

apache

### Database

MySQL 8

### Did you customize Shield?

No

### What happened?

Calling `model(UserModel::class)->save()` with a `array` representing an existing user causes `$tempUser` to be `null`: https://github.com/codeigniter4/shield/blob/f4cdfb672b600a032a6f0bfc0b7735411bee0cae/src/Models/UserModel.php#L250

Which makes the call to `saveEmailIdentity()` fail: https://github.com/codeigniter4/shield/blob/f4cdfb672b600a032a6f0bfc0b7735411bee0cae/src/Models/UserModel.php#L261

### Steps to Reproduce

This is obviously contrived but should cause the error:
```php
$user = fake(UserModel::class);
$user->email = 'test@example.com';

model(UserModel::class)->save($user->toArray());
```

### Expected Output

No exception

### Anything else?

Trace:
```console
1) Tests\Integration\UserShieldTest::testSave
Error: Call to a member function saveEmailIdentity() on null

vendor/codeigniter4/shield/src/Models/UserModel.php:261
vendor/codeigniter4/framework/system/BaseModel.php:656
vendor/codeigniter4/shield/src/Models/UserModel.php:284
```

kenjis:
Fixes #471

### DIFF
--- a/src/Models/UserModel.php
+++ b/src/Models/UserModel.php
@@ -253,6 +253,11 @@ class UserModel extends Model
             /** @throws DataException */
             $result = parent::update($id, $data);
         } catch (DataException $e) {
+            // When $data is an array.
+            if ($this->tempUser === null) {
+                throw $e;
+            }
+
             $messages = [
                 lang('Database.emptyDataset', ['update']),
             ];
--- a/tests/Unit/UserModelTest.php
+++ b/tests/Unit/UserModelTest.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Tests\Unit;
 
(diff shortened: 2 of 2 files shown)


MGatner (commented):
Thanks @kenjis, this was mine and I never got around to it 🙄 I can't even remember which project it was for now...

MGatner (approved):

PR #508, opened 2022-11-03
HTML Documentation

activivan:
Proposal of generating a HTML Documentation with MkDocs to be hosted on GitHub Pages (https://squidfunk.github.io/mkdocs-material/)


datamweb:
Hi, 
I don't know why, but I still couldn't finish it.

https://shield.passkey.ir

kenjis:
https://shield.passkey.ir/ seems good.
It would be better if the PHP code is highlighted.

activivan:
And with https://shield.passkey.ir/ you'd probably want to have the Navigation in the right order. The problem with syntax highlighting is (at least for my MkDocs version), that many code sections do not include the PHP opening Tag `<?php`, and so technically it isn't PHP code but rather Text/HTML, so the compiler doesn't highlight it. You can see this here https://activivanstudios.de/cishield/install/ (first PHP codeblock has opening tag and it's being highlighted, the following ones not)

datamweb:
@activivan Thank you for this PR.
I leave the decision on this matter to other members.
It is good in my opinion.

mkdocs.yml
```diff
@@ -0,0 +1,46 @@
...
+    repo: fontawesome/brands/github
+  font:
+    text: Raleway
+  palette:
+    primary: deep orange
+  features:
+    - navigation.sections
+
+extra:
+  homepage: https://codeigniter.net
```
kenjis, line 18:
```suggestion
  homepage: https://codeigniter.com

```

MGatner:
The highlight.js convention is to supply the language type in your block opener:
```php
// This is PHP
$variable = 'string';
```
I'm not interested in changing our convention to support a new tool. I am sure there is either a different tool that already does this or a way to support highlight.JS in MKDocs. 
image

MGatner:
Looks like it is on by default:
> In addition to the default theme configuration options, the mkdocs theme supports the following options:
> * highlightjs: Enables highlighting of source code in code blocks using the highlight.js JavaScript library. Default: True.

https://www.mkdocs.org/user-guide/choosing-your-theme/

Am I just wrong about the partials?
```php
// This is PHP
$variable = 'string';

function test() {
    echo $variable;
}
```

```php
<?php
// This is PHP
$variable = 'string';

function test() {
    echo $variable;
}
```

MGatner:
Both of those seem to work on GitHub, which uses highlight.js, so I'm not sure!

activivan:
I managed to add HighlightJS, highlighting seems to work normally now

kenjis:
https://activivanstudios.de/cishield/ seems good!

docs/assets/hljs.js
```diff
@@ -0,0 +1,3 @@
+document.addEventListener('DOMContentLoaded', (event) => {
+    hljs.highlightAll();
+});
```
kenjis, line 3: Please add a newline.

kenjis (approved):
LGTM!

jozefrebjak:
👌

datamweb:
@activivan some of your commits are unsigned, please check [GPG-Signing Old Commits](https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/workflow.md#gpg-signing-old-commits).

MGatner:
It looks great! Some of the highlighting is a little odd - if the code block doesn't have a language specified it looks like it is trying to highlight like a language still. We should go through and specify all code blocks languages anyway, so I'm good with this. 

image

MGatner (approved):

MGatner:
Awaiting commit signature.

activivan:
Ok, I don't get it. May I create a new fork, add the changes and create a new pull request?

Issue #516, opened 2022-11-10
Bug: 


ATCLeHue:
### PHP Version

7.4.29

### CodeIgniter4 Version

4.2.5

### Shield Version

1.0.0-beta.3

### Which operating systems have you tested for this bug?

Windows

### Which server did you use?

cli

### Database

MariaDB  10.4.24

### Did you customize Shield?

No, 

### What happened?

When i run "php spark shield:setup"
system\Validation\FileRules.php error line 42
I think this file plus valid type request cli,  sorry if i am wrong.
![image](https://user-images.githubusercontent.com/86993639/200988597-8ea1d00f-b645-4ca2-868f-71ddab622aa3.png)

### Steps to Reproduce

I think we need more valid cli-type requests or more condition switch cases when using cli.

### Expected Output

run setup shield success

datamweb:
Hello, in the report you announced version `1.0.0-beta.3`, in the re-production process you announced `1.0.0-beta` ! 
Which one is correct?
Also, Shield requires version PHP 7.4.3+ **not 7.4.29**.

Please increase the CI version.

ATCLeHue:
> Hello, in the report you announced version `1.0.0-beta.3`, in the re-production process you announced `1.0.0-beta` ! Which one is correct? Also, Shield requires version PHP 7.4.3+ **not 7.4.29**.
> 
> Please increase the CI version.

i try both version, first i run 1.0.0-beta, it's error.
Then i go to github download 1.0.0-beta.3 and copy vendor folder, it's error too.

datamweb:
This problem does not apply to shield,
please change your CI framework version to v4.2.7+.

@Kenjis do you increase the minimum CI version in docs?

paulbalandan:
This is not an issue with shield. Upgrading CI to v4.2.6 removes the error:
```console
> php spark shield:setup -f

CodeIgniter v4.2.6 Command Line Tool - Server Time: 2022-11-09 23:47:52 UTC-06:00

  Created: APPPATH\Config\Auth.php
  Created: APPPATH\Config\AuthGroups.php
  Skipped APPPATH\Controllers\BaseController.php. It has already been updated.
  Skipped APPPATH\Config\Routes.php. It has already been updated.
  Security Setup: Everything is fine.
  Run `spark migrate --all` now? [y, n]: y
Running all new migrations...

[CodeIgniter\Database\Exceptions\DatabaseException]

Unable to connect to the database.
[...]

PR #517, opened 2022-11-10
chore: update minimum CI4 version to 4.2.7

kenjis:
See #516

It requires v4.2.6, but it has the Security Advisory: https://github.com/codeigniter4/CodeIgniter4/security/advisories/GHSA-745p-r637-7vvp


datamweb (approved):
Thanks!

PR #527, opened 2022-11-15
fix: minimum password length check

kenjis:
In the current implementation, `$minimumPasswordLength` means bytes, not characters.

### DIFF
--- a/src/Authentication/Passwords/CompositionValidator.php
+++ b/src/Authentication/Passwords/CompositionValidator.php
@@ -32,7 +32,7 @@ class CompositionValidator extends BaseValidator implements ValidatorInterface
             throw AuthenticationException::forUnsetPasswordLength();
         }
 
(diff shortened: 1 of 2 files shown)


lonnieezell (approved):

src/Authentication/Passwords/CompositionValidator.php
```diff
@@ -32,7 +32,7 @@ public function check(string $password, ?User $user = null): Result
             throw AuthenticationException::forUnsetPasswordLength();
         }
 
-        $passed = strlen($password) >= $this->config->minimumPasswordLength;
+        $passed = mb_strlen($password) >= $this->config->minimumPasswordLength;
```
datamweb, line 35:
```suggestion
        $passed = mb_strlen($password, 'UTF-8') >= $this->config->minimumPasswordLength;
```

MGatner (approved):

PR #393, opened 2022-08-19
feat: add filter permission and group

jlopes90:
Supersedes https://github.com/codeigniter4/shield/pull/270

PR #535, opened 2022-11-21
Add filter permission and group

lonnieezell:
Supersedes #393 

Provides new Groups/Permission filters.

Most of the original work was done by @jlopes90, I just tweaked the filters and added tests.


kenjis:
There is a conflict. Please rebase.
https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/workflow.md#updating-your-branch

docs/authorization.md
```diff
@@ -128,6 +130,28 @@ if (! $user->hasPermission('users.create')) {
 }
 ```
 
+#### Authorizing via Filters
+
+You can restrict access to multiple routes through a [Controller Filter](https://codeigniter.com/user_guide/incoming/filters.html). One is provided for both restricting via groups the user belongs to, as well as which permission they need. The filters are automatically registered with the system undeer the `group` and `permission` aliases, respectively. You can define the protectections within `app/Config/Filters.php`:
+
```
samsonasik, line 136: Typo: undeer -> under

samsonasik: Protetection -> protection

docs/authorization.md
```diff
@@ -128,6 +130,28 @@ if (! $user->hasPermission('users.create')) {
 }
 ```
 
+#### Authorizing via Filters
+
+You can restrict access to multiple routes through a [Controller Filter](https://codeigniter.com/user_guide/incoming/filters.html). One is provided for both restricting via groups the user belongs to, as well as which permission they need. The filters are automatically registered with the system undeer the `group` and `permission` aliases, respectively. You can define the protectections within `app/Config/Filters.php`:
```
kenjis, line 135:
```suggestion
You can restrict access to multiple routes through a [Controller Filter](https://codeigniter.com/user_guide/incoming/filters.html). One is provided for both restricting via groups the user belongs to, as well as which permission they need. The filters are automatically registered with the system under the `group` and `permission` aliases, respectively. You can define the protections within `app/Config/Filters.php`:
```

lonnieezell:
I rebased. Though it's giving me Rector fails now. If I fix the Rector issues then it becomes PHPStan issues since it doesn't know what type of variable it is, so I don't know how to please both, here. 

```
public function testFilterIncorrectGroupWithPrevious()
     {
-        /** @var User */
         $user = fake(UserModel::class);
         $user->addGroup('beta');
    ----------- end diff -----------

Applied rules:
 * RemoveNonExistingVarAnnotationRector (https://github.com/phpstan/phpstan/commit/d17e459fd9b45129c5deafe12bca56f30ea5ee99#diff-9f3541876405623b0d1[86](https://github.com/codeigniter4/shield/actions/runs/3580591390/jobs/6022847424#step:7:87)31259763dc1)
```

vs: 

```
------ ------------------------------------------------------- 
  Line   tests/Authentication/Filters/PermissionFilterTest.php  
 ------ ------------------------------------------------------- 
  :34    Cannot call method addPermission() on array|object.    
  :50    Cannot call method addPermission() on array|object.    
  :65    Cannot call method addPermission() on array|object.    
 ------ ------------------------------------------------------- 
```

tests/Authentication/Filters/GroupFilterTest.php
```diff
@@ -0,0 +1,81 @@
...
+        $result->assertRedirectTo('/login');
+
+        $result = $this->get('open-route');
+        $result->assertStatus(200);
+        $result->assertSee('Open');
+    }
+
+    public function testFilterSuccess()
+    {
+        /** @var User */
```
kenjis, line 33:
```suggestion
        /** @var User user */
```

kenjis:
Lang items are missing.

```
1) Tests\Language\TurkishTranslationTest::testAllConfiguredLanguageKeysAreIncluded with data set "tr" ('tr')
Failed asserting that the language key "Auth.notEnoughPrivilege" in the main repository is included for translation in "tr" locale.
Failed asserting that an array is empty.

/home/runner/work/shield/shield/tests/Language/AbstractTranslationTestCase.php:174

2) Tests\Language\ItalianTranslationTest::testAllConfiguredLanguageKeysAreIncluded with data set "it" ('it')
Failed asserting that the language key "Auth.notEnoughPrivilege" in the main repository is included for translation in "it" locale.
Failed asserting that an array is empty.

/home/runner/work/shield/shield/tests/Language/AbstractTranslationTestCase.php:174
```
https://github.com/codeigniter4/shield/actions/runs/3580771788/jobs/6023181104

tests/Authentication/Filters/PermissionFilterTest.php
```diff
@@ -0,0 +1,83 @@
...
+        /** @var User $user */
+        $user = fake(UserModel::class);
+        $user->addPermission('beta.access');
+
+        $result = $this
+            ->actingAs($user)
+            ->withSession(['_ci_previous_url' => site_url('open-route')])
+            ->get('protected-route');
+
+        // Should redirect to home page since previous_url is not set
```
kenjis, line 78: This is also not correct.

tests/Authentication/Filters/GroupFilterTest.php
```diff
@@ -0,0 +1,83 @@
...
+        /** @var User $user */
+        $user = fake(UserModel::class);
+        $user->addGroup('beta');
+
+        $result = $this
+            ->actingAs($user)
+            ->withSession(['_ci_previous_url' => site_url('open-route')])
+            ->get('protected-route');
+
+        // Should redirect to home page since previous_url is not set
```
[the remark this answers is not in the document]
lonnieezell: Done

MGatner (changes requested):
Since really all that is different is the actually Auth check (`inGroup()` versus `can()`) can we make an abstract filter class that handles all the rest and just processes the arguments in a separate method?
```php
abstract protected function isAuthorized(array $arguments);
```

This would make a convenient starting point for developers making their own authorization filters as well.

lonnieezell:
@MGatner Good call. Done.

MGatner (approved):
Looks good for me! Maybe @kenjis or @datamweb could take one last look?

src/Filters/AbstractAuthFilter.php
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        if ($this->isAuthorized($arguments)) {
+            return;
+        }
+
+        // If the previous_url is from this site, then
+        // we can redirect back to it.
+        if (strpos(previous_url(), site_url()) === 0) {
+            return redirect()->back()->with('error', lang('Auth.notEnoughPrivilege'));
```
datamweb, line 43: I used this PR in a practical way. I did not see a problem.

What bothers me is displaying the **error message** to the user.
**Bonfire2** uses [Tatter/Alerts](https://github.com/tattersoftware/codeigniter4-alerts) to display errors, but the following code should be used to display errors on the shield:
```php
    <?php if (session('error') !== null) : ?>
        <div class="alert alert-danger" role="alert"><?= session('error') ?></div>
    <?php elseif (session('errors') !== null) : ?>
        <div class="alert alert-danger" role="alert">
            <?php if (is_array(session('errors'))) : ?>
                <?php foreach (session('errors') as $error) : ?>
                    <?= $error ?>
                    <br>
                <?php endforeach ?>
            <?php else : ?>
                <?= session('errors') ?>
            <?php endif ?>
        </div>
    <?php endif ?>
```

In fact, I'm not sure how Shield users should know what to do to display error message (`Auth.notEnoughPrivilege`). Perhaps the need to add an explanation in the documentation or something else.

datamweb (approved):
Thanks all.

PR #548, opened 2022-11-27
fix: findByCredentials() returns User when email is empty string

### LINKED ISSUE #546: Bug: findByCredentials() returns User when email is empty string
### PHP Version

7.4

### CodeIgniter4 Version

4.2.10

### Shield Version

develop

Windows

### Which server did you use?

apache

### Database

MariaDB 10.4.18

### Did you customize Shield?

No.

### What happened?

`findByCredentials()` returns a valid user object when the email parameter value is empty.

### Steps to Reproduce

Do this:
```
$users = model(UserModel);
$user = $users->findByCredentials(['email' => '']);
dd($user);
```
This returns a valid `CodeIgniter\Shield\Entities\User` instance. 

### Expected Output

I expect `null` to be returned as no email address was entered in the value field.

### Anything else?

However, if I add a space with the value like so: `$user = $users->findByCredentials(['email' => ' ']);`, I get `null`.

kenjis:
Fixes  #546

This is a case that `empty()` causes a bug.

### DIFF
--- a/tests/Unit/UserModelTest.php
+++ b/tests/Unit/UserModelTest.php
@@ -48,7 +48,7 @@ final class UserModelTest extends TestCase
     /**
      * @see https://github.com/codeigniter4/shield/issues/546
      */
-    public function testfindByCredentialsEmptyEmail(): void
+    public function testFindByCredentialsEmptyEmail(): void
     {
         $users = $this->createUserModel();
         $user  = $this->createNewUser();
--- a/tests/Unit/UserModelTest.php
+++ b/tests/Unit/UserModelTest.php
@@ -45,6 +45,20 @@ final class UserModelTest extends TestCase
         ]);
     }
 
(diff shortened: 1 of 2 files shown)


MGatner (approved):
Thanks for the quick fix!

tests/Unit/UserModelTest.php
```diff
@@ -45,6 +45,22 @@ public function testSaveInsertUser(): void
         ]);
     }
 
+    /**
+     * @see https://github.com/codeigniter4/shield/issues/546
+     */
+    public function testfindByCredentialsEmptyEmail(): void
```
datamweb, line 51:
```suggestion
    public function testFindByCredentialsEmptyEmail(): void
```

datamweb (approved):
Thanks, the code works fine.

PR #491, opened 2022-10-31
feat: add `shield:model` command for make custom `UserModel`

datamweb:
This command helps to create custom user model.

- `php spark shield:model MyCustomUser`
- `php spark shield:model MyCustomUser --force`
- `php spark shield:model MyCustomUser --suffix`
- `php spark shield:model MyCustomUser --namespace CodeIgniter\Shield` 
## output 
```php 
<?php

declare(strict_types=1);

namespace App\Models;

use CodeIgniter\Shield\Models\UserModel;

class MyCustomUser extends UserModel
{
    protected function initialize(): void
    {
        // Merge properties with parent
        $this->allowedFields = array_merge($this->allowedFields, [
            // Add here your custom fields
            // 'first_name',

        ]);
    }
} 
```
see: #https://github.com/codeigniter4/shield/pull/489#issuecomment-1296417150

PR #558, opened 2022-12-13
feat: add `shield:model` command

paulbalandan:
Supersedes and closes #491

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -36,6 +36,8 @@ command in the terminal:
 php spark shield:model UserModel
 ```
 
+The class name is optional. If none is provided, the generated class name would be `UserModel`.
+
 You should set `Config\Auth::$userProvider` as follows:
 
 ```php
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -29,17 +29,17 @@ A default model is provided for you by the `CodeIgniter\Shield\Models\UserModel`
 this in the `Config\Auth::$userProvider` setting. The only requirement is that your new class
 MUST extend the provided `UserModel`.
 
-Shield has a CLI command to quickly create a custom `MyUserModel` class by running the following
+Shield has a CLI command to quickly create a custom `UserModel` class by running the following
 command in the terminal:
 
 ```console
-php spark shield:model MyUserModel
+php spark shield:model UserModel
 ```
 
 You should set `Config\Auth::$userProvider` as follows:
 
 ```php
-public string $userProvider = \App\Models\MyUserModel::class;
+public string $userProvider = \App\Models\UserModel::class;
 ```
 
 ## User Identities
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -24,11 +24,13 @@ on the standard Config class if nothing is found in the database.
 
 ## User Providers
 
-You can use your own models to handle user persistence. Shield calls this the "User Provider" class. A default model
-is provided for you at `CodeIgniter\Shield\Models\UserModel`. You can change this in the `Config\Auth->userProvider` setting.
-The only requirement is that your new class MUST extend the provided `UserModel`.
+You can use your own models to handle user persistence. Shield calls this the "User Provider" class.
+A default model is provided for you by the `CodeIgniter\Shield\Models\UserModel` class. You can change
+this in the `Config\Auth::$userProvider` setting. The only requirement is that your new class
+MUST extend the provided `UserModel`.
 
-Shield has a CLI command to quickly create a custom `MyUserModel` by running the following command in terminal:
+Shield has a CLI command to quickly create a custom `MyUserModel` class by running the following
+command in the terminal:
 
 ```console
 php spark shield:model MyUserModel
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -25,11 +25,19 @@ on the standard Config class if nothing is found in the database.
 ## User Providers
 
 You can use your own models to handle user persistence. Shield calls this the "User Provider" class. A default model
-is provided for you at `CodeIgniter\Shield\Models\UserModel`. You can change this in the `Config\Auth::$userProvider` setting.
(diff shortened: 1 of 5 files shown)


docs/concepts.md
```diff
@@ -25,11 +25,19 @@ on the standard Config class if nothing is found in the database.
 ## User Providers
 
 You can use your own models to handle user persistence. Shield calls this the "User Provider" class. A default model
-is provided for you at `CodeIgniter\Shield\Models\UserModel`. You can change this in the `Config\Auth::$userProvider` setting.
+is provided for you at `CodeIgniter\Shield\Models\UserModel`. You can change this in the `Config\Auth->userProvider` setting.
```
kenjis, line 28: I feel `Config\Auth::$userProvider` is better than `Config\Auth->userProvider`.

paulbalandan: Wait. I thought I changed this already during the conflict resolution?

paulbalandan: Fixed.

MGatner (commented):
Renaming it **My**UserModel means that factories won't pick it up automatically when searching for UserModel. This seems a loss to me, maybe not for the library itself but for any supporting code.

kenjis:
`UserModel` instead of `MyUserModel` seems better for the default sample.
There is no need to prefix `My`.

kenjis:
Oh, I get an error if I specify `UserModel`. It's a bug.

src/Commands/Generators/Views/usermodel.tpl.php
```diff
@@ -0,0 +1,20 @@
+<@php
+
+declare(strict_types=1);
+
+namespace {namespace};
+
+use CodeIgniter\Shield\Models\UserModel;
```
lonnieezell, line 7:
```suggestion
use CodeIgniter\Shield\Models\UserModel as ShieldUserModel;
```

src/Commands/Generators/Views/usermodel.tpl.php
```diff
@@ -0,0 +1,20 @@
...
+
+class {class} extends UserModel
```
lonnieezell, line 9:
```suggestion
class {class} extends ShieldUserModel
```

tests/Commands/UserModelGeneratorTest.php
```diff
@@ -0,0 +1,107 @@
...
+
+        stream_filter_remove($this->streamFilter);
+        $this->deleteTestFiles();
+    }
+
+    private function deleteTestFiles(): void
+    {
+        $possibleFiles = [
+            APPPATH . 'Models/MyUserModel.php',
+            HOMEPATH . 'src/Models/MyUserModel.php',
```
lonnieezell, line 41:
```suggestion
            APPPATH . 'Models/UserModel.php',
            HOMEPATH . 'src/Models/UserModel.php',
```

lonnieezell (commented):
When I tried with UserModel as the generated filename, I think the only issue was the class name conflict. I've made suggestions here to fix that in the code (though not the docs). 

I think this will fix @MGatner concerns, which I agree with. 

Other than that, my only suggestion might be to provide a default classname at the cli prompt.

MGatner (approved):
I haven't run this but the code looks good. Thanks for making that change.

lonnieezell (approved):
Looks great. Thanks for taking this one over.

src/Commands/Generators/UserModelGenerator.php
```diff
@@ -0,0 +1,78 @@
...
+        $this->component = 'Model';
+        $this->directory = 'Models';
+        $this->template  = 'usermodel.tpl.php';
+
+        $this->classNameLang = 'CLI.generator.className.model';
+        $this->setHasClassName(false);
+
+        $class = $params[0] ?? CLI::getSegment(2) ?? 'UserModel';
+
+        if (in_array(strtolower($class), ['shielduser', 'shieldusermodel'], true)) {
```
datamweb, line 68:
```suggestion
        if (in_array(strtolower($class), ['shielduser', 'shieldusermodel', 'shielduser_model', 'shield_user_model', 'shield_user', 'shield_user_'], true)) {
```
Although nothing should be chosen as the **class name**. Just for note. And probably a few more that I don't know.

datamweb (approved):

datamweb:
> It's a bug.

@kenjis waiting for you. Did your worries go away?

src/Commands/Generators/UserModelGenerator.php
kenjis: Why do we need `shielduser`? Cannot generate class `ShieldUser`?

kenjis, line 68: Why do we need `shielduser`? Cannot generate class `ShieldUser`?

kenjis: How about this?

```php
        $classToCheck = CLI::getOption('suffix') ? $class . 'Model' : $class;
        if (strtolower($classToCheck) === 'shieldusermodel') {
            CLI::error('Cannot use `ShieldUserModel` as class name as this conflicts with the parent class.', 'light_gray', 'red');

            return; // @TODO when CI4 is at v4.3+, change this to `return 1;` to signify failing exit
        }
```

paulbalandan: Updated. See https://github.com/codeigniter4/shield/pull/558/commits/7e5c4a2069cf570f7e08b934a0f7633c17f7c12e

kenjis: Okay, thanks!

kenjis (approved):
LGTM!

src/Commands/Generators/UserModelGenerator.php
kenjis: @datamweb If you are okay, please merge this.

datamweb: @paulbalandan thank you.
 Really, in many cases, your deep understanding of the code amazes me.

datamweb (approved):
Thanks all.

PR #565, opened 2022-12-19
dev: Allow easier overriding of views in developer applications.

### LINKED ISSUE #555: Dev: Extend Actions to allow custom rendering methods.
Currently, the `Email2FA` and `EmailActivation` methods use the built-in `view()` method to display a view. This works great unless someone has provided an extended View layer. For example, in Bonfire we have a theme system that uses `$this->render()` in the controllers to integrate with the theme system. This would require duplicating the actions into the developers code when all they want to do is change the render method. That leads to maintenance headaches. 

I propose we add a `renderView()` method to both of the Actions so that all a developer has to do is extend the Action class and override that one method, keeping all of the other logic intact.

lonnieezell:
Fixes  #555

When using a theme/template system other than the built in CI features, it is very difficult to integrate without copying large amounts of logic from Actions and controllers that a developer should not have to maintain. 

This adds a new trait, `Viewable`, that provides a simple, consistent way to override just how views are rendered.

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -93,7 +93,7 @@ class LoginController extends ShieldLogin
 
 ## Integrating Custom View Libraries
 
-If your application uses a different method to convert view files to HTML than CodeIgniter's built-in `view()` helper you can easily integrate your system anywhere that a view is rendered within Shield. All controllers and actions use the `CodeIgniter\Shield\Traits\Viewable` trait which provides a simple `renderView()` method that takes the same arguments as the `view()` helper. This allows you to extend the Action or Controller and only change the single method of rendering the view, leaving all of the logic untouched so your app will not need to maintain Shield logic when it doesn't need to change it.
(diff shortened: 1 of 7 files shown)


docs/customization.md
```diff
@@ -90,6 +91,25 @@ class LoginController extends ShieldLogin
...
+
+```php
+use Acme\Themes\Traits\Themeable;
+use CodeIgniter\Shield\Controllers\LoginController;
+
+class MyLoginController extends LoginController
+{
+    use Themable;
+
+    protected function view(string $view, array $data=[], array $options=[])
```
kenjis, line 106:
```suggestion
    protected function renderView(string $view, array $data = [], array $options = [])
```

MGatner (commented):
Looks good to me with @kenjis' fix.

docs/customization.md
```diff
@@ -90,6 +91,25 @@ class LoginController extends ShieldLogin
...
```
kenjis, line 106:
```suggestion
    protected function view(string $view, array $data = [], array $options = []): string
```

MGatner: Let's launch.

MGatner (approved):
@kenjis Last check?

kenjis (approved):

PR #577, opened 2023-01-01
Fix: Unnamed Auth routes returns an error when inside a route group

sammyskills:
In situations when a developer decides to put the routes in a group like so:
```php
$routes->group('accounts', static function($routes) {
    service('auth')->routes($routes);
});
```
Calls to `route_to('login')` and `route_to('register')` will return an error because there isn't any named route for that.

By explicitly setting named routes for each of the auth routes, the routing works well whether in groups or not.

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -27,7 +27,7 @@ $routes->get('register', '\App\Controllers\Auth\RegisterController::registerView
 ## Custom Redirect URLs
 
 By default, a successful login or register attempt will all redirect to `/`, while a logout action
-will redirect to a [named route](https://codeigniter4.github.io/CodeIgniter4/incoming/routing.html#using-named-routes "See routing docs") `login` or a *URI path* `/login`. You can change the default URLs used within the  `Auth` config file:
+will redirect to a [named route](https://codeigniter.com/user_guide/incoming/routing.html#using-named-routes "See routing docs") `login` or a *URI path* `/login`. You can change the default URLs used within the `Auth` config file:
 
 ```php
 public array $redirects = [
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -24,12 +24,10 @@ $routes->get('login', '\App\Controllers\Auth\LoginController::loginView');
 $routes->get('register', '\App\Controllers\Auth\RegisterController::registerView');
 ```
 
-
-
 ## Custom Redirect URLs
 
 By default, a successful login or register attempt will all redirect to `/`, while a logout action
-will redirect to `/login`. You can change the default URLs used within the `Auth` config file:
+will redirect to a [named route](https://codeigniter4.github.io/CodeIgniter4/incoming/routing.html#using-named-routes "See routing docs") `login` or a *URI path* `/login`. You can change the default URLs used within the  `Auth` config file:
 
 ```php
 public array $redirects = [
@@ -63,13 +61,10 @@ Shield has the following controllers that can be extended to handle
 various parts of the authentication process:
 
 - **ActionController** handles the after-login and after-registration actions, like Two Factor Authentication and Email Verification.
-
 - **LoginController** handles the login process.
-
 - **RegisterController** handles the registration process. Overriding this class allows you to customize the User Provider, the User Entity, and the validation rules.
-
 - **MagicLinkController** handles the "lost password" process that allows a user to login with a link sent to their email. This allows you to
-override the message that is displayed to a user to describe what is happening, if you'd like to provide more information than simply swapping out the view used.
+  override the message that is displayed to a user to describe what is happening, if you'd like to provide more information than simply swapping out the view used.
 
 It is not recommended to copy the entire controller into **app/Controllers** and change its namespace. Instead, you should create a new controller that extends
 the existing controller and then only override the methods needed. This allows the other methods to stay up to date with any security
--- a/src/Config/Auth.php
+++ b/src/Config/Auth.php
@@ -383,7 +383,7 @@ class Auth extends BaseConfig
                 break;
 
             case route_to($url) !== null: // URL is a named-route
-                $final_url = route_to($url);
+                $final_url = url_to($url);
                 break;
 
             default: // URL is a route (URI path)
--- a/src/Config/Auth.php
+++ b/src/Config/Auth.php
@@ -375,20 +375,22 @@ class Auth extends BaseConfig
     protected function getUrl(string $url): string
     {
         // To accommodate all url patterns
-        $final_url = "";
+        $final_url = '';
+
         switch ($url) {
             case strpos($url, 'http') === 0: // URL begins with 'http'. E.g. http://example.com
                 $final_url = $url;
-            break;
-            
-            case route_to($url) != null: // URL is a named-route
+                break;
+
+            case route_to($url) !== null: // URL is a named-route
                 $final_url = route_to($url);
-            break;
+                break;
 
             default: // URL is a route (URI path)
                 $final_url = rtrim(site_url($url), '/ ');
-            break;
+                break;
         }
+
         return $final_url;
     }
 }
--- a/src/Config/Auth.php
+++ b/src/Config/Auth.php
@@ -385,7 +385,7 @@ class Auth extends BaseConfig
         $final_url = '';
 
         switch (true) {
-            case strpos($url, 'http://') === 0 || strpos($url, 'https://') === 0: // URL begins with 'http'. E.g. http://example.com
+            case strpos($url, 'http://') === 0 || strpos($url, 'https://') === 0: // URL begins with 'http' or 'https'. E.g. http://example.com
                 $final_url = $url;
                 break;
 
--- a/src/Config/Auth.php
+++ b/src/Config/Auth.php
@@ -45,7 +45,7 @@ class Auth extends BaseConfig
      * auth actions. This can be either of the following:
      *
      * 1. An absolute URL. E.g. http://example.com OR https://example.com
-     * 2. A named route that can be accessed using route_to() or url_to
+     * 2. A named route that can be accessed using route_to() or url_to()
      * 3. A URI path within the application. e.g 'admin', 'login', 'expath'
      *
      * If you need more flexibility you can override the `getUrl()` method
--- a/src/Config/Auth.php
+++ b/src/Config/Auth.php
@@ -39,7 +39,7 @@ class Auth extends BaseConfig
 
     /**
      * --------------------------------------------------------------------
-     * Redirect urLs
+     * Redirect URLs
      * --------------------------------------------------------------------
      * The default URL that a user will be redirected to after
      * various auth actions. If you need more flexibility you can
--- a/src/Config/Auth.php
+++ b/src/Config/Auth.php
@@ -45,7 +45,7 @@ class Auth extends BaseConfig
      * auth actions. This can be either of the following:
      *
      * 1. An absolute URL. E.g. http://example.com OR https://example.com
-     * 2. A named route that can be accessed using route_to() or url_to()
+     * 2. A named route that can be accessed using `route_to()` or `url_to()`
      * 3. A URI path within the application. e.g 'admin', 'login', 'expath'
      *
      * If you need more flexibility you can override the `getUrl()` method
--- a/src/Config/Auth.php
+++ b/src/Config/Auth.php
@@ -382,7 +382,7 @@ class Auth extends BaseConfig
                 $final_url = $url;
                 break;
 
-            case route_to($url) !== null: // URL is a named-route
+            case route_to($url) !== false: // URL is a named-route
                 $final_url = url_to($url);
                 break;
 
--- a/src/Config/Auth.php
+++ b/src/Config/Auth.php
@@ -372,6 +372,12 @@ class Auth extends BaseConfig
         return $this->getUrl($url);
     }
 
+    /**
+     * Accepts a string which can be an absolute URL or
+     * a named route or just a URI path.
+     * @param string $url
+     * @return string
+     */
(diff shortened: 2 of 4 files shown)


lonnieezell:
I think this should be fine. @codeigniter4/core-team can you think of any reason this might be problematic?

MGatner (approved):
Looks good to me! Those should've been named routes to begin with.

datamweb (changes requested):
Hi, thanks for PR.
If you use the following code. 
```php
$routes->group('accounts', static function($routes) {
    service('auth')->routes($routes);
});
```
The `logout` path is not recognized because it is set to `login` by default.

https://github.com/codeigniter4/shield/blob/3fede093f21cca09e7997fc46cf8c81e0b9a648d/src/Config/Auth.php#L51

So I expect you to consider this or at least add some explanation in the documentation.

sammyskills:
Not sure I understand clearly what you mean. 

What you pointed out is the logout redirect, which by default, redirects to the login page. This PR fixes issues that may arise when the auth route is placed inside a group. 

If we are also considering the login/register/logout redirects, then I think using `route_to()` should fix that. 

Is this what you mean @datamweb ?

datamweb:
> what you mean

**Step1**: add 
```php
$routes->group('accounts', static function($routes) {
    service('auth')->routes($routes);
});
```
to file `app/Config/Routes.php`.

**Step2**:  login with **http://localhost:8080/accounts/login**.
**Step3**: is to logout with route **http://localhost:8080/accounts/logout**.

what will happen? 404 error(404 - File Not Found
Can't find a route for 'get: login'.). Because by default the after `logout` path is set to **http://localhost:8080/login** NOT **http://localhost:8080/accounts/login**.

Now if you change the `'logout'   => 'accounts/login'`. It will not be a problem.
https://github.com/codeigniter4/shield/blob/3fede093f21cca09e7997fc46cf8c81e0b9a648d/src/Config/Auth.php#L51

So my point is that it should be explained that the grouping affects the exit performance of the user should know how to fix this issue or use [as](https://codeigniter4.github.io/CodeIgniter4/incoming/routing.html#redirecting-routes).

kenjis:
As @datamweb says, it seems 404 error will be reported.
The setting is URL (path), not a route name.

https://github.com/codeigniter4/shield/blob/3fede093f21cca09e7997fc46cf8c81e0b9a648d/src/Config/Auth.php#L42-L52

But the issue is not related to this PR?
If a dev changes the URL of login, it is no wonder s/he needs to change the setting.

datamweb:
> The setting is URL (path), not a route name.

Yes, exactly that.

> But the issue is not related to this PR?

Well, if you think this is not related to PR, go ahead. merge.

sammyskills:
Now I understand your concern @datamweb.

Can we change the `$redirects` setting to a route name instead of just a path? I think this will fix the issue for both the grouped and ungrouped cases.

What do you think?

kenjis:
> Can we change the `$redirects` setting to a route name (at least for the logout) instead of just a path? I think this will fix the issue for both the grouped and ungrouped cases.

We cannot change it easily. Because it is a URL (path), not a route name.
If we change it, it is a breaking change.

The value will be passed to `site_url()`:
https://github.com/codeigniter4/shield/blob/3fede093f21cca09e7997fc46cf8c81e0b9a648d/src/Config/Auth.php#L375-L380

kenjis:
Shield is now in beta releases. So we could introduce breaking changes.

When we are using named routes in Shield,
it is better to use named routes in all places including the `$redirects` values.

So it seems better to change the `$redirects` values to:
1. a URL if it begins with `http`. E.g. `http://example.com/logout`
2. a named route if there is the named route. E.g. `logout`
3. a route (URI path). E.g. `logout`

We can do that if we change `getUrl()`, and existing system won't break even if it is not updated.

kenjis:
> can you think of any reason this might be problematic?

If a site has more than one login URL (e.g. `login` and `admin/login`) and
the `admin/login` has the route name `login`, this PR will break the system.
The same goes with `logout` and `register`.

But there is probably no such site, so this change will not a problem.

datamweb:
> If a site has more than one login URL (e.g. `login` and `admin/login`) and the `admin/login` has the route name `login`, this PR will break the system. 

That was interesting. I had not thought about this position.
This is an example from Joomla. **Admin** and **user** login section.

joomla

> But there is probably no such site, so this change will not a problem.

However, I agree with this.

src/Config/Auth.php
```diff
@@ -374,8 +374,23 @@ public function registerRedirect(): string
 
     protected function getUrl(string $url): string
     {
-        return strpos($url, 'http') === 0
-            ? $url
-            : rtrim(site_url($url), '/ ');
+        // To accommodate all url patterns
+        $final_url = '';
+
+        switch ($url) {
```
kenjis, line 380:
```suggestion
        switch (true) {
```

src/Config/Auth.php
```diff
@@ -374,8 +374,23 @@ public function registerRedirect(): string
...
+            case strpos($url, 'http') === 0: // URL begins with 'http'. E.g. http://example.com
```
kenjis, line 381: This condition is too loose.
If there is a named route starting with `http`, it matches this.

src/Config/Auth.php
```diff
@@ -374,8 +374,23 @@ public function registerRedirect(): string
...
+                $final_url = $url;
+                break;
+
+            case route_to($url) !== null: // URL is a named-route
```
kenjis, line 385:
```suggestion
            case route_to($url) !== false: // URL is a named-route
```

src/Config/Auth.php
```diff
@@ -374,8 +374,23 @@ public function registerRedirect(): string
 
     protected function getUrl(string $url): string
```
kenjis, line 375: The `$url` is no longer a URL. It is a URL or a route name or a route (URI path).
Please add PHPDoc `@param` and the explanation.

src/Config/Auth.php
sammyskills: If a named route starts with `http`, that means it is an absolute path, so it should be the `$final_url`.

kenjis: I don't think anyone is likely to name like this, but it is possible.:
```
+--------+-------+-----------+------------------------------+----------------+---------------+
| Method | Route | Name      | Handler                      | Before Filters | After Filters |
+--------+-------+-----------+------------------------------+----------------+---------------+
| GET    | /     | http-home | \App\Controllers\Home::index |                | toolbar       |
+--------+-------+-----------+------------------------------+----------------+---------------+
```
If it starts with `http://` or `https://` it is an absolute URL.

src/Config/Auth.php
```diff
@@ -372,10 +372,31 @@ public function registerRedirect(): string
         return $this->getUrl($url);
     }
 
+    /**
+     * Accepts a string which can be an absolute URL or
+     * a named route or just a URI path.
+     * @param string $url
```
kenjis, line 378:
```suggestion
     * @param string $url an absolute URL or a named route or just a URI path.
```

src/Config/Auth.php
sammyskills: Ah yes, you are correct.

src/Config/Auth.php
```diff
@@ -372,10 +372,31 @@ public function registerRedirect(): string
...
+        // To accommodate all url patterns
+        $final_url = '';
+
+        switch (true) {
+            case strpos($url, 'http') === 0: // URL begins with 'http'. E.g. http://example.com
+                $final_url = $url;
+                break;
+
+            case route_to($url) !== false: // URL is a named-route
+                $final_url = url_to($url);
```
kenjis, line 392: Since we are trimming in the case of URI path, existing tests will fail if we do not do so here as well.

```suggestion
                $final_url = rtrim(url_to($url), '/ ');
```

src/Config/Auth.php
```diff
@@ -372,10 +372,32 @@ public function registerRedirect(): string
...
     protected function getUrl(string $url): string
     {
-        return strpos($url, 'http') === 0
-            ? $url
-            : rtrim(site_url($url), '/ ');
+        // To accommodate all url patterns
+        $final_url = '';
+
+        switch (true) {
+            case strpos($url, 'http://') === 0 || strpos($url, 'https://'): // URL begins with 'http'. E.g. http://example.com
```
kenjis, line 388:
```suggestion
            case strpos($url, 'http://') === 0 || strpos($url, 'https://') === 0: // URL begins with 'http'. E.g. http://example.com
```

sammyskills: How did I miss that? 

Thanks @kenjis

src/Config/Auth.php
```diff
@@ -372,10 +372,32 @@ public function registerRedirect(): string
...
     protected function getUrl(string $url): string
     {
-        return strpos($url, 'http') === 0
-            ? $url
-            : rtrim(site_url($url), '/ ');
+        // To accommodate all url patterns
+        $final_url = '';
+
+        switch (true) {
+            case strpos($url, 'http://') === 0 || strpos($url, 'https://') === 0: // URL begins with 'http'. E.g. http://example.com
```
datamweb, line 388:
```suggestion
            case strpos($url, 'http://') === 0 || strpos($url, 'https://') === 0: // URL begins with 'http' or 'https'. E.g. http://example.com
```

datamweb:
Another issue is that we need to be careful when the user uses grouping. It should change as follows.
```php
public $globals = [
    'before' => [
        // ...
        'session' => ['except' => ['login*', 'register', 'auth/a/*']],
    ],
    // ...
];
```
```php
public $globals = [
    'before' => [
        // ...
        'session' => ['except' => ['accounts/login*', 'accounts/register', 'accounts/auth/a/*']],
    ],
    // ...
];
```

sammyskills:
Hi @datamweb , 

I think that should go to the docs. I'll push a commit to that effect.

datamweb (approved):
Thank you for being with us.

We all strive to make Shield better.

datamweb (approved):
LGMT!

kenjis:
Sorry, I meant please update the explanation. Because it will be not only URL, but also named route.
https://github.com/codeigniter4/shield/blob/83dd09a66088af51ac6d365002fa77f0bb1eeeb4/src/Config/Auth.php#L41-L46

sammyskills:
Is this really necessary? 

I'm asking because no change was done to the `$redirects`.

kenjis:
> while a logout action will redirect to /login.
https://github.com/codeigniter4/shield/blob/develop/docs/customization.md#custom-redirect-urls

This description also needs to be updated.

sammyskills:
Oh, I get the point now, thank you.

kenjis:
> Is this really necessary?

In the default state, the route `login`'s path is `login` . There is no difference between the route name and path.
However, the meaning of the value `login` is the route name.

So it would be better to have an explanation that if the route name is set, it is the route name.

src/Config/Auth.php
```diff
@@ -39,11 +39,17 @@ class Auth extends BaseConfig
...
+     * Redirect URLs
      * --------------------------------------------------------------------
-     * The default URL that a user will be redirected to after
-     * various auth actions. If you need more flexibility you can
-     * override the `getUrl()` method to apply any logic you may need.
+     * The default URL that a user will be redirected to after various auth
+     * auth actions. This can be either of the following:
+     *
+     * 1. An absolute URL. E.g. http://example.com OR https://example.com
+     * 2. A named route that can be accessed using route_to() or url_to
```
kenjis, line 48:
```suggestion
     * 2. A named route that can be accessed using route_to() or url_to()
```

docs/customization.md
```diff
@@ -24,12 +24,10 @@ $routes->get('login', '\App\Controllers\Auth\LoginController::loginView');
 $routes->get('register', '\App\Controllers\Auth\RegisterController::registerView');
 ```
 
-
-
 ## Custom Redirect URLs
 
 By default, a successful login or register attempt will all redirect to `/`, while a logout action
-will redirect to `/login`. You can change the default URLs used within the `Auth` config file:
+will redirect to a [named route](https://codeigniter4.github.io/CodeIgniter4/incoming/routing.html#using-named-routes "See routing docs") `login` or a *URI path* `/login`. You can change the default URLs used within the  `Auth` config file:
```
kenjis, line 30: There is an extra space. And it seems we are using the user guide in codeigniter.com.
```suggestion
will redirect to a [named route](https://codeigniter.com/user_guide/incoming/routing.html#using-named-routes "See routing docs") `login` or a *URI path* `/login`. You can change the default URLs used within the `Auth` config file:
```

kenjis (approved):
LGTM!

src/Config/Auth.php
```diff
@@ -39,11 +39,17 @@ class Auth extends BaseConfig
...
+     * Redirect URLs
      * --------------------------------------------------------------------
-     * The default URL that a user will be redirected to after
-     * various auth actions. If you need more flexibility you can
-     * override the `getUrl()` method to apply any logic you may need.
+     * The default URL that a user will be redirected to after various auth
+     * auth actions. This can be either of the following:
+     *
+     * 1. An absolute URL. E.g. http://example.com OR https://example.com
+     * 2. A named route that can be accessed using route_to() or url_to()
```
datamweb, line 48:
```suggestion
     * 2. A named route that can be accessed using `route_to()` or `url_to()`
```

kenjis (approved):

datamweb (approved):
Awaiting approval from **lonnieezell**.

kenjis:
@lonnieezell Can you review?
If you don't have time, we will merge this.

lonnieezell (approved):

lonnieezell:
Approved

PR #578, opened 2023-01-01
Fix: Remove unnecessary register action checks

sammyskills:
The `registerView()` method of the RegisterController already has the loggedIn and allowRegistration checks, so I don't think (IMO), it is still necessary to have those checks in the `registerAction()` method. Remember that _"under normal circumstance"_, a user cannot arrive at the `POST` request of the registration if it doesn't generate from the `GET` (view), so having the checks in both methods is just overkill.

The `LoginController` has a better sample.

### DIFF
--- a/src/Controllers/RegisterController.php
+++ b/src/Controllers/RegisterController.php
@@ -58,6 +58,7 @@ class RegisterController extends BaseController
      */
     public function registerAction(): RedirectResponse
     {
+        // Get the userProvider
         $users = $this->getUserProvider();
 
         // Validate here first, since some things,
--- a/src/Controllers/RegisterController.php
+++ b/src/Controllers/RegisterController.php
@@ -58,10 +58,6 @@ class RegisterController extends BaseController
      */
     public function registerAction(): RedirectResponse
     {
-        if (auth()->loggedIn()) {
-            return redirect()->to(config('Auth')->registerRedirect());
-        }
-
         // Check if registration is allowed
         if (! setting('Auth.allowRegistration')) {
             return redirect()->back()->withInput()
--- a/src/Controllers/RegisterController.php
+++ b/src/Controllers/RegisterController.php
@@ -58,12 +58,6 @@ class RegisterController extends BaseController
      */
     public function registerAction(): RedirectResponse
     {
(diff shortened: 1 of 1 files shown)


lonnieezell:
The problem with removing these check is that anyone can navigate to the URL manually and be a bad actor. Keeping the checks in place hurts nothing and helps to protect this endpoint and, hopefully, avoid potential security issue.

sammyskills:
>

sammyskills:
Yes, I understand and agree with you that it hurts nothing to keep those checks, but forgive me if I'm asking too many questions:

* Can an innocent user navigate straight to the POST endpoint without first hitting the GET endpoint?
* If the answer to the above question is yes, what then is the usefulness of the CSRF protection that is added to the form?

kenjis:
A innocent user does not matter. We need to protect our site from attackers.
A normal user does not navigate straight to the POST endpoint without first hitting the GET endpoint.

HTTP is a stateless protocol, so attackers can send requests to any endpoint directly.
So we need to have the checks in all endpoint.
If we assume normal page flaw, it may cause vulnerabilities.

sammyskills:
Going by your explanations, that means the checks in the `LoginController::loginView()` method should also be applied to the `LoginController::loginAction()` method, correct?

See https://github.com/codeigniter4/shield/blob/3fede093f21cca09e7997fc46cf8c81e0b9a648d/src/Controllers/LoginController.php#L32-L35

kenjis:
Yes. It is applied already.

https://github.com/codeigniter4/shield/blob/3fede093f21cca09e7997fc46cf8c81e0b9a648d/src/Controllers/LoginController.php#L67-L70

sammyskills:
Sorry @kenjis, I copied the wrong permalink. This is what I wanted to refer to:
https://github.com/codeigniter4/shield/blob/3fede093f21cca09e7997fc46cf8c81e0b9a648d/src/Controllers/LoginController.php#L25-L27

kenjis:
Yes, it should also be applied.

If you send a request with correct credentials to the `loginAction()` directly after logged in,
probably LogicException will occur.

PR #580, opened 2023-01-04
feat: User activation checks and utility functions.

### LINKED ISSUE #459: Bug: when active field in users table is `false` users can login
### PHP Version

8.1.5

### CodeIgniter4 Version

4.2.5

### Shield Version

dev

Windows

### Which server did you use?

apache

### Database

MySQL 5.6

### Did you customize Shield?

No.

### What happened?

The login operation for a user who is **not active** is completed successfully. In general, a user who is not active for **any reason should not login** to the system.

### Steps to Reproduce

1. Register a new user.
2. Change the value of active `true` to `false`.
3. Perform the login operation.
4. **A user who is inactive login without a problem!!**

### Expected Output

Failure of the user login to the system and display the error message "**Your account is inactive**".

### Anything else?

see https://forum.codeigniter.com/showthread.php?tid=83481

lonnieezell:
Fixes #459

Previously, the `active` column for the user was not being used except that the `EmailActivator` flow would set it. This PR addresses this by: 

- Adding a new `Activatable` trait that is used on the `User` entity.  This provides methods to check if a user is activated or not, and utility methods to activate/deactivate that user.
- The `SessionAuth` and `TokenAuth` filters are updated to check the user's active status and redirect to login/return a 403 status code, respectively, when they aren't active. This takes into account if no activator is required after registration.

### DIFF
--- a/docs/authorization.md
+++ b/docs/authorization.md
@@ -252,7 +252,7 @@ if ($user->isActivated()) {
 }
 ```
 
-Note: If no activator is specified in the `Auth` config file, `actions['register']` property, then this will always return `true`.
+> **Note** If no activator is specified in the `Auth` config file, `actions['register']` property, then this will always return `true`.
 
 You can check if a user has not been activated yet via the `isNotActivated()` method.
 
--- a/docs/authorization.md
+++ b/docs/authorization.md
@@ -20,6 +20,10 @@
       - [removeGroup()](#removegroup)
       - [syncGroups()](#syncgroups)
       - [getGroups()](#getgroups)
+  - [User Activation](#user-activation)
+    - [Checking Activation Status](#checking-activation-status)
+  - [Activating a User](#activating-a-user)
+  - [Deactivating a User](#deactivating-a-user)
 
 Authorization happens once a user has been identified through authentication. It is the process of
 determining what actions a user is allowed to do within your site.
@@ -232,4 +236,44 @@ Returns all groups this user is a part of.
 
 ```php
 $user->getGroups();
+```
+
+## User Activation
+
+All users have an `active` flag. This is only used when the [`EmailActivation` action](./auth_actions.md), or a custom action used to activate a user, is enabled.
+
+### Checking Activation Status
+
+You can determine if a user has been activated with the `isActivated()` method.
+
+```php
+if ($user->isActivated()) {
+    //
+}
+```
+
+Note: If no activator is specified in the `Auth` config file, `actions['register']` property, then this will always return `true`.
+
+You can check if a user has not been activated yet via the `isNotActivated()` method.
+
+```php
+if ($user->isNotActivated()) {
+    //
+}
+```
+
+## Activating a User
+
+Users are automatically activated withih the `EmailActivator` action. They can be manually activated via the `activate()` method on the User entity.
+
+```php
+$user->activate();
+```
+
+## Deactivating a User
+
+Users can be manually deactivated via the `deactivate()` method on the User entity.
+
+```php
+$user->deactivate();
 ```
--- a/src/Filters/SessionAuth.php
+++ b/src/Filters/SessionAuth.php
@@ -49,6 +49,15 @@ class SessionAuth implements FilterInterface
                 $authenticator->recordActiveDate();
             }
 
+            // Block inactive users when Email Activation is enabled
+            $user = $authenticator->getUser();
+            if ($user !== null && ! $user->isActivated()) {
+                $authenticator->logout();
+
+                return redirect()->route('login')
+                    ->with('error', lang('Auth.activationBlocked'));
+            }
+
             return;
         }
 
--- a/src/Filters/TokenAuth.php
+++ b/src/Filters/TokenAuth.php
@@ -50,7 +50,7 @@ class TokenAuth implements FilterInterface
 
         if (! $result->isOK() || (! empty($arguments) && $result->extraInfo()->tokenCant($arguments[0]))) {
             return service('response')
-                ->setStatusCode(Response::HTTP_FORBIDDEN)
+                ->setStatusCode(Response::HTTP_UNAUTHORIZED)
                 ->setJson(['message' => lang('Auth.badToken')]);
         }
 
--- a/src/Filters/TokenAuth.php
+++ b/src/Filters/TokenAuth.php
@@ -49,11 +49,23 @@ class TokenAuth implements FilterInterface
         ]);
 
         if (! $result->isOK() || (! empty($arguments) && $result->extraInfo()->tokenCant($arguments[0]))) {
-            return redirect()->to('/login');
+            return service('response')
+                ->setStatusCode(Response::HTTP_FORBIDDEN)
+                ->setJson(['message' => lang('Auth.badToken')]);
         }
 
         if (setting('Auth.recordActiveDate')) {
             $authenticator->recordActiveDate();
+        }
+
+        // Block inactive users when Email Activation is enabled
+        $user = $authenticator->getUser();
+        if ($user !== null && ! $user->isActivated()) {
+            $authenticator->logout();
+
+            return service('response')
+                ->setStatusCode(Response::HTTP_FORBIDDEN)
+                ->setJson(['message' => lang('Auth.activationBlocked')]);
         }
     }
 
--- a/src/Language/en/Auth.php
+++ b/src/Language/en/Auth.php
@@ -48,7 +48,7 @@ return [
     'successLogout'      => 'You have successfully logged out.',
 
     // Passwords
-    'errorPasswordLength'       => 'Passwords must be at least {0, number} characters long.',
+    'errorPasswordLength'       => 'Passwords must be at least {0activationBlocked, number} characters long.',
     'suggestPasswordLength'     => 'Pass phrases - up to 255 characters long - make more secure passwords that are easy to remember.',
     'errorPasswordCommon'       => 'Password must not be a common password.',
     'suggestPasswordCommon'     => 'The password was checked against over 65k commonly used passwords or passwords that have been leaked through hacks.',
@@ -87,6 +87,7 @@ return [
     'emailActivateMailBody' => 'Please use the code below to activate your account and start using the site.',
     'invalidActivateToken'  => 'The code was incorrect.',
     'needActivate'          => 'You must complete your registration by confirming the code sent to your email address.',
+    'activationBlocked'     => 'You must activate your account before logging in.',
 
     // Groups
     'unknownGroup' => '{0} is not a valid group.',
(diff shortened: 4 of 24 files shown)


docs/authorization.md
```diff
@@ -228,3 +232,43 @@ Returns all groups this user is a part of.
...
+
+You can determine if a user has been activated with the `isActivated()` method.
+
+```php
+if ($user->isActivated()) {
+    //
+}
+```
+
+Note: If no activator is specified in the `Auth` config file, `actions['register']` property, then this will always return `true`.
```
datamweb, line 250:
```suggestion
> **Note** If no activator is specified in the `Auth` config file, `actions['register']` property, then this will always return `true`.
```

src/Filters/SessionAuth.php
```diff
@@ -49,6 +49,15 @@ public function before(RequestInterface $request, $arguments = null)
                 $authenticator->recordActiveDate();
             }
 
+            // Block inactive users when Email Activation is enabled
+            $user = $authenticator->getUser();
+            if ($user !== null && ! $user->isActivated()) {
+                $authenticator->logout();
+
+                return redirect()->route('login')
+                    ->with('error', lang('Auth.activationBlocked'));
```
datamweb, line 58: This message is suitable if the user has not activated it himself.
error

Suppose the user is activated. For some reason, the administrator has disabled it.

Is it appropriate to display the following message now?

src/Filters/TokenAuth.php
```diff
@@ -49,12 +49,24 @@ public function before(RequestInterface $request, $arguments = null)
...
         }
+
+        // Block inactive users when Email Activation is enabled
+        $user = $authenticator->getUser();
+        if ($user !== null && ! $user->isActivated()) {
+            $authenticator->logout();
+
+            return service('response')
+                ->setStatusCode(Response::HTTP_FORBIDDEN)
+                ->setJson(['message' => lang('Auth.activationBlocked')]);
```
datamweb, line 68: Same topic as above.

src/Language/en/Auth.php
```diff
@@ -50,7 +50,7 @@
     'successLogout'      => 'You have successfully logged out.',
 
     // Passwords
-    'errorPasswordLength'       => 'Passwords must be at least {0, number} characters long.',
+    'errorPasswordLength'       => 'Passwords must be at least {0activationBlocked, number} characters long.',
```
datamweb, line 53:
```suggestion
    'errorPasswordLength'       => 'Passwords must be at least {0, number} characters long.',
```
Looks like a typo.

src/Language/fa/Auth.php
```diff
@@ -98,6 +98,7 @@
     'emailActivateMailBody' => 'لطفا برای فعالسازی حساب کاربری و استفاده از سایت از کد زیر استفاده کنید.',
     'invalidActivateToken'  => 'کد صحیح نمی باشد.',
     'needActivate'          => 'شما باید با ارائه کد ارسال شده به ایمیلتان، ثبت نام را تکمیل کنید.',
+    'activationBlocked'     => '(to be translated) You must activate your account before logging in.',
```
datamweb, line 101:
```suggestion
    'activationBlocked'     => 'قبل از تلاش برای ورود، باید اکانت خود را فعال کنید.',
```

src/Traits/Activatable.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        return ! $this->isActivated();
+    }
+
+    /**
+     * Activates the user.
+     */
+    public function activate(): bool
+    {
+        helper('auth');
```
datamweb, line 32:
```suggestion
```
Loaded by default, no need to load.

src/Traits/Activatable.php
```diff
@@ -0,0 +1,58 @@
...
+
+        return $model->update($this->id, ['active' => 1]);
+    }
+
+    /**
+     * Deactivates the user.
+     */
+    public function deactivate(): bool
+    {
+        helper('auth');
```
datamweb, line 43:
```suggestion
```
No need.

src/Traits/Activatable.php
```diff
@@ -0,0 +1,58 @@
...
+     */
+    public function isNotActivated(): bool
+    {
+        return ! $this->isActivated();
+    }
+
+    /**
+     * Activates the user.
+     */
+    public function activate(): bool
```
datamweb, line 30: I suggest that it has a parameter, for example public function `activate(...$ids)`.
By default, consider the **current user**, otherwise, if the input `$ids` is set, activate the target users. It is good for the management department and prevents writing a new query.

I have the same opinion for `deactivate()`.

kenjis: `Activatable` is used in `User`. So `activate(...$ids)` is not appropriate.

```php
$user->activate(1, 2, 3)
```
It means a user can activate other users. It is bad design.

If we need the api `activate(...$ids)`, it should be in other class.

lonnieezell: I agree with @kenjis  that this is not appropriate for this use case.

src/Filters/SessionAuth.php
lonnieezell: We do not have any way to know whether the admin did it or the user has never activated currently. I think that's likely ok, though. For applications where an admin might remove it, they could easily change the language string to reflect a more generic situation. 

Banning might be more appropriate in these situations, also?

src/Traits/Activatable.php
lonnieezell: I forgot it was loaded via Composer now. Was thinking it was still in the filters. Thanks for reminder.

src/Filters/SessionAuth.php
kenjis: I agree with @lonnieezell 

If an admin deactivates a user, the error message seems to be still appropriate.
Only thing the user to do is to activate again.
If the message is not appropriate or ambiguous, devs would change the message.

Activation is not banning. I don't see use cases that an admin deactivates users.

src/Language/ja/Auth.php
```diff
@@ -89,6 +89,7 @@
     'emailActivateMailBody' => '以下のコードを使用してアカウントを有効化し、サイトの利用を開始してください。', // 'Please use the code below to activate your account and start using the site.',
     'invalidActivateToken'  => 'コードが間違っています。', // 'The code was incorrect.',
     'needActivate'          => 'メールアドレスに送信されたコードを確認し、登録を完了する必要があります。', // 'You must complete your registration by confirming the code sent to your email address.',
+    'activationBlocked'     => '(to be translated) You must activate your account before logging in.',
```
kenjis, line 92:
```suggestion
    'activationBlocked'     => 'ログインする前にアカウントを有効化する必要があります。',
```

src/Filters/TokenAuth.php
```diff
@@ -49,12 +49,24 @@ public function before(RequestInterface $request, $arguments = null)
         ]);
 
         if (! $result->isOK() || (! empty($arguments) && $result->extraInfo()->tokenCant($arguments[0]))) {
-            return redirect()->to('/login');
+            return service('response')
+                ->setStatusCode(Response::HTTP_FORBIDDEN)
+                ->setJson(['message' => lang('Auth.badToken')]);
```
kenjis, line 54: This is a breaking change. It is better to document it.

#433 suggests 401, not 403. Which is better?

jozefrebjak: +1 for 401. For example, I already using it in few deployments with own api-filter. I would like to switch to this one if it's will be suitable for me.

jozefrebjak: > 401 Unauthorized is the status code to return when the client provides no credentials or invalid credentials. 403 Forbidden is the status code to return when a client has valid credentials but not enough privileges to perform an action on a resource. 

> Receiving a 403 response is the server telling you, “I’m sorry. I know who you are–I believe who you say you are–but you just don’t have permission to access this resource. Maybe if you ask the system administrator nicely, you’ll get permission. But please don’t bother me again until your predicament changes.”

In summary, a 401 Unauthorized response should be used for missing or bad authentication, and a 403 Forbidden response should be used afterwards, when the user is authenticated but isn’t authorized to perform the requested operation on the given resource.

@lonnieezell  Here we are mixing authentication and authorization. Maybe we need to firstly check if client provided valid token and if not return 401 and after that check scopes and if there is no valid scope return 403.

src/Traits/Activatable.php
datamweb: Ok, I realized.

src/Filters/TokenAuth.php
kenjis: > The HyperText Transfer Protocol (HTTP) 401 Unauthorized response status code indicates that the client request has not been completed because it lacks valid authentication credentials for the requested resource. 

> This status code is similar to the [403 Forbidden](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) status code, except that in situations resulting in this status code, user authentication can allow access to the resource.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401

erikkraijenoord:
Is this still going to be released or is it BC?
I'm struggling with the fact it shows the default login rather than the response.

src/Filters/TokenAuth.php
lonnieezell: These make sense. First check will result in a 401, with the not-activated check being a 403.

src/Traits/Activatable.php
```diff
@@ -0,0 +1,56 @@
...
```
kenjis, line 30: Can you change the return type to `void`?
The `UserModel::update()` throws an exception when the update fails.
So returning `bool` does not make sense.

src/Traits/Activatable.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        $model = auth()->getProvider();
+
+        return $model->update($this->id, ['active' => 1]);
+    }
+
+    /**
+     * Deactivates the user.
+     */
+    public function deactivate(): bool
```
kenjis, line 40: The same above.

datamweb:
We have a new language file `pt-BR`, please add:
```php
    'activationBlocked'     => '(to be translated) You must activate your account before logging in.',
```

Also conflict in file **src/Entities/User.php**.

src/Traits/Activatable.php
lonnieezell: Unless it's changing in a PR, `update()` currently [returns a boolean](https://github.com/codeigniter4/shield/blob/develop/src/Models/UserModel.php#L268). Changing this to `void` throws errors.

datamweb (approved):
@kenjis, we can address your opinion separately in PR. This PR has taken too long.

kenjis (approved):

src/Traits/Activatable.php
kenjis: I sent a PR #626

lonnieezell:
Thanks!

PR #595, opened 2023-01-11
Update UserModelGenerator for CI 4.3.0

kenjis:
See #596

- update UserModelGenerator::run()
- update tests

### DIFF
--- a/src/Commands/Generators/UserModelGenerator.php
+++ b/src/Commands/Generators/UserModelGenerator.php
@@ -54,7 +54,7 @@ class UserModelGenerator extends BaseCommand
     /**
      * Actually execute the command.
      */
-    public function run(array $params): void
+    public function run(array $params): int
     {
         $this->component = 'Model';
         $this->directory = 'Models';
@@ -68,12 +68,14 @@ class UserModelGenerator extends BaseCommand
         if (! $this->verifyChosenModelClassName($class, $params)) {
             CLI::error('Cannot use `ShieldUserModel` as class name as this conflicts with the parent class.', 'light_gray', 'red');
 
-            return; // @TODO when CI4 is at v4.3+, change this to `return 1;` to signify failing exit
+            return 1;
         }
 
         $params[0] = $class;
 
-        $this->execute($params);
+        $this->generateClass($params);
+
+        return 0;
     }
 
     /**
--- a/src/Commands/Generators/UserModelGenerator.php
+++ b/src/Commands/Generators/UserModelGenerator.php
@@ -73,7 +73,8 @@ class UserModelGenerator extends BaseCommand
 
         $params[0] = $class;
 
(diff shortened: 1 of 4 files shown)


kenjis:
Some checks failed, but they will be fixed in the next PR #597

src/Commands/Generators/UserModelGenerator.php
```diff
@@ -68,12 +68,14 @@ public function run(array $params): void
...
             CLI::error('Cannot use `ShieldUserModel` as class name as this conflicts with the parent class.', 'light_gray', 'red');
 
-            return; // @TODO when CI4 is at v4.3+, change this to `return 1;` to signify failing exit
+            return 1;
         }
 
         $params[0] = $class;
 
-        $this->execute($params);
+        $this->generateClass($params);
```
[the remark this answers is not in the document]
kenjis: Good point! I will fix it.

tests/Commands/UserModelGeneratorTest.php
```diff
@@ -18,10 +19,17 @@ protected function setUp(): void
     {
         parent::setUp();
 
-        CITestStreamFilter::$buffer = '';
+        // @phpstan-ignore-next-line
+        if (version_compare(CodeIgniter::CI_VERSION, '4.3.0', '>=')) {
```
paulbalandan, line 23: Why do you need this if statement? Why not go with 4.3 and change related lines?

src/Commands/Generators/UserModelGenerator.php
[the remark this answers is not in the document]
kenjis: Fixed.

tests/Commands/UserModelGeneratorTest.php
kenjis: 4.3.0 has many breaking changes. I think dropping 4.2 now is too early.

paulbalandan: Okay. I think that's the reason for your PR testing against lowest dependencies, right?

kenjis: Yes!

paulbalandan (approved):

tests/Commands/UserModelGeneratorTest.php
```diff
@@ -18,10 +19,17 @@ protected function setUp(): void
     {
         parent::setUp();
 
-        CITestStreamFilter::$buffer = '';
+        // @phpstan-ignore-next-line
```
paulbalandan, line 22: This can be added to phpstan's dynamic constants.

kenjis: Thanks. It is better! Done.

datamweb (approved):
Thanks!

PR #599, opened 2023-01-12
chore: update phpunit.yml

kenjis:
- add testing with lowest dependency
- update minimun version of settings and vfsstream
- reduce the number of checks (Postgre, SQLSRV, OCI8)
- add PHP 8.2 checks

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -28,19 +28,36 @@ jobs:
     if: "!contains(github.event.head_commit.message, '[ci skip]')"
     strategy:
       matrix:
-        php-versions: ['7.4', '8.0', '8.1']
-        db-platforms: ['MySQLi', 'Postgre', 'SQLite3', 'SQLSRV', 'OCI8']
-        mysql-versions: [ '5.7' ]
+        php-versions: ['7.4', '8.0', '8.1', '8.2']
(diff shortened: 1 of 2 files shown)


.github/workflows/phpunit.yml
```diff
@@ -121,12 +147,16 @@ jobs:
...
+
       - name: Install dependencies
         run: |
           if [ -f composer.lock ]; then
-            composer install --no-progress --no-interaction --prefer-dist --optimize-autoloader
+            composer install ${{ env.COMPOSER_UPDATE_FLAGS }} --no-progress --no-interaction --prefer-dist --optimize-autoloader
           else
-            composer update --no-progress --no-interaction --prefer-dist --optimize-autoloader
+            composer update ${{ env.COMPOSER_UPDATE_FLAGS }} --no-progress --no-interaction --prefer-dist --optimize-autoloader
           fi
```
paulbalandan, line 160: Instead of additional step and global env, you can make it a step env.
```suggestion
          fi
      env:
          COMPOSER_UPDATE_FLAGS: ${{ matrix.dependencies == 'lowest' && '--prefer-lowest' || '' }}
```

kenjis: Done.

PR #619, opened 2023-02-02
refactor: remove `namespace` from language files

datamweb:
CI4 is built for PHP 7.4+, and everything in the framework is namespaced,
  except for the helper and lang files.

see https://github.com/codeigniter4/CodeIgniter4/pull/7210

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -97,6 +97,7 @@ return static function (RectorConfig $rectorConfig): void {
         // Ignore files that should not be namespaced to their folder
         NormalizeNamespaceByPSR4ComposerAutoloadRector::class => [
             __DIR__ . '/src/Helpers',
+            __DIR__ . '/src/Language',
             __DIR__ . '/tests/_support',
         ],
 
--- a/src/Language/de/Auth.php
+++ b/src/Language/de/Auth.php
@@ -2,8 +2,6 @@
 
 declare(strict_types=1);
 
-namespace CodeIgniter\Shield\Language\de;
-
 return [
     // Exceptions
     'unknownAuthenticator'  => '{0} ist kein gültiger Authentifikator.',
--- a/src/Language/en/Auth.php
+++ b/src/Language/en/Auth.php
@@ -2,8 +2,6 @@
 
 declare(strict_types=1);
 
-namespace CodeIgniter\Shield\Language\en;
-
 return [
     // Exceptions
     'unknownAuthenticator'  => '{0} is not a valid authenticator.',
--- a/src/Language/es/Auth.php
+++ b/src/Language/es/Auth.php
@@ -2,8 +2,6 @@
 
 declare(strict_types=1);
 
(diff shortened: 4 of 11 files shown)


jozefrebjak:
The Rector rule `NormalizeNamespaceByPSR4ComposerAutoloadRector` is trying to add a namespace to the not-needed classes. I had the same problem in my projects, where rector  tried to add a namespace also to the `app/Config` files. I ended with the removal of that rule.

samsonasik (approved):

kenjis (approved):

datamweb:
@jozefrebjak  As you know, CI4 heavily uses the namespace. On the other hand, completely eliminating class `NormalizeNamespaceByPSR4ComposerAutoloadRector` does not seem like a good practice.

I don't understand why you want to not use namespace in **app/Config**. Anyway you can do as I did.
You can see a more complete explanation in the [ignoring-rules-or-paths](https://getrector.com/documentation/ignoring-rules-or-paths).

datamweb:
@samsonasik Thank you for your quick review.

**Kenjis**, I prefer to merge after https://github.com/codeigniter4/CodeIgniter4/pull/7210.

jozefrebjak:
> @jozefrebjak As you know, CI4 heavily uses the namespace. On the other hand, completely eliminating class `NormalizeNamespaceByPSR4ComposerAutoloadRector` does not seem like a good practice.
> 
> I don't understand why you want to not use namespace in **app/Config**. Anyway you can do as I did. You can see a more complete explanation in the [ignoring-rules-or-paths](https://getrector.com/documentation/ignoring-rules-or-paths).

By default is for example `app/Config/App.php` namespaced like:

```php
namespace Config;
```

after enabling that rule as it's provided in devkit is rector trying to change it to:

```php
namespace  App/Config;
```

and it ends with error:

```sh
Fatal error: Uncaught TypeError: CodeIgniter\CodeIgniter::__construct(): Argument #1 ($config) must be of type Config\App, App\Config\App given, called in /app/vendor/codeigniter4/framework/system/Config/Services.php on line 150 and defined in /app/vendor/codeigniter4/framework/system/CodeIgniter.php:170 ```

I know how to namespace and I was lazy so I decided to remove that rule till I find a solutions.

**Comment by jozefrebjak** [comment]
@datamweb Oh, I checked the latest project which I am working on and I already did it in the past :D So sorry about my comment here. 

```php
        // Ignore files that should not be namespaced to their folder
        NormalizeNamespaceByPSR4ComposerAutoloadRector::class => [
            __DIR__ . '/app/Common.php',
            __DIR__ . '/app/Config',
            __DIR__ . '/app/Language',
            __DIR__ . '/app/Helpers',
            __DIR__ . '/tests/_support',
        ],
```

datamweb:
> So sorry about my comment here.

No problem, I'm glad you're here.

PR #617, opened 2023-02-01
Feature: Banning Users

davidnsai:
Closes #509

A completed feature for banning and unbanning users.

PR #628, opened 2023-02-06
feat:  customize name of Shield Tables

datamweb:


datamweb:
What has happened PHPStan! There is no error in local!

Screenshot 2023-02-06 212606

And PHPUNIT test !

Screenshot 2023-02-06 213030

src/Config/Auth.php
```diff
@@ -37,6 +37,36 @@ class Auth extends BaseConfig
...
+     */
+    public array $authTables = [
+        'users'                  => 'users',
+        'auth_identities'        => 'auth_identities',
+        'auth_logins'            => 'auth_logins',
+        'auth_token_logins'      => 'auth_token_logins',
+        'auth_remember_tokens'   => 'auth_remember_tokens',
+        'auth_groups_users'      => 'auth_groups_users',
+        'auth_permissions_users' => 'auth_permissions_users',
+    ];
```
kenjis, line 68:
```suggestion
    public array $tables = [
        'users'             => 'users',
        'identities'        => 'auth_identities',
        'logins'            => 'auth_logins',
        'token_logins'      => 'auth_token_logins',
        'remember_tokens'   => 'auth_remember_tokens',
        'groups_users'      => 'auth_groups_users',
        'permissions_users' => 'auth_permissions_users',
    ];
```

kenjis:
> What has happened PHPStan! There is no error in local!

I also have no error on my local environment.
But `config()` returns `object|null`. 
So `null->authTables` is "Trying to get property 'authTables' of non-object".

kenjis:
@datamweb 
> And PHPUNIT test !

The test does not seem to pass on Windows from the beginning.

`CLI::write()` outputs `PHP_EOL`, but the test code uses `LF` as newline code.

src/Config/Auth.php
datamweb: Done.

datamweb:
@kenjis , thanks for the explanation, it made sense.
  But how can I fix error PHPStan?

src/Config/Constants.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Constants Name of Shield Tables
+ */
+define('SHIELD_TABLES', config('Auth')->tables);
```
kenjis, line 8: How about this?
```suggestion
/** @var \CodeIgniter\Shield\Config\Auth authConfig */
$authConfig = config('Auth');
define('SHIELD_TABLES', $authConfig->tables);
```

datamweb:
I thought it could be fixed.
I'm surprised, even when I tried to pass the check with `@phpstan-ignore-line` the error was still detected.

src/Config/Constants.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Constants Name of Shield Tables
+ */
+/** @var \CodeIgniter\Shield\Config\Auth authConfig */
```
kenjis, line 8: Sorry, there was no `$`.
```suggestion
/** @var \CodeIgniter\Shield\Config\Auth $authConfig */
```

kenjis:
Why?

```
Error: Attempt to read property "tables" on null
 ------ ------------------------------------------- 
  Line   src/Config/Constants.php                   
 ------ ------------------------------------------- 
  10     Attempt to read property "tables" on null  
 ------ ------------------------------------------- 
```

datamweb:
**PHP 7.4:**
```console 
Error: Trying to get property 'tables' of non-object
 ------ ----------------------------------------------- 
  Line   src/Config/Constants.php                       
 ------ ----------------------------------------------- 
  10     Trying to get property 'tables' of non-object  
 ------ -----------------------------------------------
```

datamweb:
@kenjis , after many efforts, I finally found the solution.
But I don't know exactly where the main problem is, I think it's in the framework , please let me know if you understand something.

kenjis:
@datamweb I don't know what's wrong with PHPStan.
To begin with, I cannot reproduce the error.

But why don't we remove the constant?
I think using global constant is not good practice especially in libraries.
Global constants are for a user app, and a library should not consume them.

Ion Auth also havs $table, but it does not use a constant.
https://github.com/benedmunds/CodeIgniter-Ion-Auth/blob/6ebcd542011dbc35045466acad0172d2b68ae051/Config/IonAuth.php#L49-L54

We can get the values with `config('Auth')->tables`.

datamweb:
> But why don't we remove the constant?

The reason I used **constant** was to add this feature with minimal changes.(
const
) vs 
(wo_const)

Also, I think it is easier to read and review with this implementation. For example, look at PR [commit](https://github.com/codeigniter4/shield/pull/617/commits/85180598ca3b4e9b42f3cd58f2187d8db2be7f85#diff-74b8b9ef03e167dc3bc070a014429d38eeea0506f27edd419af96f658be1685aR10), it's easier to suggest a change for `SHIELD_TABLES['users']` or 
```php 
    /**
     * Auth Table names
     */
    private array $tables;

    public function __construct(?Forge $forge = null)
    {
        parent::__construct($forge);

        /** @var Auth $authConfig */
        $authConfig   = config('Auth');
        $this->tables = $authConfig->tables;
    }
// and ...

```
In fact, I believe that using **constant** makes our codes shorter and more readable.
Also I was going to include a constant for the shield version, which now I don't know where it fits with the no **constant** implementation.

Technically yes you are right, but we have an example of using **constant** in Bonfire2.

src/Config/Registrar.php
```diff
@@ -14,6 +14,8 @@
 use CodeIgniter\Shield\Filters\SessionAuth;
 use CodeIgniter\Shield\Filters\TokenAuth;
 
+include_once __DIR__ . '/Constants.php';
```
kenjis, line 17: I don't know we should use the constant or not.
But even if we use it, I would like to remove the include_once line and the constant file.

kenjis:
If we add `SHIELD_VERSION`, we should add it as a class constant like `CodeIgniter\CodeIgniter::CI_VERSION`.
Class constants do not consume the global namespace.

datamweb:
Okay, we will continue in your PR.

PR #629, opened 2023-02-07
fix: Custom login fields should work for logging in.

### LINKED ISSUE #334: Support for custom login credentials
Hi all,
When we create a custom column in table `users`, for example `phone_number`.
And then we set the value to 
```php 
    /**
     * Valid login fields
     * Fields that are available to be used as credentials for login.
     */
    public array $validFields = [
      //  'email',
        //'username',
        'phone_number',
    ];
```
It is not possible to log into the system. If there should be no problem to login.

We need to make some changes in this area.

https://github.com/codeigniter4/shield/blob/7855d91908f78a0a67b6e40eb877979bd34f52ef/src/Authentication/Authenticators/Session.php#L251-L258

Do you agree?

lonnieezell:
Closes #334

Previously, only `email` and `username` were actually working as valid fields to login with. This was not the original intention. 

This changes this to ensure that any field listed within `$validFields` that is within the `users` table can be used as a valid login credential for the Session authenticator. All login attempts must still have a password associated.

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -247,26 +247,28 @@ This only works with the Session authenticator.
 1. Create a [migration](http://codeigniter.com/user_guide/dbmgmt/migration.html) that adds a new column to the `users` table.
 2. Edit `app/Config/Auth.php` so that the new column you just created is within the `$validFields` array.
 
-```php
-public array $validFields = [
-    'employee_id'
-];
-```
+    ```php
+    public array $validFields = [
+        'employee_id'
+    ];
+    ```
 
-If you have multiple login forms on your site that use different credentials, you must have all of the valid identifying fields in the array.
+    If you have multiple login forms on your site that use different credentials, you must have all of the valid identifying fields in the array.
 
-```php
-public array $validFields = [
-    'email',
-    'employee_id'
-];
-```
+    ```php
+    public array $validFields = [
+        'email',
+        'employee_id'
+    ];
+    ```
+    > **Warning**
+    > It is very important for security that if you add a new column for identifier you must write a new **Validation Rules** and then set it using the [custom-validation-rules](https://github.com/codeigniter4/shield/blob/develop/docs/customization.md#custom-validation-rules) description.
 
 3. Edit the login form to change the name of the default `email` input to the new field name.
 
-```php
-<!-- Email -->
-<div class="mb-2">
-    <input type="text" class="form-control" name="employee_id" autocomplete="new-employee-id" placeholder="12345" value="<?= old('employee_id') ?>" required />
-</div>
-```
+    ```php
+    <!-- Email -->
+    <div class="mb-2">
+        <input type="text" class="form-control" name="employee_id" autocomplete="new-employee-id" placeholder="12345" value="<?= old('employee_id') ?>" required />
+    </div>
+    ```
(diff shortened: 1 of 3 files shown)


docs/customization.md
```diff
@@ -236,3 +237,36 @@ After creating the class, set the `$userProvider` property in **app/Config/Auth.
...
+```
+
+If you have multiple login forms on your site that use different credentials, you must have all of the valid identifying fields in the array.
+
+```php
+public array $validFields = [
+    'email',
+    'employee_id'
+];
+```
```
datamweb, line 263:
```suggestion
    ```php
    public array $validFields = [
        'employee_id'
    ];
    ```

    If you have multiple login forms on your site that use different credentials, you must have all of the valid identifying fields in the array.

    ```php
    public array $validFields = [
        'email',
        'employee_id'
    ];
    ```
    > **Warning**
    > It is very important for security that if you add a new column for identifier you must write a new **Validation Rules** and then set it using the [custom-validation-rules](https://github.com/codeigniter4/shield/blob/develop/docs/customization.md#custom-validation-rules) description.
```

docs/customization.md
```diff
@@ -236,3 +237,36 @@ After creating the class, set the `$userProvider` property in **app/Config/Auth.
...
+
+3. Edit the login form to change the name of the default `email` input to the new field name.
+
+```php
+<!-- Email -->
+<div class="mb-2">
+    <input type="text" class="form-control" name="employee_id" autocomplete="new-employee-id" placeholder="12345" value="<?= old('employee_id') ?>" required />
+</div>
+```
```
datamweb, line 272:
```suggestion
    ```php
    <!-- Email -->
    <div class="mb-2">
        <input type="text" class="form-control" name="employee_id" autocomplete="new-employee-id" placeholder="12345" value="<?= old('employee_id') ?>" required />
    </div>
    ```
```

docs/customization.md
datamweb: Screenshot 2023-02-08 043740

**lonnieezell** , I know **paulbalandan**  was concerned about security, I added a warning, you can change the text.

lonnieezell: That text sounds pretty good to me, actually. Thanks for pointing that step out.

tests/Authentication/Authenticators/SessionAuthenticatorTest.php
```diff
@@ -433,4 +440,43 @@ public function testAttemptUsernameOnly(): void
...
+     * Test that any field within the user table can be used as the
+     * login identifier.
+     *
+     * @see https://github.com/codeigniter4/shield/issues/334
+     */
+    public function testAttemptCustomField(): void
+    {
+        // We don't need email, but do need a password set....
+        $this->user->createEmailIdentity([
+            'email'    => '',
```
datamweb, line 454: PHPUnit / PHP 7.4 - OCI8 failed. Because Oracle sees `''` as `NULL `.
We have faced this problem before and used the following method to fix it.

https://github.com/kenjis/codeigniter-shield/blob/75c37d0de32937b258156a8819478062111d5f16/src/Models/LoginModel.php#L61-L63

More Info see : #402

lonnieezell: Oh, great. Thanks!

datamweb (approved):
LGTM!
Very thanks for addressing this issue.

lonnieezell:
Glad to see this one fixed, myself! you're welcome.

PR #641, opened 2023-02-16
fix: protect route `change-password` via `force_reset` if usre not login

datamweb:
Hello,
filter `session `may not be used as `globals`.
So a user should not have access to page **change-password**.
I don't consider it a good practice not to control this issue in filter `force_reset`.


datamweb:
I changed my mind. I close it.

PR #665, opened 2023-03-12
lang: update sk translation

jozefrebjak:

PR #669, opened 2023-03-13
test: fix failed test because of DST

kenjis:
See https://github.com/codeigniter4/shield/pull/665#issuecomment-1465333910


datamweb (approved):
どうも

PR #685, opened 2023-03-18
Notification on existing e-mail on the registration page.

atakanau:
Replace 
"`The Email Address field must contain a unique value.`" 
with 
"`Email is already registered. Try logging in.`"


atakanau:
register

datamweb:
Thank you for PR.
Please run the following command to fix error PHPCSFixer.

```console
composer cs-fix
```
We at SHIELD have the possibility to customize the rules, did you send this PR knowing about this?

https://github.com/codeigniter4/shield/blob/develop/docs/customization.md#custom-validation-rules

PR #695, opened 2023-04-10
fix: only validate data from POST request body

miguel-rn:
With the current data validation checks in the controllers, an empty POST request with valid GET parameters will cause the validation to pass. Since this data is always later fetched from the POST request body, it will be null and errors will be thrown.

### DIFF
--- a/src/Controllers/LoginController.php
+++ b/src/Controllers/LoginController.php
@@ -47,7 +47,7 @@ class LoginController extends BaseController
         // like the password, can only be validated properly here.
         $rules = $this->getValidationRules();
 
-        if (! $this->validate($rules)) {
+        if (! $this->validateData($this->request->getPost(), $rules)) {
             return redirect()->back()->withInput()->with('errors', $this->validator->getErrors());
         }
 
--- a/src/Controllers/MagicLinkController.php
+++ b/src/Controllers/MagicLinkController.php
@@ -65,7 +65,7 @@ class MagicLinkController extends BaseController
     {
         // Validate email format
         $rules = $this->getValidationRules();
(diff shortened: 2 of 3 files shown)


kenjis (approved):
Thank you. You're correct. Using `$this->validate()` is not a good practice.

datamweb (approved):

miguel-rn:
I noticed a related issue in the docs, I will make a pull request to update.

sammyskills:
This is good, but I have to ask.

_Is it possible for a `$_GET` request to pass through a route defined as `$_POST`, i.e., `$route->post()` ?_

sammyskills:
Maybe I didn't explain properly, I was actually talking about the `GET` request, in respect to the route definitions:
```php
$routes->get();
$routes->post();
```
For a route defined explicitly as `$route->post()`, can `$this->request->getGet()` pass through it?

sammyskills:
My main concern is, for example, there is a route definition for the `loginAction()` method like so:
```php
$routes->post('login', 'LoginController::loginAction');
```
Is it possible that `GET` parameters will pass the validation using `$this->validate()` method?

kenjis:
> For a route defined explicitly as `$route->post()`, can `$this->request->getGet()` pass through it?

Yes. Because `$this->request->getGet()` returns `$_GET` values.

> Is it possible that `GET` parameters will pass the validation using `$this->validate()` method?

Yes. Because `$this->validate()` maybe validates `$request->getVar()` data.
See the Warning: https://codeigniter4.github.io/CodeIgniter4/incoming/controllers.html#this-validate

PR #698, opened 2023-04-11
docs: update guides/mobile_apps.md

sammyskills:
1. Currently, the documentation on mobile apps does not include the password field for logging in, which makes the request to fail. This PR adds the complete credentials for logging in, as seen below:
https://github.com/codeigniter4/shield/blob/799c6477504dac7dc0dc45e2e9c1d8a063e23ca8/src/Controllers/LoginController.php#L54-L56

2. The `device_name` POST request is required to generate the access token, so, it is supposed to be included in the validation rules. Not passing it returns this error:
```
Argument 1 passed to CodeIgniter\\Shield\\Entities\\User::generateAccessToken() must be of the type string, null given
```

### DIFF
--- a/docs/guides/mobile_apps.md
+++ b/docs/guides/mobile_apps.md
@@ -40,8 +40,13 @@ class LoginController extends BaseController
                 ->setStatusCode(422);
         }
 
+        // Get the credentials for login
+        $credentials             = $this->request->getPost(setting('Auth.validFields'));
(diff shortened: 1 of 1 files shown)


sammyskills:
@kenjis, @lonnieezell, @datamweb Can you take a look?

kenjis (approved):

kenjis:
@sammyskills Thank you!

PR #703, opened 2023-04-18
fix: incorrect base class for exceptions

kenjis:
All exceptions by Shield should implement BaseException.


lonnieezell (approved):

kenjis:
@lonnieezell Thank you!

PR #712, opened 2023-04-24
fix: validation password 

GeorgKott:
password validation has error when email doesn't contain the @ symbol


paulbalandan:
How can an email address not contain the @ symbol?

GeorgKott:
> How can an email address not contain the @ symbol?

Because we check the email for a valid value and then check the password with the same email value, even if the email is invalid. And any value can be sent from the frontend

datamweb:
@GeorgKott Thanks to PR.

Please answer these questions?

1. Do you use `service('passwords')` directly?
2. Why not use defined rules for email validation?
   https://github.com/codeigniter4/shield/blob/35fcaaa2b8d5902b49553776fa86a0473c81c7dc/src/Config/AuthSession.php#L26-L35

Please sign the all commits:

https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/workflow.md#gpg-signing-old-commits

and please write php unit test see:

https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#unit-testing

kenjis:
> Because we check the email for a valid value and then check the password with the same email value, even if the email is invalid. And any value can be sent from the frontend

If the email is invalid, wouldn't the validation fail and complete at that point?
Please write test code to reproduce that error.

PR #714, opened 2023-04-26
docs: improve descriptions on permissions

kenjis:
See https://forum.codeigniter.com/showthread.php?tid=87457

- add description for user-level and group-level permissions
- update the comments in `Config\AuthGroups`.


datamweb (approved):
thanks, I think change will help the following:

https://github.com/codeigniter4/shield/blob/ce68e9f7f0a5170adc327493e3f9cba02210837a/src/Config/AuthGroups.php#L75-L80

kenjis:
Updated the comments in Config\AuthGroups.

datamweb (approved):
LGTM! 👍

PR #723, opened 2023-04-30
fix: nothing personal validator error with bad email value

GeorgKott:
**superseded:** #712 

Sorry for my previous pull request. I closed it because after applying the gpg addition, conflicts occurred in the code for earlier commits. 

after that
> git rebase -i --root --exec 'git commit --amend --no-edit --no-verify -S'

And I thought it would be easier to create a new pull request than to resolve these conflicts. 
I also added some tests (integration and unit) that describe the issue I faced.

If you don't apply the changes I made, I will get an error in the tests. 

```
There was 1 error:

1) Tests\Controllers\RegisterTest::testRegisterActionWithBadEmailValue
ErrorException: Undefined offset: 1

\PhpstormProjects\shield\src\Authentication\Passwords\NothingPersonalValidator.php:78
\PhpstormProjects\shield\src\Authentication\Passwords\NothingPersonalValidator.php:27
\PhpstormProjects\shield\src\Authentication\Passwords.php:132
\PhpstormProjects\shield\src\Authentication\Passwords\ValidationRules.php:45
\PhpstormProjects\shield\vendor\codeigniter4\framework\system\Validation\Validation.php:311
\PhpstormProjects\shield\vendor\codeigniter4\framework\system\Validation\Validation.php:170
\PhpstormProjects\shield\vendor\codeigniter4\framework\system\Controller.php:155
\PhpstormProjects\shield\src\Controllers\RegisterController.php:103
\PhpstormProjects\shield\vendor\codeigniter4\framework\system\CodeIgniter.php:934
\PhpstormProjects\shield\vendor\codeigniter4\framework\system\CodeIgniter.php:499
\PhpstormProjects\shield\vendor\codeigniter4\framework\system\CodeIgniter.php:368
\PhpstormProjects\shield\vendor\codeigniter4\framework\system\Test\FeatureTestTrait.php:181
\PhpstormProjects\shield\vendor\codeigniter4\framework\system\Test\FeatureTestTrait.php:226
\PhpstormProjects\shield\tests\Controllers\RegisterTest.php:300
```
i think that a POST request can contain any data, not necessarily a valid email value. After all, it's not the front-end's responsibility to validate, but the back-end.
Thank you

### DIFF
--- a/src/Authentication/Passwords/NothingPersonalValidator.php
+++ b/src/Authentication/Passwords/NothingPersonalValidator.php
@@ -72,10 +72,15 @@ class NothingPersonalValidator extends BaseValidator implements ValidatorInterfa
             $needles = $this->strip_explode($userName);
 
             // extract local-part and domain parts from email as separate needles
-            [
-                $localPart,
-                $domain,
-            ] = explode('@', $email);
+            if (str_contains($email, '@')) {
+                [
+                    $localPart,
+                    $domain,
+                ] = explode('@', $email);
+            } else {
+                $localPart = $email;
+                $domain    = null;
+            }
             // might be john.doe@example.com and we want all the needles we can get
             $emailParts = $this->strip_explode($localPart);
             if (! empty($domain)) {
--- a/tests/Controllers/RegisterTest.php
+++ b/tests/Controllers/RegisterTest.php
@@ -294,6 +294,24 @@ final class RegisterTest extends DatabaseTestCase
         $result->assertRedirectTo(config('Auth')->registerRedirect());
     }
 
+    public function testRegisterActionWithBadEmailValue(): void
+    {
+        $result = $this->withSession()->post('/register', [
+            'username'         => 'JohnDoe',
+            'email'            => 'john.doe',
+            'password'         => '123456789aa',
+            'password_confirm' => '123456789aa',
+        ]);
+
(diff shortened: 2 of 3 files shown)


datamweb (approved):

kenjis (approved):
LGTM! Thank you.

src/Authentication/Passwords/NothingPersonalValidator.php
```diff
@@ -72,10 +72,15 @@ protected function isNotPersonal(string $password, ?User $user): bool
             $needles = $this->strip_explode($userName);
 
             // extract local-part and domain parts from email as separate needles
-            [
-                $localPart,
-                $domain,
-            ] = explode('@', $email);
+            if (str_contains($email, '@')) {
```
kenjis, line 75: str_contains() requires PHP8. So using it here is a bug.
But we use polyfill-php80, so the test does not fail on PHP 7.4.

```
$ composer why symfony/polyfill-php80
friendsofphp/php-cs-fixer v3.16.0 requires symfony/polyfill-php80 (^1.27)
```

paulbalandan: But it is a transitive dependency of a dev dependency, so using it here is still a bug.

PR #728, opened 2023-05-02
fix: typo in `getMaxLengthRule()` method name

bgeneto:
The getMaxLenghtRule() method name has a typo and should be renamed to getMaxLengthRule() in 3 source files.


datamweb (approved):
Thanks, good catch.

kenjis (approved):
Thank you!

PR #736, opened 2023-05-12
Customization db group

arashsaffari:
This modification is made to add the capability of changing the name of the database connection. #727 

Please note that the variable "DBGroup" cannot be changed in the "initialize()" function. It needs to be changed in the "__construct()" function.


kenjis:
You signed only one commit, but we need you to sign all your commits.

Don't use `git merge`. Use `git rebase` instead.

kenjis:
Read this: https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md

And if you don't know git well, it may be better to create new another PR.

datamweb:
@arashsaffari , first of all I want to make sure which of the following:
Execute command:
```console
git config --global commit.gpgsign true
```
Then run the following command:
```console
git config --global user.signingkey 5FE32B838BA49CD5
```
The above causes all commits to be signed automatically from now on.

```console
git fetch upstream
git switch develop
git merge upstream/develop
git switch customization_db_group
git branch -m customization_db_group customization_db_group.bk1
git switch -c customization_db_group develop
code .
```

After that, change **src/Config/Auth.php** and **src/Models/BaseModel.php** and run:
```console
composer cs-fix
```

```console
git add .
git commit -m "feat: add `DBGroup` for customization db group"
git push -f origin customization_db_group
```

datamweb:
@arashsaffari I tried to implement this feature in the branch below. If you find the time to test it, please let us know.

https://github.com/datamweb/shield/tree/add_customization_db_group

kenjis:
The first commit is not signed.

And please update your branch to pass GitHub Action checks.
See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/workflow.md#updating-your-branch

PR #742, opened 2023-05-15
fix: simplify extraction of email parts

paulbalandan:
Replace call to PHP 8.0's `str_contains`.


kenjis (approved):

datamweb:
@paulbalandan thank you!

Issue #741, opened 2023-05-15
Bug: Database exceptions "Specified key was too long; max key length is 1000 bytes"


kpeu3u:
### PHP Version

7.4

### CodeIgniter4 Version

4.3.4

### Shield Version

develop

### Which operating systems have you tested for this bug?

Linux

### Which server did you use?

apache

### Database

MariaDB 10.6.13

### Did you customize Shield?

no

### What happened?

If deploy to shared hosting and MySQL default_storage_engine is set to MyISAM there is no possibility to change it to InnoDB, spark migrate --all return Exceptions
 
```console
CodeIgniter v4.3.4 Command Line Tool - Server Time: 2023-05-14 15:46:19 UTC+00:00

Running all new migrations...

[CodeIgniter\Database\Exceptions\DatabaseException]

Specified key was too long; max key length is 1000 bytes

at SYSTEMPATH/Database/BaseConnection.php:646
```

### Steps to Reproduce

run `spark migrate --all` return on MySQL server configured with default_storage_engine=MyISAM

### Expected Output

Migrations complete without error.

### Anything else?

It is a good to have `['ENGINE' => 'InnoDB']` set as an attribute in createTable methods for tables that uses Foreign key or to all methods.

kenjis:
It is documented.
See https://codeigniter4.github.io/shield/install/#note-migration-error

datamweb:
As Kenjis pointed out, we don't consider this a bug.

However, I agree to make changes to the migration file. Can you sent PR?

kpeu3u:
@kenjis It is documented but in some cases someones have no superadmin or admin access on shared hosting servers and default_storage_engine can NOT be changed!

kenjis:
@kpeu3u That's too bad. If the default storage engine is MyISAM, it is out of dated.
But I understand that if you use such servers, you need `['ENGINE' => 'InnoDB']`.

Shield does not support MyISAM, and almost uses use InnoDB. So I'm fine to add `['ENGINE' => 'InnoDB']` to the migration file when using MySQL.

Can you send a PR for it?

kpeu3u:
I have sent PR [#745](https://github.com/codeigniter4/shield/pull/745), those changes also can be moved in Auth config.

datamweb:
> can be moved in Auth config.

I don't see any need for this.

gridphp:
I am still getting this error, installed latest build.

[...]

PR #745, opened 2023-05-16
feat: Add InnoDB engine to migration create table methods

kpeu3u:
resolved: #741

### DIFF
--- a/src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
+++ b/src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
@@ -16,17 +16,15 @@ class CreateAuthTables extends Migration
     private array $tables;
 
     private array $attributes;
-    private bool $ifNotExists;
 
     public function __construct(?Forge $forge = null)
     {
         parent::__construct($forge);
 
         /** @var Auth $authConfig */
-        $authConfig        = config('Auth');
-        $this->tables      = $authConfig->tables;
-        $this->ifNotExists = false;
-        $this->attributes  = ($this->db->getPlatform() === 'MySQLi') ? ['ENGINE' => 'InnoDB'] : [];
+        $authConfig       = config('Auth');
+        $this->tables     = $authConfig->tables;
+        $this->attributes = ($this->db->getPlatform() === 'MySQLi') ? ['ENGINE' => 'InnoDB'] : [];
     }
 
     public function up(): void
@@ -45,7 +43,7 @@ class CreateAuthTables extends Migration
         ]);
         $this->forge->addPrimaryKey('id');
         $this->forge->addUniqueKey('username');
-        $this->forge->createTable($this->tables['users'], $this->ifNotExists, $this->attributes);
+        $this->createTable($this->tables['users']);
 
         /*
          * Auth Identities Table
@@ -69,7 +67,7 @@ class CreateAuthTables extends Migration
         $this->forge->addUniqueKey(['type', 'secret']);
         $this->forge->addKey('user_id');
         $this->forge->addForeignKey('user_id', $this->tables['users'], 'id', '', 'CASCADE');
-        $this->forge->createTable($this->tables['identities'], $this->ifNotExists, $this->attributes);
+        $this->createTable($this->tables['identities']);
 
         /**
          * Auth Login Attempts Table
@@ -90,7 +88,7 @@ class CreateAuthTables extends Migration
         $this->forge->addKey(['id_type', 'identifier']);
         $this->forge->addKey('user_id');
         // NOTE: Do NOT delete the user_id or identifier when the user is deleted for security audits
-        $this->forge->createTable($this->tables['logins'], $this->ifNotExists, $this->attributes);
+        $this->createTable($this->tables['logins']);
 
         /*
          * Auth Token Login Attempts Table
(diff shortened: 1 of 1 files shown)


kenjis:
Add the attribute only when using MySQL.
```
There were 212 errors:

1) Tests\Authentication\ForcePasswordResetTest::testForceGlobalPasswordReset
CodeIgniter\Database\Exceptions\DatabaseException: SQLite3::exec(): near "'InnoDB'": syntax error

/home/runner/work/shield/shield/vendor/codeigniter4/framework/system/Database/BaseConnection.php:646
/home/runner/work/shield/shield/vendor/codeigniter4/framework/system/Database/Forge.php:546
/home/runner/work/shield/shield/src/Database/Migrations/2020-12-28-223112_create_auth_tables.php:48
/home/runner/work/shield/shield/vendor/codeigniter4/framework/system/Database/MigrationRunner.php:877
/home/runner/work/shield/shield/vendor/codeigniter4/framework/system/Database/MigrationRunner.php:188
/home/runner/work/shield/shield/vendor/codeigniter4/framework/system/Test/DatabaseTestTrait.php:151
/home/runner/work/shield/shield/vendor/codeigniter4/framework/system/Test/DatabaseTestTrait.php:109
/home/runner/work/shield/shield/vendor/codeigniter4/framework/system/Test/DatabaseTestTrait.php:55
/home/runner/work/shield/shield/vendor/codeigniter4/framework/system/Test/CIUnitTestCase.php:248
/home/runner/work/shield/shield/tests/_support/TestCase.php:21
/home/runner/work/shield/shield/tests/Authentication/ForcePasswordResetTest.php:31
```
https://github.com/codeigniter4/shield/actions/runs/4989110786/jobs/8932607159?pr=745

src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
```diff
@@ -15,13 +15,18 @@ class CreateAuthTables extends Migration
      */
     private array $tables;
 
+    private array $attributes;
+    private bool $ifNotExists;
```
datamweb, line 19: Currently, we use `false`. I think we don't need to use a new variable, using `false` is enough.
e.g.
```php
$this->forge->createTable($this->tables['users'], false, $this->attributes);
```

src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
```diff
@@ -15,13 +15,18 @@ class CreateAuthTables extends Migration
...
     {
         parent::__construct($forge);
 
         /** @var Auth $authConfig */
-        $authConfig   = config('Auth');
-        $this->tables = $authConfig->tables;
+        $authConfig        = config('Auth');
+        $this->tables      = $authConfig->tables;
+        $this->ifNotExists = false;
+        $this->attributes  = ($this->db->getPlatform() === 'MySQLi') ? ['ENGINE' => 'InnoDB'] : [];
```
datamweb, line 29: @kenjis , as far as I know, **MySQLi**, **OCI8**, **Postgre** and **SQLSRV** accept the attributes `['ENGINE' => 'InnoDB']`.
 And only `SQLite3` does not understand `['ENGINE' => 'InnoDB']`. Isn't it better to act as follows?
```php
$this->attributes = ($this->db->getPlatform() === 'SQLite3') ? [] :  ['ENGINE' => 'InnoDB'];
```

kenjis: No. `InnoDB` (and `ENGINE`) is configration only in MySQL. 
I believe OCI8, Postgre and SQLSRV does not understand the attribute. 
I don't know why these drivers do not report an error like SQLite.

src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
kenjis: It is certainly preferable to have no property that is not needed.

src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
```diff
@@ -40,7 +45,7 @@ public function up(): void
         ]);
         $this->forge->addPrimaryKey('id');
         $this->forge->addUniqueKey('username');
-        $this->forge->createTable($this->tables['users']);
+        $this->forge->createTable($this->tables['users'], $this->ifNotExists, $this->attributes);
```
kenjis, line 48: This code appears many times. Why don't you move it to a private method?

src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
kpeu3u: So it will stay like i did it, e.g. restricted to MySQL only!

src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
kpeu3u: i have removed property

src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
kpeu3u: i have moved create table to a private method, so we will get rid of repeated code

kenjis (approved):

datamweb (approved):

datamweb:
@kpeu3u Thank you!

PR #747, opened 2023-05-18
fix: add support for all request types when building a user from request.

### LINKED ISSUE #612: Bug: `ValidationRules::buildUserFromRequest()` does not support json input
### PHP Version

8

### CodeIgniter4 Version

4.3.0

### Shield Version

dev-develop

macOS

### Which server did you use?

apache

### Database

MariaDB

### Did you customize Shield?

```php
   protected function buildUserFromRequest(): User
    {
        $fields = $this->prepareValidFields();
        $request = service('request');
        $data = $request->getJSON() ? (array) $request->getJSON() : $request->getPost($fields);
        return new User($data);
    }
```

### What happened?

buildUserFromRequest support only form data inputs, not json input from service("request").
Before:
```php
$data = $request->getPost($fields);
```
After:
```php
$data = $request->getJSON() ? (array) $request->getJSON() : $request->getPost($fields);
```

### Steps to Reproduce

```php
$this->validation->setRules([
			"email" => ["label" => "Email", "rules" => "required|max_length[254]|valid_email"],
			"newPassword" => ["label" => "New password", "rules" => "required|strong_password"],
			"newPasswordAgain" => ["label" => "Repeat password", "rules" => "required|matches[newPassword]"],
		]);
```

it was catched when "strong_password" rule was applied.

### Expected Output

```php
/**
     * Builds a new user instance from the global request.
     */
    protected function buildUserFromRequest(): User
    {
        $fields = $this->prepareValidFields();

        /** @var IncomingRequest $request */
        $request = service('request');

        $data = $request->getJSON() ? (array) $request->getJSON() : $request->getPost($fields);

        return new User($data);
    }
```

### Anything else?

_No response_

sammyskills:
Previously, the `ValidationRules::buildUserFromRequest()` only checks for `POST`ed requests.

With this PR, it can now check for any type of request including json requests. 

Fixes #612

### DIFF
--- a/src/Authentication/Passwords/ValidationRules.php
+++ b/src/Authentication/Passwords/ValidationRules.php
@@ -67,6 +67,7 @@ class ValidationRules
      * Builds a new user instance from the global request.
      * 
      * @deprecated This will be removed soon.
+     * @see https://github.com/codeigniter4/shield/pull/747#discussion_r1198778666
      */
     protected function buildUserFromRequest(): User
     {
--- a/src/Authentication/Passwords/ValidationRules.php
+++ b/src/Authentication/Passwords/ValidationRules.php
@@ -67,6 +67,8 @@ class ValidationRules
     /**
      * Builds a new user instance from the global request.
      *
+     * @phpstan-ignore-next-line
+     *
      * @deprecated This will be removed soon.
      *
      * @see https://github.com/codeigniter4/shield/pull/747#discussion_r1198778666
--- a/src/Authentication/Passwords/ValidationRules.php
+++ b/src/Authentication/Passwords/ValidationRules.php
@@ -73,7 +73,7 @@ class ValidationRules
         /** @var IncomingRequest $request */
         $request = service('request');
 
-        $data = $request->getVar($fields);
+        $data = $request->getPost($fields);
 
         return new User($data);
     }
--- a/src/Authentication/Passwords/ValidationRules.php
+++ b/src/Authentication/Passwords/ValidationRules.php
@@ -65,6 +65,8 @@ class ValidationRules
 
     /**
      * Builds a new user instance from the global request.
+     * 
+     * @deprecated This will be removed soon.
      */
     protected function buildUserFromRequest(): User
     {
--- a/src/Authentication/Passwords/ValidationRules.php
+++ b/src/Authentication/Passwords/ValidationRules.php
@@ -65,8 +65,9 @@ class ValidationRules
 
     /**
      * Builds a new user instance from the global request.
-     * 
+     *
      * @deprecated This will be removed soon.
+     *
      * @see https://github.com/codeigniter4/shield/pull/747#discussion_r1198778666
      */
     protected function buildUserFromRequest(): User
--- a/src/Authentication/Passwords/ValidationRules.php
+++ b/src/Authentication/Passwords/ValidationRules.php
@@ -73,7 +73,7 @@ class ValidationRules
         /** @var IncomingRequest $request */
         $request = service('request');
 
-        $data = $request->getPost($fields);
+        $data = $request->getVar($fields);
 
         return new User($data);
     }
--- a/src/Authentication/Passwords/ValidationRules.php
+++ b/src/Authentication/Passwords/ValidationRules.php
@@ -40,6 +40,7 @@ class ValidationRules
         if (function_exists('auth') && auth()->user()) {
             $user = auth()->user();
         } else {
+            /** @phpstan-ignore-next-line */
             $user = empty($data) ? $this->buildUserFromRequest() : $this->buildUserFromData($data);
         }
 
@@ -66,8 +67,6 @@ class ValidationRules
 
     /**
      * Builds a new user instance from the global request.
-     *
-     * @phpstan-ignore-next-line
      *
      * @deprecated This will be removed soon.
      *
--- a/docs/guides/strengthen_password.md
+++ b/docs/guides/strengthen_password.md
@@ -17,7 +17,7 @@ The longer the password, the stronger it is. Consider increasing the value.
 
 > **Note**
 >
-> This checking works when you validate passwords with the `strong_password`
+> This checking works when you validate passwords with the `strong_password[]`
 > validation rule.
 >
 > If you disable `CompositionValidator` (enabled by default) in `$passwordValidators`,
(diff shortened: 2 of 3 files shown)


datamweb (commented):
LGTM.
See what Kenjis says.

src/Authentication/Passwords/ValidationRules.php
```diff
@@ -73,7 +73,7 @@ protected function buildUserFromRequest(): User
         /** @var IncomingRequest $request */
         $request = service('request');
 
-        $data = $request->getPost($fields);
+        $data = $request->getVar($fields);
```
kenjis, line 76: To be honest, I would like to deprecate `getVar()` because it is easy to create vulnerability to bypass the validation,
and we need to consider there is no attack vectors, but it is very difficult and takes a lot of time.
We must consider all cases that attacker could send.

Since `getVar()` retrieves data from various sources (POST data, GET data, JSON data, etc), only at the runtime we know which data it actually retrieves, because an attacker may send any kind of request. The data that a developer expects to be validated (e.g. POST data) may differ from the data that is actually validated (e.g. GET data), resulting in bypassing the validation because unvalidated data is processed. 


Are you sure there is no attack vectors that bypass this validation?

kenjis:
Can you tell me why you need this? A use case?

sammyskills:
> Can you tell me why you need this? A use case?

Yes. 

A use case is trying to register users using JWT authentication. 
If the developer expects the request to be sent as a JSON request, and the strong password rule`strong_password` is added to the validation rules, it will return an error.

src/Authentication/Passwords/ValidationRules.php
sammyskills: I understand your concern about security, but these are the areas I'm currently looking at:

1. If developers are asked to add a square bracket at the end of the rule like so: `strong_password[]`, just like you suggested in [this comment](https://github.com/codeigniter4/shield/issues/612#issuecomment-1411620121), it means that the docs for passwords validations would have to be updated.
2. If `getVar()` would pose some security issues, would it be better to first, check for `getPost()` and fall back to `getJSON()`?

jozefrebjak: @sammyskills I agree with @kenjis. `getVar()` is not good idea here.

datamweb: @sammyskills Can't you make a custom rule, e.g. (`strong_password_with_getJSON`)?

sammyskills: That's fine. So what do you think about my suggestion: 1 or 2?

sammyskills: > @sammyskills Can't you make a custom rule, e.g. (`strong_password_with_getJSON`)?

Of course, I can create a custom rule. But let's see what @kenjis or @lonnieezell has to say about my suggestions.

datamweb: > That's fine. So what do you think about my suggestion: 1 or 2?

To be honest, I already knew that Kenjis was against using `getVar()`. But I personally like the support for `getJSON()`, and I think it's worth adding.

sammyskills: Sure. 

We can still have support for JSON requests by doing something like so: 
```php
$request->getPost() ?? $request->getJSON();
```
But like I said, I need to hear from @kenjis for the best approach to take in this case. Who knows, we may even go for your suggestion of creating a new rule 😉

kenjis: I think `buildUserFromRequest()` should be removed. 
So I think using `strong_password[]` in the docs and the source code is better.

The true issue is `strong_password` guesses what data to be used for validation.
But `strong_password` never know what data should be used for validation.
It is determined by a developer (in a controller).
 In your use case, JSON data is the data to be used,
and POST data should not be used. But the following code uses POST data if exists.

```php
$request->getPost() ?? $request->getJSON();
```

All data for the validation is in JSON data in a request, and it is passed as `$data` if you use `strong_password[]`:
https://github.com/codeigniter4/shield/blob/a17a09cacffcace0cc3fa3957d2e0f23413e7470/src/Authentication/Passwords/ValidationRules.php#L31-L34
Why don't you use `$data`?

The responsibility of a validation rule is to validate passed data, not to guess what data to be used. The tight coupling of the validation rule and request object is a bad design. The problem arises because strong_password does unnecessary thing.

kenjis:
@sammyskills Thanks. I think `strong_password[]` works.

src/Authentication/Passwords/ValidationRules.php
sammyskills: > I think `buildUserFromRequest()` should be removed. So I think using `strong_password[]` in the docs and the source code is better.

Doing this _(removing `buildUserFromRequest()`)_ will result in a breaking change. All applications using the `strong_password` rule without the `[]` suffix will fail. I just tested it now.

What do you think?

kenjis: Yes, it is a breaking change. So we cannot remove it easily. I don't say do it now.

sammyskills: So what do you suggest for now?

To use `strong_password[]` for JSON requests and add an explanation in the jwt docs?

sammyskills: More so, since shield is still in beta, introducing a BC is not a bad idea at all.

kenjis: To use `strong_password[]` for all requests.

If you agree with removing `buildUserFromRequest()`, make it deprecated for now.

sammyskills: Of course I agree.

kenjis (approved):

kenjis:
We cannot remove the method, so suppress the PHPStan error.
```
 ------ -------------------------------------------------------------- 
  Line   src/Authentication/Passwords/ValidationRules.php              
 ------ -------------------------------------------------------------- 
  42     Call to deprecated method buildUserFromRequest() of class     
         CodeIgniter\Shield\Authentication\Passwords\ValidationRules:  
         This will be removed soon.                                    
 ------ -------------------------------------------------------------- 
```
https://github.com/codeigniter4/shield/actions/runs/5031981449/jobs/9025363859?pr=747

kenjis (approved):

kenjis:
@sammyskills Thank you!

PR #748, opened 2023-05-19
refactor: use ::class keyword in config()

kenjis:

PR #762, opened 2023-07-05
Revert "refactor: use ::class keyword in config()"

### LINKED ISSUE #761: Bug: can't use Auth config in different namespace than App
### PHP Version

8.1.16

### CodeIgniter4 Version

4.3.6

### Shield Version

dev-develop

macOS

### Which server did you use?

fpm-fcgi

### Database

MariaDB 10.2

### Did you customize Shield?

No.

### What happened?

After merge of https://github.com/codeigniter4/shield/pull/748 we cant use Auth config in different namespace than App.

### Steps to Reproduce

We are storing Auth as a module in `ROOTPATH/modules/Auth` folder. There is `Config` folder with `Auth.php` config. This worked before merge https://github.com/codeigniter4/shield/pull/748, after merge our App is broken. We recently updated to latest dev-develop commit hash and go back commit to commit to find when something changed.

We are using App/Config/Autoload.php to load module like:

```php
        'Auth'           => ROOTPATH . 'modules/Auth/',
```

### Expected Output

We would like to store Auth as module in our App.

### Anything else?

Workaround is to move `Auth.php` to `App/Config`.

kenjis:
Fixes #761
Reverts codeigniter4/shield#748

### DIFF
--- a/src/Auth.php
+++ b/src/Auth.php
@@ -8,8 +8,6 @@ use CodeIgniter\Router\RouteCollection;
 use CodeIgniter\Shield\Authentication\Authentication;
 use CodeIgniter\Shield\Authentication\AuthenticationException;
 use CodeIgniter\Shield\Authentication\AuthenticatorInterface;
-use CodeIgniter\Shield\Config\Auth as AuthConfig;
-use CodeIgniter\Shield\Config\AuthRoutes;
 use CodeIgniter\Shield\Entities\User;
 use CodeIgniter\Shield\Models\UserModel;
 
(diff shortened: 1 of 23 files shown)


paulbalandan:
Then, should similar reverts be made also in the other repos?

paulbalandan (approved):

kenjis:
I think we should use short classnames for `config()` in all library repositories.

MGatner:
Yes. Even if we find a way to make Factories prioritize children or certain namespaces this would still be a breaking change for non-App namespaced configs.

MGatner (approved):

PR #766, opened 2023-07-14
Update 2020-12-28-223112_create_auth_tables.php

solomon-ochepa:
The `$ifNotExists` parameter should be set to `true` to enable overriding.

### DIFF
--- a/src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
+++ b/src/Database/Migrations/2020-12-28-223112_create_auth_tables.php
@@ -175,6 +175,6 @@ class CreateAuthTables extends Migration
 
     private function createTable(string $tableName): void
     {
-        $this->forge->createTable($tableName, false, $this->attributes);
+        $this->forge->createTable($tableName, true, $this->attributes);
     }
 }


solomon-ochepa:
If the $ifNotExists is `false` (default), using a custom `users` table migration will throw an error - `Table 'users' already exists`; likewise to every other.

datamweb:
@solomon-ochepa thank you for PR.

> The `$ifNotExists` parameter should be set to `true` to enable overriding.

This is intentional. I am opposed to the overriding, and I believe it does more potential harm than good.

> If the $ifNotExists is `false` (default), using a custom `users` table migration will throw an error - `Table 'users' already exists`; likewise to every other.

In custom `users` table migration you **don't** have to create a **new table**, in fact you can add desired fields to `users` table.

```php
$fields = [
  'first_name' => [
  'type'       => 'text',
  'constraint' => '50',
  'null'       => TRUE
  ],
  'last_name' => [
  'type'       => 'VARCHAR',
  'constraint' => '50',
  'null'       => TRUE
  ],
];

$this->forge->addColumn('users', $fields);
```
more info see [here](https://github.com/codeigniter4/shield/discussions/333#discussioncomment-3311090).

kenjis:
> If the $ifNotExists is false (default), using a custom users table migration will throw an error - Table 'users' already exists; likewise to every other.

It should throw an error, if Table 'users' already exists.
Shield does not work without the table. So the migration should create the table.
The existing table might not have necessary columns.

Issue #777, opened 2023-08-09
Bug: controller run even if $allowMagicLinkLogins is false


mshannaq:
### PHP Version

8.2.7

### CodeIgniter4 Version

4.3.7

### Shield Version

dev-develop

### Which operating systems have you tested for this bug?

Linux

### Which server did you use?

apache

### Database

MySQl 8.0

### Did you customize Shield?

if you use direct link
http://localhost/login/magic-link
even if $allowMagicLinkLogins set to false in Config/Auth.php
the view form magic-link-login '\CodeIgniter\Shield\Views\magic_link_form' will be shown. and if the form submitted the email will be sent

### What happened?

disable magic link $allowMagicLinkLogins  = false in Config/Auth.php
goto http://localhost/login/magic-link directly

### Steps to Reproduce

disable magic link $allowMagicLinkLogins  = false in Config/Auth.php
goto http://localhost/login/magic-link directly

### Expected Output

not to show the form while allowMagicLinkLogins  = false

datamweb:
Hello @mshannaq , thanks for your report.
I sent PR #778.

kenjis:
@mshannaq Thank you for reporting. I've confirmed the bug.

PR #778, opened 2023-08-09
fix: Magic Link Login can be used even if `$allowMagicLinkLogins` is false

datamweb:
Fixed :#777


datamweb:
IMG_20230809_233657

kenjis:
Re-ran failed jobs.

kenjis:
We have three action methods.
```
$ php spark routes | grep MagicLinkController
| GET    | login/magic-link        | magic-link         | \CodeIgniter\Shield\Controllers\MagicLinkController::loginView     |                | toolbar       |
| GET    | login/verify-magic-link | verify-magic-link  | \CodeIgniter\Shield\Controllers\MagicLinkController::verify        |                | toolbar       |
| POST   | login/magic-link        | »                  | \CodeIgniter\Shield\Controllers\MagicLinkController::loginAction   |                | toolbar       |
```
Can you add the `if` statement to `verify()` just in case?

src/Language/ja/Auth.php
```diff
@@ -51,6 +51,7 @@
     'magicLinkExpired'   => '申し訳ございません、リンクは切れています。', // 'Sorry, link has expired.'
     'checkYourEmail'     => 'メールをチェックしてください！', // 'Check your email!'
     'magicLinkDetails'   => 'ログインリンクが含まれたメールを送信しました。これは {0} 分間だけ有効です。', // 'We just sent you an email with a Login link inside. It is only valid for {0} minutes.'
+    'magicLinkDisabled'  => '(To be translated) Use of MagicLink is currently not allowed.',
```
kenjis, line 54:
```suggestion
    'magicLinkDisabled'  => 'マジックリンクは使えません。', // 'Use of MagicLink is currently not allowed.'
```

datamweb:
> Can you add the `if` statement to `verify()` just in case?

Honestly, I thought about this, isn't it better to have a private function?
```php
    private function checkAllowMagicLink( ): RedirectResponse
   { 
        // Check if magic-link is not allowed
        if (!setting('Auth.allowMagicLinkLogins')) {
            return redirect()->route('login')->with('error', lang('Auth.magicLinkDisabled'));
        }
   }
```

kenjis (approved):

kenjis:
The three `if` statements are the exactly the same and the same meaning.
So it is better not to repeat them.

src/Controllers/MagicLinkController.php
```diff
@@ -50,6 +50,8 @@ public function __construct()
      */
     public function loginView()
     {
+        $this->displayErrorMagicLinkDisabled();
```
kenjis, line 53: `RedirectResponse` must be returned from this method if `allowMagicLinkLogins` is false.

datamweb:
@kenjis, method `verify(): RedirectResponse` uses native return type , using a private method causes **breaking change**. Is it worth it?
Or am I missing something?

kenjis:
> using a private method causes breaking change.

No, it is not a breaking change.

But this PR works, so I merge for now.

kenjis (approved):

PR #790, opened 2023-08-16
feat: add redirect denied for filter

datamweb:
See https://forum.codeigniter.com/showthread.php?tid=88262

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -25,6 +25,28 @@ following steps will be done.
     $this->helpers = array_merge($this->helpers, ['setting']);
     ```
 
+#### Config\Auth
+
+The following items have been added. Copy the properties in **src/Config/Auth.php**.
+
+- `$permission_denied` and `$group_denied` are added to `Config\Auth::$redirects`.
+- `permissionDeniedRedirect()` and `groupDeniedRedirect()` are added.
+
+#### Fix Custom Filter If extends `AbstractAuthFilter`
+If you have written a custom filter that extends `AbstractAuthFilter`. Now you need to add and implement method `redirectToDeniedUrl()` to your custom filter.
+The following example is related to the above explanation for **group** filter.
+```php
+/**
+ * If there is no necessary access to the group,
+ * it will redirect the user to the set URL with an error message.
+ */
+protected function redirectToDeniedUrl(): RedirectResponse
+{
+    return redirect()->to(config('Auth')->groupDeniedRedirect())
+        ->with('error', lang('Auth.notEnoughPrivilege'));
+} 
+```
+
 ## Version 1.0.0-beta.6 to 1.0.0-beta.7
 
 ### The minimum CodeIgniter version
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -29,16 +29,17 @@ following steps will be done.
 
 The following items have been added. Copy the properties in **src/Config/Auth.php**.
 
-- `$permission_denied` and `$group_denied` are added to `Config\Auth::$redirects`.
+- `permission_denied` and `group_denied` are added to `Config\Auth::$redirects`.
 - `permissionDeniedRedirect()` and `groupDeniedRedirect()` are added.
 
 #### Fix Custom Filter If extends `AbstractAuthFilter`
-If you have written a custom filter that extends `AbstractAuthFilter`. Now you need to add and implement method `redirectToDeniedUrl()` to your custom filter.
+
+If you have written a custom filter that extends `AbstractAuthFilter`, now you need to add and implement the `redirectToDeniedUrl()` method to your custom filter.
 The following example is related to the above explanation for **group** filter.
+
 ```php
 /**
- * If there is no necessary access to the group,
- * it will redirect the user to the set URL with an error message.
+ * If the user does not belong to the group, redirect to the configured URL with an error message.
  */
 protected function redirectToDeniedUrl(): RedirectResponse
 {
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -32,7 +32,7 @@ The following items have been added. Copy the properties in **src/Config/Auth.ph
 - `permission_denied` and `group_denied` are added to `Config\Auth::$redirects`.
 - `permissionDeniedRedirect()` and `groupDeniedRedirect()` are added.
 
-#### Fix Custom Filter If extends `AbstractAuthFilter`
+### Fix Custom Filter If extends `AbstractAuthFilter`
 
 If you have written a custom filter that extends `AbstractAuthFilter`, now you need to add and implement the `redirectToDeniedUrl()` method to your custom filter.
 The following example is related to the above explanation for **group** filter.
--- a/src/Config/Auth.php
+++ b/src/Config/Auth.php
@@ -65,10 +65,12 @@ class Auth extends BaseConfig
      * to apply any logic you may need.
      */
     public array $redirects = [
-        'register'    => '/',
-        'login'       => '/',
-        'logout'      => 'login',
-        'force_reset' => '/',
+        'register'                => '/',
+        'login'                   => '/',
+        'logout'                  => 'login',
+        'force_reset'             => '/',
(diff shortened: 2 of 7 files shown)


src/Filters/AbstractAuthFilter.php
```diff
@@ -7,14 +7,16 @@
...
-use CodeIgniter\HTTP\Response;
 use CodeIgniter\HTTP\ResponseInterface;
+use CodeIgniter\Shield\Config\Auth;
 
 /**
  * Group Authorization Filter.
  */
 abstract class AbstractAuthFilter implements FilterInterface
 {
+    protected string $filterName;
```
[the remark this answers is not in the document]
datamweb: @samsonasik  Thanks for the review, honestly, it blew my mind.

This implementation can only be used for `group` and `permission` (It is available in the shield by default) according to the defined `switch`.
 It effectively renders inheritance from extend useless.
It seems that it should be marked as `@internal`.

In this regard, I am not sure which is the better way.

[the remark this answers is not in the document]
samsonasik: I never use this library, so let see other opinions ;)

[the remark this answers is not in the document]
kenjis: @datamweb This change for `AbstractAuthFilter` is simply bad design.
Because abstraction should not know details (concrete classes).

The `AbstractAuthFilter` is the base class. Devs can extend it and create their own filters. But the change makes it impossible.

I think the default two filters is not enough for all needs. Someone may need a filter to check both group and permission.

src/Config/Auth.php
```diff
@@ -94,10 +94,12 @@ class Auth extends BaseConfig
...
-        'register'    => '/',
-        'login'       => '/',
-        'logout'      => 'login',
-        'force_reset' => '/',
+        'register'                => '/',
+        'login'                   => '/',
+        'logout'                  => 'login',
+        'force_reset'             => '/',
+        'after_permission_denied' => '/',
+        'after_group_denied'      => '/',
```
kenjis, line 102: It seems `after_` is not needed.
```suggestion
        'permission_denied' => '/',
        'group_denied'      => '/',
```

src/Config/Auth.php
```diff
@@ -447,6 +449,28 @@ public function forcePasswordResetRedirect(): string
         return $this->getUrl($url);
     }
 
+    /**
+     * Returns the URL the user should be redirected to
+     * if permission denied.
+     */
+    public function afterPermissionDeniedRedirect(): string
```
kenjis, line 456: `permissionDeniedRedirect()` is fine.

src/Config/Auth.php
```diff
@@ -447,6 +449,28 @@ public function forcePasswordResetRedirect(): string
...
+        $url = setting('Auth.redirects')['after_permission_denied'];
+
+        return $this->getUrl($url);
+    }
+
+    /**
+     * Returns the URL the user should be redirected to
+     * if group denied.
+     */
+    public function afterGroupDeniedRedirect(): string
```
kenjis, line 467: `groupDeniedRedirect()` is fine.

src/Config/Auth.php
datamweb: Done.

src/Filters/AbstractAuthFilter.php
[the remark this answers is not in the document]
datamweb: @Kenjis I changed the implementation, I think now there is no problem to create a custom filter that inherits from `AbstractAuthFilter`.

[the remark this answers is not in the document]
kenjis: Much better!

But this is still a breaking change. The `redirectToDeniedUrl()` method must be added if a dev extends `AbstractAuthFilter`.

Do we write to documentation for upgrade?
https://github.com/codeigniter4/shield/blob/develop/UPGRADING.md
Or if we add the default implementation (redirect to `/`)  in `AbstractAuthFilter`, it will not be a breaking change.

datamweb:
Added how to upgrade and use config('Auth').

UPGRADING.md
```diff
@@ -25,6 +25,28 @@ following steps will be done.
...
     ```
 
+#### Config\Auth
+
+The following items have been added. Copy the properties in **src/Config/Auth.php**.
+
+- `$permission_denied` and `$group_denied` are added to `Config\Auth::$redirects`.
+- `permissionDeniedRedirect()` and `groupDeniedRedirect()` are added.
+
+#### Fix Custom Filter If extends `AbstractAuthFilter`
```
kenjis, line 35:
```suggestion
#### Fix Custom Filter If extends `AbstractAuthFilter`

```

UPGRADING.md
```diff
@@ -25,6 +25,28 @@ following steps will be done.
...
+If you have written a custom filter that extends `AbstractAuthFilter`. Now you need to add and implement method `redirectToDeniedUrl()` to your custom filter.
+The following example is related to the above explanation for **group** filter.
```
kenjis, line 37:
```suggestion
The following example is related to the above explanation for **group** filter.

```

UPGRADING.md
```diff
@@ -25,6 +25,28 @@ following steps will be done.
...
 
+#### Config\Auth
+
+The following items have been added. Copy the properties in **src/Config/Auth.php**.
+
+- `$permission_denied` and `$group_denied` are added to `Config\Auth::$redirects`.
+- `permissionDeniedRedirect()` and `groupDeniedRedirect()` are added.
+
+#### Fix Custom Filter If extends `AbstractAuthFilter`
+If you have written a custom filter that extends `AbstractAuthFilter`. Now you need to add and implement method `redirectToDeniedUrl()` to your custom filter.
```
kenjis, line 36:
```suggestion
If you have written a custom filter that extends `AbstractAuthFilter`, now you need to add and implement the `redirectToDeniedUrl()` method to your custom filter.
```

src/Filters/GroupFilter.php
```diff
@@ -17,4 +19,14 @@ protected function isAuthorized(array $arguments): bool
     {
         return auth()->user()->inGroup(...$arguments);
     }
+
+    /**
+     * If there is no necessary access to the group,
+     * it will redirect the user to the set URL with an error message.
```
kenjis, line 25:
```suggestion
     * If the user does not belong to the groups, redirect to the configured URL with an error message.
```

src/Filters/PermissionFilter.php
```diff
@@ -23,4 +25,14 @@ protected function isAuthorized(array $arguments): bool
 
         return false;
     }
+
+    /**
+     * If there is no necessary access to the permission,
+     * it will redirect the user to the set URL with an error message.
```
kenjis, line 31:
```suggestion
     * If the user does not have the permission, redirect to the configured URL with an error message.
```

UPGRADING.md
```diff
@@ -25,6 +25,28 @@ following steps will be done.
...
+The following example is related to the above explanation for **group** filter.
+```php
+/**
+ * If there is no necessary access to the group,
+ * it will redirect the user to the set URL with an error message.
```
kenjis, line 41:
```suggestion
 * If the user does not belong to the group, redirect to the configured URL with an error message.
```

UPGRADING.md
```diff
@@ -25,6 +25,29 @@ following steps will be done.
...
     ```
 
+#### Config\Auth
+
+The following items have been added. Copy the properties in **src/Config/Auth.php**.
+
+- `permission_denied` and `group_denied` are added to `Config\Auth::$redirects`.
+- `permissionDeniedRedirect()` and `groupDeniedRedirect()` are added.
+
+#### Fix Custom Filter If extends `AbstractAuthFilter`
```
kenjis, line 35: This is not for "Mandatory Config Changes".
```suggestion
### Fix Custom Filter If extends `AbstractAuthFilter`
```

kenjis (approved):
LGTM! Thanks.

PR #791, opened 2023-08-16
feat: add support for multiple permissions check for users

sammyskills:
Added support for multiple permissions check for users via the `Authorizable` trait.

With this pull request, instead of writing:
```php
if ($user->can('user.create') || $user->can('user.edit')){
    //
}
```
we use this:
```php
if ($user->can('user.create', 'user.edit')){
    //
}
```

### DIFF
--- a/docs/authorization.md
+++ b/docs/authorization.md
@@ -119,7 +119,8 @@ permissions (**group-level permissions**) to determine if they are allowed.
 if ($user->can('users.create')) {
     //
 }
-// Or
+
+// If multiple permissions are specified, true is returned if the user has any of them.
 if ($user->can('users.create', 'users.edit')) {
     //
 }
--- a/docs/authorization.md
+++ b/docs/authorization.md
@@ -119,7 +119,7 @@ permissions (**group-level permissions**) to determine if they are allowed.
 if ($user->can('users.create')) {
     //
 }
-OR
+// Or
 if ($user->can('users.create', 'users.edit')) {
     //
 }
--- a/docs/authorization.md
+++ b/docs/authorization.md
@@ -111,12 +111,16 @@ The `Authorizable` trait on the `User` entity provides the following methods to 
 
 #### can()
 
-Allows you to check if a user is permitted to do a specific action. The only argument is the permission string. Returns
+Allows you to check if a user is permitted to do a specific action or group or actions. The permission string(s) should be passed as the argument(s). Returns
 boolean `true`/`false`. Will check the user's direct permissions (**user-level permissions**) first, and then check against all of the user's groups
 permissions (**group-level permissions**) to determine if they are allowed.
 
 ```php
 if ($user->can('users.create')) {
+    //
+}
+OR
+if ($user->can('users.create', 'users.edit')) {
     //
 }
 ```
--- a/src/Authorization/Traits/Authorizable.php
+++ b/src/Authorization/Traits/Authorizable.php
@@ -226,9 +226,10 @@ trait Authorizable
 
     /**
      * Checks user permissions and their group permissions
-     * to see if the user has a specific permission.
+     * to see if the user has a specific permission or group
+     * of permissions.
      *
-     * @param string $permission string(s) consisting of a scope and action, like `users.create`
+     * @param string $permissions string(s) consisting of a scope and action, like `users.create`
      */
     public function can(string ...$permissions): bool
     {
--- a/src/Authorization/Traits/Authorizable.php
+++ b/src/Authorization/Traits/Authorizable.php
@@ -228,47 +228,50 @@ trait Authorizable
      * Checks user permissions and their group permissions
      * to see if the user has a specific permission.
      *
-     * @param string $permission string consisting of a scope and action, like `users.create`
+     * @param string $permission string(s) consisting of a scope and action, like `users.create`
      */
-    public function can(string $permission): bool
+    public function can(string ...$permissions): bool
     {
-        if (strpos($permission, '.') === false) {
-            throw new LogicException(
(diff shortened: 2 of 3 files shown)


lonnieezell (approved):

tests/Authorization/AuthorizableTest.php
```diff
@@ -288,6 +288,13 @@ public function testCanCascadesToGroupsSimple(): void
         $this->assertTrue($this->user->can('admin.access'));
     }
 
+    public function testCanCascadesToGroupsMultiple(): void
```
datamweb, line 291: I want the tests to check more things.
For example, adding a few permission's with `addPermission()` and then checking the correct PR function.

docs/authorization.md
```diff
@@ -111,14 +111,18 @@ The `Authorizable` trait on the `User` entity provides the following methods to
...
-Allows you to check if a user is permitted to do a specific action. The only argument is the permission string. Returns
+Allows you to check if a user is permitted to do a specific action or group or actions. The permission string(s) should be passed as the argument(s). Returns
 boolean `true`/`false`. Will check the user's direct permissions (**user-level permissions**) first, and then check against all of the user's groups
 permissions (**group-level permissions**) to determine if they are allowed.
 
 ```php
 if ($user->can('users.create')) {
     //
 }
+OR
```
datamweb, line 122:
```suggestion
// Or
```

tests/Authorization/AuthorizableTest.php
sammyskills: Maybe I don't understand you clearly, but why should the `addPermission()` be called in the test? Or are you trying to say that this PR may fail when a permission is added?

datamweb: @sammyskills I'm sorry if I couldn't explain my meaning clearly.

Something like this:
```php
$user->addPermission('admin.access', 'beta.access');

$res= $this->user->can('beta.access','admin.access');
$this->assertTrue($res);

$res= $this->user->can('users.edit','users.delete');
$this->assertFalse($res);

$user->removePermission('admin.access');
$res= $this->user->can('admin.access','users.delete');
$this->assertFalse($res);

$res= $this->user->can('admin.access','beta.access');
$this->assertTrue($res);
```

sammyskills: @datamweb Thank you for the explanation.

`addPermission()` and `removePermission()` are already covered in the `AuthorizableTest`. Why should they be repeated here?

datamweb: My goal is not `addPermission()` or `removePermission()`, my goal is to test method `can()` if multiple threads are used.
Is this covered? If yes, link please?

sammyskills: > multiple threads

What do you mean?

datamweb: > What do you mean?

`$user->can('users.create', 'users.edit', 'etc')`

datamweb: `can()` will check the user's **direct permissions (user-level permissions) first**, and **then** check against all of the user's **groups permissions** (group-level permissions) to determine if they are allowed.

Q. Did the tests cover `$user->can('users.create', 'users.edit', 'etc')` in the first case(**direct permissions (user-level permissions**) first?

sammyskills: Oh, I now understand what you mean. 

The check for single permission is already covered in `testAddPermissionWithNoExistingPermissions()` and `testAddPermissionWithExistingPermissions()`, but not for multiple threads like you pointed out.

datamweb: Well, you get my point, 😂

src/Authorization/Traits/Authorizable.php
```diff
@@ -226,49 +226,53 @@ public function hasPermission(string $permission): bool
...
+                    'A permission must be a string consisting of a scope and action, like `users.create`.'
+                    . ' Invalid permission: ' . $permission
+                );
+            }
 
-        // Check user's permissions
-        if (in_array($permission, $this->permissionsCache, true)) {
-            return true;
-        }
+            $this->populatePermissions();
```
kenjis, line 245: It seems we don't need to call `populatePermissions()` again and again in the loop.

docs/authorization.md
```diff
@@ -111,14 +111,18 @@ The `Authorizable` trait on the `User` entity provides the following methods to
...
-Allows you to check if a user is permitted to do a specific action. The only argument is the permission string. Returns
+Allows you to check if a user is permitted to do a specific action or group or actions. The permission string(s) should be passed as the argument(s). Returns
 boolean `true`/`false`. Will check the user's direct permissions (**user-level permissions**) first, and then check against all of the user's groups
 permissions (**group-level permissions**) to determine if they are allowed.
 
 ```php
 if ($user->can('users.create')) {
     //
 }
+// Or
```
kenjis, line 122:
```suggestion

// If multiple permissions are specified, true is returned if the user has any of them.
```

kenjis:
The implementation looks good.

jozefrebjak:
@sammyskills maybe add test also for wildcard case `users.*`

sammyskills:
> @sammyskills maybe add test also for wildcard case `users.*`

Yes, you are correct @jozefrebjak. I have added that now.

datamweb (approved):
LGTM!
@sammyskills and @jozefrebjak thanks.

kenjis (approved):

kenjis:
@sammyskills Thank you!

PR #799, opened 2023-08-24
fix: unauthorized users for groups and permissions are not redirected to the previous url

### LINKED ISSUE #798: Bug: beforeLoginUrl is not stored in session if a user is logged out while in a protected group route.
### PHP Version

7.4

### CodeIgniter4 Version

4.3.7

### Shield Version

develop

Windows

### Which server did you use?

apache

### Database

MariaDB 10.4.18

### Did you customize Shield?

No.

### What happened?

There seems to be a problem with the merged PR #793.

As it stands, the entrance URL will only be filled if the route is protected by the session filter directly, as seen here: 
https://github.com/codeigniter4/shield/blob/970c67e746f02142c51badda176e0a4b26c353ab/src/Filters/SessionAuth.php#L78-L81

This looks and works fine if the pages of an application are protected using the configuration below in the **app/Config/Filters.php** file:
```php
public $globals = [
    'before' => [
        // ...
        'session' => ['except' => ['login*', 'register', 'auth/a/*']],
    ],
    // ...
];
```

A challenge arises when the application contains protected URLs for different user groups. For example, users: `users/dashboard`, admin: `admin/dashboard`. To make sure that users cannot access URLs they are not permitted, we apply the group filters: `group:user` and `group:admin` respectively. This would not have been a problem, but AFAIK, CI does not yet support arguments in filters via **app/Config/Filters.php** file, but it can be done via routes, in the **app/Config/Routes.php** file (see [docs](https://codeigniter.com/user_guide/incoming/filters.html#filter-arguments)), like so:
```php
$routes->get('users/dashboard', 'UsersController::index', ['filter' => 'group:user']);
$routes->get('admin/dashboard', 'UsersController::index', ['filter' => 'group:admin']);
```
With the configurations in the filter and routes, when we run `php spark filter:check get users/dashboard`, we get the following:
```
+--------+-----------------+----------------------------------------------------------+-----------------------------+
| Method | Route           | Before Filters                                           | After Filters               |
+--------+-----------------+----------------------------------------------------------+-----------------------------+
| GET    | users/dashboard | group session                                            | group                       |
+--------+-----------------+----------------------------------------------------------+-----------------------------+
``` 
From the output above, it means that if a user's session expired while trying to visit the `users/dashboard` page, there will be no temporary URL stored in the session, for redirection after logging in. This is because the `group` filter will run first, before the `session` filter. The **AbstractAuthFilter** redirects to the login page if a user is not logged in, before checking for the group:
https://github.com/codeigniter4/shield/blob/970c67e746f02142c51badda176e0a4b26c353ab/src/Filters/AbstractAuthFilter.php#L32-L34

A workaround is to add the filters (session and group) in the routes config file, but from CI docs, it is strongly discouraged as it breaks backward compatibility:

> Multiple filters is disabled by default. Because it breaks backward compatibility. If you want to use it, you need to configure. See [Multiple Filters for a Route](https://codeigniter.com/user_guide/installation/upgrade_415.html#upgrade-415-multiple-filters-for-a-route) for the details.

### Steps to Reproduce

Add this to the **GroupFilterTest**:
```php
public function testFilterNotAuthorizedStoresRedirectToEntranceUrlIntoSession(): void
{
    $result = $this->call('get', 'protected-route');

    $result->assertRedirectTo('/login');

    $session = session();
    $this->assertNotEmpty($session->get('beforeLoginUrl'));
    $this->assertSame(site_url('protected-route'), $session->get('beforeLoginUrl'));
}
```

### Expected Output

I expected that the entrance URL is stored in the session before logging out an unauthorized user.

### Anything else?

There are two possible solutions I am looking at:
1. Advice devs to use route to configure multiple filters and update the docs.
2. Store the entrance URL into session before logging a user out in the **AbstractAuthFilter** class.

sammyskills:
Fixes #798

### DIFF
--- a/src/Filters/AbstractAuthFilter.php
+++ b/src/Filters/AbstractAuthFilter.php
@@ -30,6 +30,12 @@ abstract class AbstractAuthFilter implements FilterInterface
         }
 
         if (! auth()->loggedIn()) {
+            // Set the entrance url to redirect a user after successful login
+            if (! url_is('login')) {
+                $session = session();
+                $session->setTempdata('beforeLoginUrl', current_url(), 300);
+            }
+
             return redirect()->route('login');
         }
 
--- a/tests/Authentication/Filters/GroupFilterTest.php
+++ b/tests/Authentication/Filters/GroupFilterTest.php
@@ -31,6 +31,17 @@ final class GroupFilterTest extends AbstractFilterTestCase
         $result->assertSee('Open');
     }
 
+    public function testFilterNotAuthorizedStoresRedirectToEntranceUrlIntoSession(): void
+    {
+        $result = $this->call('get', 'protected-route');
+
+        $result->assertRedirectTo('/login');
+
+        $session = session();
+        $this->assertNotEmpty($session->get('beforeLoginUrl'));
+        $this->assertSame(site_url('protected-route'), $session->get('beforeLoginUrl'));
+    }
+
     public function testFilterSuccess(): void
     {
         /** @var User $user */
--- a/tests/Authentication/Filters/GroupFilterTest.php
+++ b/tests/Authentication/Filters/GroupFilterTest.php
@@ -37,9 +37,8 @@ final class GroupFilterTest extends AbstractFilterTestCase
 
         $result->assertRedirectTo('/login');
 
-        $session = session();
-        $this->assertNotEmpty($session->get('beforeLoginUrl'));
-        $this->assertSame(site_url('protected-route'), $session->get('beforeLoginUrl'));
+        $this->assertNotEmpty(session()->getTempdata('beforeLoginUrl'));
+        $this->assertSame(site_url('protected-route'), session()->getTempdata('beforeLoginUrl'));
     }
 
     public function testFilterSuccess(): void
--- a/tests/Authentication/Filters/PermissionFilterTest.php
+++ b/tests/Authentication/Filters/PermissionFilterTest.php
@@ -37,9 +37,8 @@ final class PermissionFilterTest extends AbstractFilterTestCase
 
         $result->assertRedirectTo('/login');
 
-        $session = session();
-        $this->assertNotEmpty($session->get('beforeLoginUrl'));
-        $this->assertSame(site_url('protected-route'), $session->get('beforeLoginUrl'));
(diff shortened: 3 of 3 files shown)


datamweb (commented):
I am on mobile. I did not check the issue practically.

But the explanation and how to implement it seems logical.

sammyskills:
The option for adding arguments to filter configuration has been added to CI v4.4.0, see [docs](https://codeigniter4.github.io/CodeIgniter4/incoming/filters.html#id2).

This PR might need to be updated to either:
1. Update the docs and encourage developers to upgrade to v4.4.0
2. Maintain this PR to cover all use cases.

WDYT?

datamweb:
> The option for adding arguments to filter configuration has been added to CI v4.4.0, see [docs](https://codeigniter4.github.io/CodeIgniter4/incoming/filters.html#id2).

Yes, I remembered, I had reviewed this one😀.


> 1. Update the docs and encourage developers to upgrade to v4.4.0

Shield requires CodeIgniter 4.2.7+ at this point I am not very inclined to upgrade to 4.4.0.
The difference between CodeIgniter 4.2.7+ and 4.4.0 is huge, 4.4.0 is great, but I think we shouldn't limit the user community at this point.

sammyskills:
Hi @kenjis, @lonnieezell, @MGatner  can you take a look?

kenjis (approved):
LGTM!

datamweb:
@sammyskills is aware of this, I haven't played with the code yet, I'll try to do it tomorrow.

src/Filters/AbstractAuthFilter.php
```diff
@@ -30,6 +30,12 @@ public function before(RequestInterface $request, $arguments = null)
         }
 
         if (! auth()->loggedIn()) {
+            // Set the entrance url to redirect a user after successful login
+            if (! url_is('login')) {
+                $session = session();
+                $session->setTempdata('beforeLoginUrl', current_url(), 300);
```
kenjis, line 36: I just came up with this code, but it is a copy and paste. 
Since it is the same knowledge (save the URL for 300 seconds in the session), it might be better to put it together in a method if possible.

But there doesn't seem to be an appropriate class to put it in now.

datamweb (approved):
@sammyskills I came back after a million days :smiley:. Sorry.
Thank you!

PR #804, opened 2023-08-29
lang: [ar] update arabic translate

mshannaq:
Update src/Language/ar/Auth.php

### DIFF
--- a/src/Language/ar/Auth.php
+++ b/src/Language/ar/Auth.php
@@ -30,7 +30,7 @@ return [
     'password'        => 'كلمة المرور',
     'passwordConfirm' => 'كلمة المرور (مرة اخرى)',
     'haveAccount'     => 'هل لديك حساب بالفعل؟',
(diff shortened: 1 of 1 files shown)


src/Language/ar/Auth.php
```diff
@@ -30,7 +30,7 @@
     'password'        => 'كلمة المرور',
     'passwordConfirm' => 'كلمة المرور (مرة اخرى)',
     'haveAccount'     => 'هل لديك حساب بالفعل؟',
-    'token'           => '(To be translated) Token',
+    'token'           => 'رمز Token',
```
sammyskills, line 33:
```suggestion
    'token'           => 'رمز',
```


_"Token"_ is English, and should not exist here

kenjis: @mshannaq Can you update?

mshannaq:
done in https://github.com/codeigniter4/shield/pull/810

PR #806, opened 2023-08-29
fix: UserModel::assignIdentities() always produces a DB query

michalsn:
In the `UserModel` - when we use `withIdentities()` method the query is produced properly. 

The problem occurs in the `assignIdentities()` method, when we're trying to assign identities to the users.
We're working with the `User` entity, and there is a line: 
```
$newIdentities   = $mappedUsers[$userId]->identities;
```

It will call `getIdentities()` method which will produce a query to the DB, since there are no identities assigned yet.

This PR solves this problem.

### DIFF
--- a/tests/Unit/UserTest.php
+++ b/tests/Unit/UserTest.php
@@ -106,7 +106,11 @@ final class UserTest extends DatabaseTestCase
 
             // Check the last query and see if a proper type of query was used
             $query = (string) model(UserModel::class)->getLastQuery();
-            $this->assertMatchesRegularExpression('/WHERE\s+.*\s+IN\s+\([^)]+\)/i', $query, 'Identities were not obtained with the single query (missing "WHERE ... IN" condition)');
+            $this->assertMatchesRegularExpression(
+                '/WHERE\s+.*\s+IN\s+\([^)]+\)/i',
+                $query,
+                'Identities were not obtained with the single query (missing "WHERE ... IN" condition)'
+            );
         }
 
         $this->assertCount(2, $identities);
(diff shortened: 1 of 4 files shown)


datamweb (approved):
@michalsn thank you!

tests/Unit/UserTest.php
```diff
@@ -87,6 +87,35 @@ public function testModelFindByIdWithIdentities(): void
...
 
+    public function testModelFindAllWithIdentitiesWhereInQuery(): void
+    {
+        fake(UserIdentityModel::class, ['user_id' => $this->user->id, 'type' => 'password']);
+        fake(UserIdentityModel::class, ['user_id' => $this->user->id, 'type' => 'access_token']);
+
+        // Grab the user again, using the model's identity helper
+        $users = model(UserModel::class)->withIdentities()->findAll();
+
+        $identities = 0;
```
kenjis, line 98:
```suggestion
        $identities = [];
```

kenjis:
Am I correct in assuming that this PR prevents queries to retrieve unnecessary Identities from being executed?
It's just useless and there has never been a problem with the results returned.

kenjis:
The code looks good.

michalsn:
> Am I correct in assuming that this PR prevents queries to retrieve unnecessary Identities from being executed?
It's just useless and there has never been a problem with the results returned.

There was never a problem with the correct result per se. The problem here was that we were producing n+1 queries for code like this:

```php
$users = model(UserModel::class)->withIdentities()->findAll();
foreach ($users as $user) {
    $user->email;
}
```
That would be just fine, but since we use `withIdentities()` it should not work that way.

kenjis (approved):

datamweb:
Thanks All!

PR #811, opened 2023-09-06
chore: use workflows in codeigniter4/.github

kenjis:
See https://github.com/codeigniter4/CodeIgniter4/issues/7819

- `phpcpd.yml` is not updated. See #813
- update rector to 0.18.2


datamweb (approved):
Honestly, I always have a problem with one repo being dependent on another. However, I understand the benefits of this work for the team. You are trying to eliminate duplicate tasks.

**codeigniter4/.github** It seems to be public, and I assume so any repo can use this method even if it is not under organization?

composer.json
```diff
@@ -26,7 +26,8 @@
         "codeigniter4/framework": "^4.2.7",
         "mikey179/vfsstream": "^1.6.7",
         "mockery/mockery": "^1.0",
-        "firebase/php-jwt": "^6.4"
+        "firebase/php-jwt": "^6.4",
+        "rector/rector": "0.18.1"
```
datamweb, line 30:
```suggestion
        "rector/rector": "0.18.2"
```

kenjis:
> codeigniter4/.github It seems to be public, and I assume so any repo can use this method even if it is not under organization?

Yes.
> The called workflow is stored in a public repository, and your organization allows you to use public reusable workflows.
https://docs.github.com/en/actions/using-workflows/reusing-workflows#access-to-reusable-workflows

MGatner (approved):
Even though I think using Rector as a non-dependency is technically better I think this ends up being cleaner and safer. I will move my own work in the same direction.

kenjis:
@MGatner If we don't have rector, devs cannot run rector.
 And PhpStorm always complains about "Undefined class" in `rector.php`.

`rector.php` is closely related to the version of rector; we cannot specify the version of rector without adding rector to `composer.json`.

Have you experienced any dependency conflicts with installing rector?

MGatner:
Nope! I bought into Bergmann's strong anti-dependency sentiment for static analysis when I first created this. I'm over it 😇 This is a good change.

PR #827, opened 2023-09-12
chore: fix workflow on section

kenjis:
**Description**
Follow-up #811
See https://github.com/codeigniter4/.github/issues/14

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [] Conforms to style guide


datamweb (approved):
It seems that it was done in this way in the example below.
https://docs.github.com/en/actions/using-workflows/reusing-workflows#example-caller-workflow

MGatner (approved):

PR #795, opened 2023-08-23
feat: HMAC SHA256 Authentication

tswagger:
Adding HMAC-SHA256 as an authenticator.  This method has a slight security advantage to a standard token authentication by signing the request with a shared secret.

Usage and coding mirrors closely the established Access Token Authentication classes and methods.

References:
- https://datatracker.ietf.org/doc/html/rfc4868
- https://www.techtarget.com/searchsecurity/definition/Hash-based-Message-Authentication-Code-HMAC

PR #844, opened 2023-09-19
fix: [HMAC Auth] wrong header key

kenjis:
**Description**
Follow-up #795

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


kenjis:
@tswagger Review, please.

tswagger (approved):
That is a good catch.  Looks good to me.

PR #838, opened 2023-09-16
Added revokeAccessTokenBySecret

WaaaghNL:
**Description**
After losing some tokens while i developed my app and i didnt want to remove all of the tokens i changed the `revokeAccessToken` to support removeing them by the secret that you can get with the `AccessTokens()` function.

**Checklist:**
- [x] Securely signed commits
- [x] Component(s) with PHPDoc blocks, only if necessary or adds value
- [x] Unit testing, with >80% coverage
- [x] User guide updated
- [x] Conforms to style guide

PR #858, opened 2023-09-26
feat: added `revokeAccessTokenBySecret()`

WaaaghNL:
**Description**
After losing some tokens while i developed my app and i didnt want to remove all of the tokens i changed the `revokeAccessToken` to support removeing them by the secret that you can get with the `AccessTokens()` function.

superseded: #838 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/docs/guides/api_tokens.md
+++ b/docs/guides/api_tokens.md
@@ -47,10 +47,11 @@ if ($user->tokenCan('users-read')) {
 
 ### Revoking Tokens
 
-Tokens can be revoked by deleting them from the database with the `revokeAccessToken($rawToken)` or `revokeAllAccessTokens()` methods.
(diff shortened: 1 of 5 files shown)


datamweb (approved):

kenjis:
@WaaaghNL It's a little late for that, but why do you need to use secret instead of id?

WaaaghNL:
> @WaaaghNL It's a little late for that, but why do you need to use secret instead of id?

I used the secret to keep it inline with the function revokeAccessToken($rawToken) where you insert the raw token.

kenjis (approved):

kenjis:
@WaaaghNL Thank you!

PR #862, opened 2023-09-28
docs: add "How to Build Shield Documentation"

kenjis:
**Description**
- add "How to Build Shield Documentation"

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [] Conforms to style guide

### DIFF
--- a/admin/how_to_build_docs.md
+++ b/admin/how_to_build_docs.md
@@ -1,0 +1,38 @@
+# How to Build Shield Documentation
+
+We use Material for MkDocs for our documentation.
+See https://squidfunk.github.io/mkdocs-material/getting-started/.
+
+## Requirements
+
+- Python3
+- pip
+
+## Installation
+
+```console
+pip3 install mkdocs
+pip3 install mkdocs-material
+pip3 install mkdocs-git-revision-date-localized-plugin
+```
+
+## Build the Docs
+
+```consolse
+mkdocs build
+```
+
+## See the Docs
+
+```consolse
+mkdocs serve
+```
+
+## Deploy Manually
+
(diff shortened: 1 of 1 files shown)


admin/HowToBuildDocs.md
```diff
@@ -0,0 +1,26 @@
+# How to Build Shield Documentation
```
datamweb, line 1: We use Material for MkDocs for implementation. It is better to add the description in this section that includes the line as follows.
https://squidfunk.github.io/mkdocs-material/getting-started/

admin/HowToBuildDocs.md
```diff
@@ -0,0 +1,26 @@
...
+
+```consolse
+mkdocs build
+```
+
+## See the Docs
+
+```consolse
+mkdocs serve
+```
```
datamweb, line 26: Command ` mkdocs gh-deploy` is also good for manual implementation if there is a problem. Of course, I think it needs administrator access.

admin/HowToBuildDocs.md
datamweb, line 1: We have previously named the files as follows.
**how_to_build_docs.md**

admin/HowToBuildDocs.md
kenjis: Added.

admin/HowToBuildDocs.md
kenjis: Added.

admin/HowToBuildDocs.md
kenjis: Done.

datamweb (approved):
Thank you! 🥀

PR #861, opened 2023-09-27
refactor: validation rules

kenjis:
**Description**
- put validation rules into the `ValidationRules` class
- make password rules reusable

**Checklist:**
- [x] Securely signed commits
- [ ] Component(s) with PHPDoc blocks, only if necessary or adds value
- [ ] Unit testing, with >80% coverage
- [x] User guide updated
- [x] Conforms to style guide

PR #868, opened 2023-09-30
fix: Change the visibility of validation properties for easier inheritance

sammyskills:
**Description**
Currently, the `Validation` class declares the properties `$config` and `$tables` as private which makes them impossible to use within child classes. This PR fixes that.

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/src/Validation/ValidationRules.php
+++ b/src/Validation/ValidationRules.php
@@ -9,12 +9,12 @@ use CodeIgniter\Shield\Config\Auth;
 
 class ValidationRules
 {
-    private Auth $config;
+    protected Auth $config;
 
     /**
      * Auth Table names
      */
-    private array $tables;
+    protected array $tables;
 
     public function __construct()
     {


datamweb:
Hey @sammyskills,

I can understand why you want to inherit from Validation?
As you know, you can use [customization/validation_rules](https://datamweb.github.io/shield/customization/validation_rules/).
Can you provide an example for better understanding?

sammyskills:
Hi @datamweb,

Yes, before the refactor of the validation rules from PR #861, one could easily customize the rules. But since the rules have now been placed in a class of its own, devs might want to use(reuse) some of the methods in the validation class, to create custom rules _(as another way to create or use rules)_.

Take for instance, I have an application that uses login via phone number and password. I can create a validation class in my **app/Validations** directory, named `ValidationRules`. Within this class, I can create a method: `getLoginByPhoneNumberRules()`, which also utilizes the `getPasswordRules()` method from the parent class, like so:

```php
// app/Validation/ValidationRules.php
use CodeIgniter\Shield\Validation\ValidationRules as ShieldValidationRules;

class ValidationRules extends ShieldValidationRules
{
    public function getLoginByPhoneNumberRules(): array
    {
        return [
            'phone' => 'required|numeric',
            'password' => $this->getPasswordRules()
        ]
    }
}
```
With the above, the `getValidationRules()` method of my custom LoginController _(which of course extends shield's LoginControler can return my own custom rule: `getLoginByPhoneNumberRules()`.

I hope you get what I mean.

datamweb (approved):
Thanks for the explanation,
when I reviewed the #861 I was aware that there would be no limit to customization.
Now I wanted to make sure that there was no limit.

Well, I understand with your explanation, you know that you can create and use your own methods and rules in file **app/Config/Validation.php**. However, this PR helps the integrity of the code to a great extent.
I think this PR is acceptable.

sammyskills:
> you know that you can create and use your own methods and rules in file **app/Config/Validation.php**

Yes, I know, and that is what I've been using for a while now. But the PR #861 provides yet **another** way to set validation rules, for devs who prefer not to use config files.

kenjis (approved):

kenjis:
I'm fine this change if someone wants to extend the class.

PR #869, opened 2023-09-30
fix: redirect inactive account to auth action page

sammyskills:
**Description**
The `isActivated()` check works for when a register action has been set in the auth config. Currently, if an inactive account tries to log in, the system redirects the user to the login page with the error message: 
```console
You must activate your account before logging in.
```
This begs the question: _"How do I activate my account?"_

This PR fixes this issue by redirecting the user to the auth action page, where a new token is generated and sent to the user's email.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/src/Filters/SessionAuth.php
+++ b/src/Filters/SessionAuth.php
@@ -61,10 +61,12 @@ class SessionAuth implements FilterInterface
             }
 
             if ($user !== null && ! $user->isActivated()) {
-                $authenticator->logout();
-
-                return redirect()->route('login')
-                    ->with('error', lang('Auth.activationBlocked'));
+                // If an action has been defined for register, start it up.
+                $hasAction = $authenticator->startUpAction('register', $user);
+                if ($hasAction) {
+                    return redirect()->route('auth-action-show')
+                        ->with('error', lang('Auth.activationBlocked'));
+                }
             }
 
             return;
--- a/tests/Authentication/Filters/AbstractFilterTestCase.php
+++ b/tests/Authentication/Filters/AbstractFilterTestCase.php
@@ -69,6 +69,7 @@ abstract class AbstractFilterTestCase extends TestCase
             echo 'Open';
         });
         $routes->get('login', 'AuthController::login', ['as' => 'login']);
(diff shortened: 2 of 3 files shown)


kenjis:
@sammyskills I cannot reproduce the situation.

If an inactive account tries to log in, the system redirects the user to the login page and redirect to `auth/a/show` and I see:
```
Email Activation

We just sent an email to you with a code to confirm your email address. Copy that code and paste it below.
```

sammyskills:
Hi @kenjis,

Yes, you will get the error message and it will be redirected to the `auth/a/show` page because the inactive account in your case, still has the `email_activate` identity.

Try this:
- Delete the `email_activate` identity associated with the user from the `auth_identities` table and try to login again 
OR
- Set the register action in the **app/Config/Auth.php** file to `null` temporarily
  - Register a new account
  - Set the register action back to `\CodeIgniter\Shield\Authentication\Actions\EmailActivator::class` in the **app/Config/Auth.php** file
  - Try to login with this account.

Live scenario:
- An application allowed some users to register without email verification
- But later activated email verification in the system.

kenjis:
@sammyskills "Delete the email_activate identity", this does not seem to happen.
Or it seems a bug in the application, not in Shield.
Why do you delete the identity?

The live scenario could happen. But we already set the user active:
https://github.com/codeigniter4/shield/blob/09a7cf3a95bdb5c99e1cb5f6d339f4df6142a07a/src/Controllers/RegisterController.php#L132-L133

sammyskills:
Ok, fine.

Let's look at it this way.

- You activate the register action
- Register an account
- Verify the token _(which activates the account and deletes the `email_activate` identity)_
- Now, deactivate this account `$user->deactivate()`.
- Try to login with this deactivated account.

If you get what I mean, how does a user with a deactivated account **"reactivate"** their account?

kenjis:
@sammyskills Okay, I got "You must activate your account before logging in."!

sammyskills:
Perfect!

Now, how does this user activate/reactivate their account?

kenjis (approved):
LGTM!

datamweb (approved):
@sammyskills thank you!

PR #870, opened 2023-10-01
docs: remove `Config\Security::$csrfProtection` config method by `.env`

kenjis:
**Description**
It is not recommended. Because it should be set to `session` in all environments.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [] Conforms to style guide


datamweb (approved):
👍

PR #872, opened 2023-10-02
refactor: remove Composer autoloading of email helper

kenjis:
**Description**
See https://forum.codeigniter.com/showthread.php?tid=88545

The Email helper is used in a limited number of places, and it would be inefficient to load it on every page.

Also, if a library autoloads helpers by Composer, there is no way to stop that loading. 
So I have changed my mind, it is better not using helper autoloading by Composer in libraries.
If autoloading is necessary, developers can set it up in `Config\Autoload`.

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -53,8 +53,7 @@
             "CodeIgniter\\Shield\\": "src"
         },
         "files": [
-            "src/Helpers/auth_helper.php",
-            "src/Helpers/email_helper.php"
+            "src/Helpers/auth_helper.php"
         ],
         "exclude-from-classmap": [
             "**/Database/Migrations/**"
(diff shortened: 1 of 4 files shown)


datamweb (approved):
two questions,
If we are going to be bound to CI 4.3, isn't it better to use on config/autoload.php for `auth_helper`?

Regarding filters, we load a lot of filters by default, what do you think?

kenjis:
> If we are going to be bound to CI 4.3, isn't it better to use on config/autoload.php for auth_helper?

Yes, I will send another PR.

kenjis:
> Regarding filters, we load a lot of filters by default, what do you think?

Probably no. Only the applied filters for the request URI are loaded.

kenjis:
I sent #893

PR #891, opened 2023-10-10
[Rector] Apply StringifyStrNeedlesRector

samsonasik:
**Description**

Enable `StringifyStrNeedlesRector` with set `@return array<int, string>` to `strip_explode()` method and run Rector to apply `ArraySpreadInsteadOfArrayMergeRector` as same type.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/src/Authentication/Passwords/PwnedValidator.php
+++ b/src/Authentication/Passwords/PwnedValidator.php
@@ -35,7 +35,7 @@ class PwnedValidator extends BaseValidator implements ValidatorInterface
     {
         $hashedPword = strtoupper(sha1($password));
         $rangeHash   = substr($hashedPword, 0, 5);
-        $searchHash  = substr($hashedPword, 5);
+        $searchHash  = (string) substr($hashedPword, 5);
 
         try {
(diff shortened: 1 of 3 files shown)


src/Authentication/Passwords/PwnedValidator.php
```diff
@@ -35,7 +35,8 @@ public function check(string $password, ?User $user = null): Result
     {
         $hashedPword = strtoupper(sha1($password));
         $rangeHash   = substr($hashedPword, 0, 5);
-        $searchHash  = substr($hashedPword, 5);
+        /** @var string $searchHash */
+        $searchHash = substr($hashedPword, 5);
```
samsonasik, line 39: on php 7.4, `substr()` can return `false`, that's why it uses `@var` here so phpstan now that it always return string

samsonasik:
Ready to merge 👍

kenjis (approved):
LGTM!

samsonasik:
@kenjis thank you for the review, let's merge 👍

PR #898, opened 2023-10-12
fix: phpstan errors for phpstan-codeigniter (v1.4.0.70400)

datamweb:
**Description**
Explain what you have changed, and why.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/phpstan-baseline.php
+++ b/phpstan-baseline.php
@@ -28,9 +28,19 @@ $ignoreErrors[] = [
 	'path' => __DIR__ . '/src/Authentication/Actions/Email2FA.php',
 ];
 $ignoreErrors[] = [
+	'message' => '#^Call to function model with CodeIgniter\\\\Shield\\\\Models\\\\UserIdentityModel\\:\\:class is discouraged\\.$#',
+	'count' => 2,
+	'path' => __DIR__ . '/src/Authentication/Actions/Email2FA.php',
+];
+$ignoreErrors[] = [
 	'message' => '#^Call to deprecated function random_string\\(\\)\\:
 The type \'basic\', \'md5\', and \'sha1\' are deprecated\\. They are not cryptographically secure\\.$#',
 	'count' => 1,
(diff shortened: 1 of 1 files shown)


src/Authentication/Actions/Email2FA.php
```diff
@@ -132,7 +132,7 @@ public function verify(IncomingRequest $request)
     public function createIdentity(User $user): string
     {
         /** @var UserIdentityModel $identityModel */
-        $identityModel = model(UserIdentityModel::class);
+        $identityModel = model('UserIdentityModel');
```
kenjis, line 135: At this time, we have not received any requests that this change is necessary. 
So I think it is fine to leave `::class` as it is.

Also, since we provide the config to replace the User Provider, it does not seem necessary for devs to replace other model from a module.

About Config files, #761 was created.

datamweb: So we need to lock the **codeigniter/phpstan-codeigniter** to the specific version.
 
https://github.com/codeigniter4/shield/blob/fd2dff93921c3789180cc17b9fb79b9d63aafc9b/composer.json#L32

kenjis: We can ignore the error.
See https://github.com/codeigniter4/CodeIgniter4/blob/52dbf70837526e9cdbd948248182006d17044f92/phpstan.neon.dist#L41-L42

datamweb:
@kenjis see here, phpunit /PHP 7.4 - OCI8 passed.🤒

kenjis:
Oh, yes! It probably fails because Oracle happens to take a long time to start up.
See https://github.com/codeigniter4/shield/issues/899#issuecomment-1760409681

kenjis (approved):

datamweb:
@kenjis thank you!

PR #908, opened 2023-10-16
perf: refactor Auth

kenjis:
**Description**
See https://github.com/codeigniter4/shield/discussions/905#discussioncomment-7285767

- to reduce the number of loaded files
   - in the current code, `service('auth')->routes($routes)` creates `Authentication` and `UserModel` instances.

Before:
Screenshot 2023-10-16 18 11 51

After:
Screenshot 2023-10-16 18 10 58

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/phpstan-baseline.php
+++ b/phpstan-baseline.php
@@ -2,16 +2,6 @@
 
 $ignoreErrors = [];
 $ignoreErrors[] = [
-	'message' => '#^Construct empty\\(\\) is not allowed\\. Use more strict comparison\\.$#',
-	'count' => 1,
-	'path' => __DIR__ . '/src/Auth.php',
-];
-$ignoreErrors[] = [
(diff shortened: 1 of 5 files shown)


datamweb:
@kenjis, we can use command `php spark sheild:setup` to set the value of `$collectors`. In this way, we allow the user to cancel the load of Auth collector in Debugbar even in the mode of `development`. Why not act like what we did for the helpers!

datamweb:
And another thing that got me thinking is if we can check if the file **APPPATH\Config\Auth.php** exists, if it does then the file **VENDORPATH\codeigniter4 \shield\ src\Auth.php** does not load
Is this reasonable?

kenjis:
Config\Auth extends Shield\Config\Auth. So both are always loaded.

datamweb (approved):
LGTM!
I fell in love with this PR :blush:.

PR #911, opened 2023-10-17
fix date localized-plugin and refactor action

datamweb:
**Description**
The production date by **mkdocs-git-revision-date-localized-plugin** on local is correct, this is due to local access to **.git**.
Now in the online mode, these dates are displayed incorrectly. This PR is trying to address this issue.
I already tested [here](https://github.com/datamweb/shield-oauth/pull/76).

- fix Last update&Created
- refactor action steps

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -9,11 +9,18 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-python@v4
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Setup Python
+        uses: actions/setup-python@v4
         with:
           python-version: 3.x
-      - run: pip3 install mkdocs-material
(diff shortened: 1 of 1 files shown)


kenjis:
@datamweb The shield-oauth  docs seems good, but why does this change fix the date issue?

datamweb:
@kenjis Yes, the shield-oauth docs date  problem was solved after the recent change.

The reason for solving the date problem is the following code snippet.
```yaml
uses: actions/checkout@v4
        with:
          fetch-depth: 0
```

Because it allows the plugin to access to the last commit that touched a specific file to be able to retrieve the date. 

More info see [here](https://github.com/timvink/mkdocs-git-revision-date-localized-plugin#note-when-using-build-environments) .

kenjis:
Oh, I got it!

> Only a single commit is fetched by default, for the ref/SHA that triggered the workflow. Set fetch-depth: 0 to fetch all history for all branches and tags.
https://github.com/actions/checkout

kenjis (approved):

PR #921, opened 2023-10-19
docs: add `logout` to session except list

### LINKED ISSUE #920: Bug: cannot login after logging out from url
### PHP Version

8.2.12

### CodeIgniter4 Version

4.4.1

### Shield Version

dev-develop

Windows, Linux

### Which server did you use?

apache

### Database

MariaDB

### Did you customize Shield?

No

### What happened?

Shield now attempts to redirect users to the page they were trying to access before logging in. However, if an unauthenticated user tries to directly access the logout URL (/logout route), it creates a login/logout loop, preventing them from logging in.

### Steps to Reproduce

- While not logged in, access /logout route. 
- Now try to logging, you will be redirected to logout page 

### Expected Output

I think Shield should ignore the logout route, i.e., not redirect the user to this route after login (to avoid this login/logout loop issue).

### Anything else?

_No response_

datamweb:
**Description**
The duty of **session** is to protect URLs. I think devs should know that there is no reason to protect **logout**. That's why I think we don't need to change the code.
I have already added this topic #642 .

fix #920 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/docs/customization/route_config.md
+++ b/docs/customization/route_config.md
@@ -57,7 +57,7 @@ update the paths for `except`:
 public $globals = [
     'before' => [
         // ...
-        'session' => ['except' => ['*/login*', '*/register', '*/auth/a/*', 'logout']],
+        'session' => ['except' => ['*/login*', '*/register', '*/auth/a/*', '*/logout']],
     ],
     // ...
 ];
--- a/docs/customization/route_config.md
+++ b/docs/customization/route_config.md
@@ -57,7 +57,7 @@ update the paths for `except`:
 public $globals = [
     'before' => [
         // ...
-        'session' => ['except' => ['*/login*', '*/register', '*/auth/a/*']],
+        'session' => ['except' => ['*/login*', '*/register', '*/auth/a/*', 'logout']],
     ],
     // ...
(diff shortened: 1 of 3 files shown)


datamweb:
Also, there is no reason for a user who is not logged in to have access to **/logout**.
This issue must be resolved.

```diff
/**
* Logs the current user out.
*/
public function logoutAction(): RedirectResponse
{
+// block route logout if see beforlogin
+if (!auth()->loggedIn()) {
+throw PageNotFoundException::forPageNotFound();
+}
// Capture logout redirect URL before auth logout,
// otherwise you cannot check the user in `logoutRedirect()`.
$url = config('Auth')->logoutRedirect();

auth()->logout();

return redirect()->to($url)->with('message', lang('Auth.successLogout'));
}
```

https://github.com/codeigniter4/shield/blob/50fb634742c2b4379c51855ae7175d77c34eeac3/src/Controllers/LoginController.php#L91-L100

kenjis (approved):

kenjis (approved):

PR #922, opened 2023-10-19
fix: show 404 page when see `/logout` if not login

datamweb:
**Description**
See https://github.com/codeigniter4/shield/pull/921#issuecomment-1771506477

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/src/Controllers/LoginController.php
+++ b/src/Controllers/LoginController.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace CodeIgniter\Shield\Controllers;
 
 use App\Controllers\BaseController;
+use CodeIgniter\Exceptions\PageNotFoundException;
 use CodeIgniter\HTTP\RedirectResponse;
 use CodeIgniter\Shield\Authentication\Authenticators\Session;
 use CodeIgniter\Shield\Traits\Viewable;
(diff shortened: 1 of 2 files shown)


sammyskills:
Hi @datamweb,
Thank you for this PR. But there are some things I'm looking at:
1. Why do we need to bother about unauthenticated users visiting the `/logout` page when the system automatically redirects them to the login page?
1. Why do we need to force a 404 page when the logout page actually exists?

datamweb:
@sammyskills Thanks for the feedback.

Is it reasonable to display message "You have successfully logged out." for a user who is not logged in?

sammyskills:
Okay, that is understandable. 

But instead of returning an error _(which it is not)_, how about returning to the `logoutRedirect()` url without the message?

sammyskills:
I'm glad you now understand my point. 

However, I think it will be an overkill to add a new string if we can always check if a user is logged in or not and show/not show the message, for example:
```php
// user logged in

// user not logged in
return redirect()->to($url);
```
Something along the lines of:
https://github.com/codeigniter4/shield/blob/5b50267c91e5126ab8900dff57c5284a3e42066d/src/Controllers/LoginController.php#L99

What do you think?

datamweb:
Hmm, good point. Send a PR.

sammyskills:
😊 @datamweb, thanks for the offer but I think your PR already has what is needed. Just a minor edit and we are good to go!

Please do the honours instead.

PR #924, opened 2023-10-22
phpstan: fix phpstan error

datamweb:
**Description**
see https://github.com/codeigniter4/shield/actions/runs/6600355638/job/17930257377

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/phpstan-baseline.php
+++ b/phpstan-baseline.php
@@ -436,9 +436,4 @@ $ignoreErrors[] = [
 	'count' => 2,
 	'path' => __DIR__ . '/src/Authorization/Traits/Authorizable.php',
 ];
-$ignoreErrors[] = [
-	'message' => '#^Call to function config with Config\\\\Email\\:\\:class is discouraged\\.$#',
-	'count' => 1,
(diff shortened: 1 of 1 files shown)


kenjis (approved):

datamweb:
@kenjis I suggest that we use **Schedule** in **GitHubAction**. This makes us realize faster that the tests have failed. Not that we have to wait for a PR to be established and find out.
Using  **Schedule** makes our code constantly tested, it makes the **contributors less involved**. 

The disadvantages of receiving notifications are many, of course, they can be changed according to the application settings (for example, run every two days, etc.)

More info [see ](https://medium.com/tradeling/how-to-schedule-jobs-using-github-actions-9f32667706ea)

kenjis:
In my opinion, since we are using GitHub for free, I would like to avoid increasing the load as much as possible.

While development is active, I don't think it will be much of a problem. This PR was also resolved fairly quickly.

However, yes, there is certainly a possibility that problems will be left unaddressed if a new PR is not forthcoming.

PR #934, opened 2023-10-29
lang: [ru] add Russian language

ivanBell199:
This adds Russian translation.


kenjis:
Thank you for trying.
But we don't need too many unrelated commits, and you missed signing.
Please include only commits for Russian translations.
Screenshot 2023-10-30 8 43 56

PR #942, opened 2023-11-07
Update User.php

hamransp:
if $this->name is not empty then update the name in the table auth_identities

example :
```php
   $randomString = random_string('alnum', 8);
        $dataUser = [
            'username' => 'tenri' . $randomString,
            'name' => 'tenri' . $randomString,
            'email' => 'tenri' . $randomString . '@gmail.com',
            'password_hash' => password_hash('password', PASSWORD_DEFAULT),
            'active' => true,
        ];

        $modelUser = new ModelsUser();
        $saveIdentity = new UserEntity($dataUser);
        $save = $modelUser->save($saveIdentity);
        dd($save, $saveIdentity);
```

### DIFF
--- a/src/Entities/User.php
+++ b/src/Entities/User.php
@@ -177,6 +177,10 @@ class User extends Entity
             $identity->secret = $this->email;
         }
 
+        if (!empty($this->name)) {
+            $identity->name = $this->name;
+        }
+
         if (! empty($this->password)) {
             $identity->secret2 = service('passwords')->hash($this->password);
         }


kenjis:
Please give a meaningful subject.

Also, the name in `auth_identities` is the name of that identity.
It is not used for email identity. Why do you need to update it?

hamransp:
Sometimes, If you use the registration form which allows you to add name and email then besides email, you also hope the name field will also be updated in the auth_identities table.

For now, the matter is like this
```
use App\Models\User as ModelsUser;
use CodeIgniter\Shield\Entities\User as UserEntity;
//
//
$dataUser = [
            'username' => 'tenri' . $randomString,
            'name' => 'tenri' . $randomString,
            'email' => 'tenri' . $randomString . '@gmail.com',
            'password_hash' => password_hash('password', PASSWORD_DEFAULT),
            'active' => true,
];
$modelUser = new ModelsUser();
$saveIdentity = new UserEntity($dataUser);
$save = $modelUser->save($saveIdentity);
```

then the name field will not be updated in the auth table auth_identities.

kenjis:
If you add `name` to User, you should add the `name` column in the `users` table.
If the `name` is the name for email identity, your code makes sense.
But it is not likely. The name is the name of a User.

datamweb:
@hamransp thanks for PR.

The reason mentioned is the **name**, not **first_name** or **last_name**. If you plan to create a **first_name** and **last_name**, read instruction [adding_attributes_to_users](https://github.com/codeigniter4/shield/blob/develop/docs/customization/adding_attributes_to_users.md) or refer to number #333.

In fact, you are the third person who had a wrong understanding of column **name**. Maybe it would have been better to choose another name for this column.

The next point is that Shield hashes the password by default, so you don't need to use `'password_hash' => password_hash('password', PASSWORD_DEFAULT),`
```diff
$dataUser = [
'email' => 'tenri' . $randomString . '@gmail.com',
-'password_hash' => password_hash('password', PASSWORD_DEFAULT),
+'password' => 'plain text password',
'active' => true,
];
```
I close this PR. If you think more explanation is needed, feel free to open it.

PR #943, opened 2023-11-07
Update UserModel.php

hamransp:
if $this->tempUser->name is not empty then update the name in the table auth_identities

example :
```php
$randomString = random_string('alnum', 8);
$dataUser = [
'username' => 'tenri' . $randomString,
'name' => 'tenri' . $randomString,
'email' => 'tenri' . $randomString . '@gmail.com',
'password_hash' => password_hash('password', PASSWORD_DEFAULT),
'active' => true,
];

$modelUser = new ModelsUser();
$saveIdentity = new UserEntity($dataUser);
$save = $modelUser->save($saveIdentity);
dd($save, $saveIdentity);
```

### DIFF
--- a/src/Models/UserModel.php
+++ b/src/Models/UserModel.php
@@ -342,6 +342,7 @@ class UserModel extends BaseModel
             // If you get identity (email/password), the User object must have the id.
             $this->tempUser->id = $user->id;
 
+            $user->name          = $this->tempUser->name ?? '';
             $user->email         = $this->tempUser->email ?? '';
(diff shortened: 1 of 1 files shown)


kenjis:
Please give a meaningful subject.

I don't understand why this is needed.

datamweb:
It seems that this user's understanding of column **name** is wrong. See [issue comment](https://github.com/codeigniter4/shield/pull/942#issuecomment-1798218511).

I'm closing this thread, if you think we didn't quite understand what was required, please provide further clarification.

PR #999, opened 2024-01-12
fix: `shield:user create` with default group

### LINKED ISSUE #998: Bug: spark cli does not assign a default group to newly created user
### PHP Version

8.2.14

### CodeIgniter4 Version

4.4.4

### Shield Version

dev

Windows, Linux

### Which server did you use?

apache

### Database

MariaDB 10.2

### Did you customize Shield?

No

### What happened?

While creating a new user from the cli, the `php spark shield:user create` command does not assign any group to the user (if not using the group option/parameter). 
I think the command should assign a default group to the newly created user by using the following setting: `config(\Config\AuthGroups::class)->defaultGroup` in order to avoid possible permissions issues at the first login attempt.

### Steps to Reproduce

```bash
php spark shield:user create

Username : bgeneto
Email : bgeneto@mail.com
Password : *******
Password confirmation : *******
User "bgeneto" created
```

### Expected Output

A corresponding line in the `auth_groups_users` table using the default group from `config(\Config\AuthGroups::class)->defaultGroup`.

### Anything else?

_No response_

datamweb:
**Description**
fix: #998 

The user registration form sets the default group value for the user when a new user is registered. Although I don't consider this a bug, I prefer that this process be done for command `shield:user create` as well.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
         "mockery/mockery": "^1.0",
         "phpstan/extension-installer": "^1.3",
         "phpstan/phpstan-strict-rules": "^1.5",
-        "rector/rector": "0.18.13"
+        "rector/rector": "0.19.0"
     },
     "provide": {
         "codeigniter4/authentication-implementation": "1.0"
--- a/phpstan-baseline.php
+++ b/phpstan-baseline.php
@@ -158,16 +158,7 @@ $ignoreErrors[] = [
 	'count' => 1,
 	'path' => __DIR__ . '/src/Collectors/Auth.php',
 ];
-$ignoreErrors[] = [
-	'message' => '#^PHPDoc type array\\<string, string\\> of property CodeIgniter\\\\Shield\\\\Commands\\\\Generators\\\\UserModelGenerator\\:\\:\\$arguments is not the same as PHPDoc type array of overridden property CodeIgniter\\\\CLI\\\\BaseCommand\\:\\:\\$arguments\\.$#',
-	'count' => 1,
-	'path' => __DIR__ . '/src/Commands/Generators/UserModelGenerator.php',
-];
-$ignoreErrors[] = [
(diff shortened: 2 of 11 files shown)


datamweb:
Two items.
I don't know why Rector has not been updated to the latest version by **Dependabot**. Maybe it's because of the changes in Devkit. @samsonasik please take a look if you can.

PRLintReloaded It requires configuration, which I do not have the necessary access to do, please do the necessary configuration.

lonnieezell:
In the future I'd rather see all of the unrelated fixes in a separate PR. I know Rector was being a pain though. 

While I like the idea, I think in the situations I'd be using it from the command line I'd rather be able to define which role during the creation process, instead of being forced into the default and then having to go to the database to change it. 

What do you think about reading the available roles from the config file and presenting that as a required choice to the user?

samsonasik:
@datamweb it probably related with dependabot config, 0.18.x probably only call 0.18.y, while 0.19.a need to be triggerable on specific config. CodeIgniter4 repo seems auto-update it

datamweb:
> What do you think about reading the available roles from the config file and presenting that as a required choice to the user?

Good idea, I'd like to come back to these PR at another time.

bgeneto:
> What do you think about reading

One can also make the `-g` argument available in the `create` action in order to set the user group otherwise use the default if not informed. This way we simplify the code and minimize the interactions while creating a new user via cli. What do you think @lonnieezell ?

datamweb:
> PRLintReloaded It requires configuration, which I do not have the necessary access to do, please do the necessary configuration.

I send PR #1001 for add Rules.

lonnieezell:
Sorry - forgot to reply here earlier, but PRLintReloaded is installed and should be working now.

datamweb:
> In the future I'd rather see all of the unrelated fixes in a separate PR

I sent separates PRs for fix #1002 , #1003  and #1004 .

kenjis:
Closed by #1162

PR #1012, opened 2024-02-02
lang: [cs] Add Czech Translation

MichalSkoula:
I have added Czech translation for shield ~+ fixed Slovak translation~.

PR #1015, opened 2024-02-03
lang: [cs] Add Czech Translation

MichalSkoula:
I have added a Czech translation for Shield. This update includes a signed commit and tests. Based on https://github.com/codeigniter4/shield/pull/1012


datamweb (approved):
Thanks!
Welcome to Shield Contributors.

PR #1019, opened 2024-02-08
docs: add `$mailType` with `html` value for email config

datamweb:
**Description**
We use format html to send emails, so it's better to have this in the documentation.
see https://forum.codeigniter.com/showthread.php?tid=89321

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/docs/getting_started/install.md
+++ b/docs/getting_started/install.md
@@ -124,6 +124,7 @@ your project.
     {
         public string $fromEmail  = 'your_mail@example.com';
         public string $fromName   = 'your name';
+        public string $mailType = 'html';
         // ...
     }
     ```


docs/getting_started/install.md
```diff
@@ -124,6 +124,7 @@ your project.
     {
         public string $fromEmail  = 'your_mail@example.com';
         public string $fromName   = 'your name';
+        public string $mailType = 'html';
```
kenjis, line 127: This is the Config file to the CI4 application.
It is not sure that the dev uses HTML for all emails that s/he sends.

But Shield uses HTML mails, so should Shield send the mails as `'html'` at the runtime?

kenjis: I sent #1020

Issue #608, opened 2023-01-23
DateTime hardcoded format `Y-m-d H:i:s` drives to "Error converting data type varchar to datetime"


mikeabbott10:
### PHP Version

8.1

### CodeIgniter4 Version

4.3

### Shield Version

1.0.0-beta3

### Which operating systems have you tested for this bug?

Windows

### Which server did you use?

apache

### Database

SQL Server 2019

### Did you customize Shield?

No

### What happened?

My db language is Italian and expects datetime format like 'd-m-Y H:i:s'. When you attempt to login, it tries to insert a row using the format 'Y-m-d H:i:s' by default and that's hardcoded at Shield\Models\LoginModel row 71.
https://github.com/codeigniter4/shield/blob/697f570277a4d2205852cacaae4837604b1c4b88/src/Models/LoginModel.php#L71

### Steps to Reproduce

Connect to a database with a datetime expected format not equal to 'Y-m-d H:i:s' and try to login

### Expected Output

The row gets insterted

### Anything else?

I think the format has to be customizable without the need to extend the LoginModel and override the whole recordLoginAttempt just for this.

kenjis:
Thank you for reporting.

Can SQL Server accept `YYYYY-MM-DD H:i:s` for the column? 
I believe this is standard notation.

mikeabbott10:
Yes, it can but my company wants it to not be modified. It's standard notation for some locales, not for others

mikeabbott10:
The title of this post should be edited. It's not about LoginModel only. It's about every datetime value insertion in db. The solution is to change the hardcoded 'Y-m-d H:i:s' string with a simple datetimeFormat constant somewhere

kenjis:
We thought `Y-m-d H:i:s` is safe to databases regardless of locale.
So once we added `Time::toDatabase()` but removed it before the release.
https://github.com/codeigniter4/CodeIgniter4/pull/6509
https://github.com/codeigniter4/CodeIgniter4/pull/6461

> Please tell me where these kind of constants should go so i can solve this and make a PR

I'm not sure, but
1. add `Time::toDatabase()` and make it configurable with Config file for it.
2. add in `CodeIgniter\Shield\Config\Auth`.

kenjis:
Yes, the datetime (and date only) format for database is related to data from and to database,
and all the CI4 applications. Not only for Shield.

jozefrebjak:
[...]

PR #1027, opened 2024-02-14
refactor: remove hard coded `'Y-m-d H:i:s'`

kenjis:
**Description**
See https://github.com/codeigniter4/shield/issues/608 and https://github.com/codeigniter4/CodeIgniter4/pull/8538

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -922,17 +922,17 @@ class Session implements AuthenticatorInterface
             $user,
             $selector,
             $this->hashValidator($validator),
-            $expires
+            $expires->format('Y-m-d H:i:s')
         );
 
         $this->setRememberMeCookie($rawToken);
     }
 
-    private function calcExpires(): string
+    private function calcExpires(): Time
     {
         $timestamp = Time::now()->getTimestamp() + setting('Auth.sessionConfig')['rememberLength'];
 
-        return Time::createFromTimestamp($timestamp)->format('Y-m-d H:i:s');
+        return Time::createFromTimestamp($timestamp);
     }
 
     private function setRememberMeCookie(string $rawToken): void
--- a/src/Models/LoginModel.php
+++ b/src/Models/LoginModel.php
@@ -40,7 +40,7 @@ class LoginModel extends BaseModel
         'identifier' => 'permit_empty|string',
         'user_agent' => 'permit_empty|string',
         'user_id'    => 'permit_empty|integer',
-        'date'       => 'required|valid_date',
+        'date'       => 'required',
     ];
     protected $validationMessages = [];
     protected $skipValidation     = false;
@@ -80,7 +80,7 @@ class LoginModel extends BaseModel
             'id_type'    => $idType,
             'identifier' => $identifier,
             'user_id'    => $userId,
-            'date'       => Time::now()->format('Y-m-d H:i:s'),
+            'date'       => Time::now(),
             'success'    => (int) $success,
         ]);
 
@@ -121,7 +121,7 @@ class LoginModel extends BaseModel
             'id_type'    => Session::ID_TYPE_EMAIL_PASSWORD,
             'identifier' => $faker->email(),
             'user_id'    => null,
-            'date'       => Time::parse('-1 day')->format('Y-m-d H:i:s'),
+            'date'       => Time::parse('-1 day'),
             'success'    => true,
         ]);
     }
--- a/src/Authentication/Authenticators/AccessTokens.php
+++ b/src/Authentication/Authenticators/AccessTokens.php
@@ -170,7 +170,7 @@ class AccessTokens implements AuthenticatorInterface
             ]);
         }
 
-        $token->last_used_at = Time::now()->format('Y-m-d H:i:s');
+        $token->last_used_at = Time::now();
 
         if ($token->hasChanged()) {
             $identityModel->save($token);
--- a/src/Authentication/Authenticators/HmacSha256.php
+++ b/src/Authentication/Authenticators/HmacSha256.php
@@ -186,7 +186,7 @@ class HmacSha256 implements AuthenticatorInterface
             ]);
         }
 
-        $token->last_used_at = Time::now()->format('Y-m-d H:i:s');
+        $token->last_used_at = Time::now();
 
         if ($token->hasChanged()) {
             $identityModel->save($token);
--- a/src/Authorization/Traits/Authorizable.php
+++ b/src/Authorization/Traits/Authorizable.php
@@ -390,7 +390,7 @@ trait Authorizable
                 $inserts[] = [
                     'user_id'    => $this->id,
                     $type        => $item,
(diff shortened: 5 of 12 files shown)


src/Models/LoginModel.php
```diff
@@ -40,7 +40,7 @@ class LoginModel extends BaseModel
         'identifier' => 'permit_empty|string',
         'user_agent' => 'permit_empty|string',
         'user_id'    => 'permit_empty|integer',
-        'date'       => 'required|valid_date',
+        'date'       => 'required',
```
kenjis, line 43: Do we need more validation rule?

kenjis:
I sent PR to fix: https://github.com/codeigniter4/CodeIgniter4/pull/8543

```
 ------ --------------------------------------------------------------------- 
  Line   src/Controllers/MagicLinkController.php                              
 ------ --------------------------------------------------------------------- 
  108    Parameter #1 $row of method CodeIgniter\Model::insert() expects      
         array<int|string, float|int|string|null>|object|null, array<string,  
         CodeIgniter\I18n\Time|int|string|null> given.                        
 ------ --------------------------------------------------------------------- 

 ------ --------------------------------------------------------------------- 
  Line   src/Models/LoginModel.php                                            
 ------ --------------------------------------------------------------------- 
  77     Parameter #1 $row of method CodeIgniter\Model::insert() expects      
         array<int|string, float|int|string|null>|object|null, array<string,  
         CodeIgniter\I18n\Time|int|string|null> given.                        
 ------ --------------------------------------------------------------------- 

 ------ --------------------------------------------------------------------- 
  Line   src/Models/RememberModel.php                                         
 ------ --------------------------------------------------------------------- 
  58     Parameter #1 $row of method CodeIgniter\Model::insert() expects      
         array<int|string, float|int|string|null>|object|null, array<string,  
         CodeIgniter\I18n\Time|int|string|null> given.                        
 ------ --------------------------------------------------------------------- 

 ------ --------------------------------------------------------------------- 
  Line   tests/Authentication/MagicLinkTest.php                               
 ------ --------------------------------------------------------------------- 
  113    Parameter #1 $row of method CodeIgniter\Model::insert() expects      
         array<int|string, float|int|string|null>|object|null, array<string,  
         CodeIgniter\I18n\Time|int|string|null> given.                        
  135    Parameter #1 $row of method CodeIgniter\Model::insert() expects      
         array<int|string, float|int|string|null>|object|null, array<string,  
         CodeIgniter\I18n\Time|int|string|null> given.                        
 ------ --------------------------------------------------------------------- 

 ------ --------------------------------------------------------------------- 
  Line   tests/Controllers/MagicLinkTest.php                                  
 ------ --------------------------------------------------------------------- 
  102    Parameter #1 $row of method CodeIgniter\Model::insert() expects      
         array<int|string, float|int|string|null>|object|null, array<string,  
         CodeIgniter\I18n\Time|int|string|null> given.                        
 ------ --------------------------------------------------------------------- 

Error:  [ERROR] Found 6 errors
```
https://github.com/codeigniter4/shield/actions/runs/7895251231/job/21547330117?pr=1027

src/Authentication/Authenticators/Session.php
```diff
@@ -922,17 +922,17 @@ protected function rememberUser(User $user): void
             $user,
             $selector,
             $this->hashValidator($validator),
-            $expires
+            $expires->format('Y-m-d H:i:s')
```
datamweb, line 925: why !? :monocle_face:

src/Models/LoginModel.php
datamweb: The value of **date** is not taken from the user. Therefore, I do not see the need `valid_date`.

src/Authentication/Authenticators/Session.php
datamweb: Ah, I saw @TODO.

datamweb (commented):
**dateFormat** has been added to CI version 4.5 In contrast, Shield supports version 4.3.5. What effect will these changes have on older versions?

src/Authentication/Authenticators/Session.php
kenjis: Yes, we must pass a string. Otherwise it will be a breaking change.

kenjis:
> **dateFormat** has been added to CI version 4.5 In contrast, Shield supports version 4.3.5. What effect will these changes have on older versions?

In my understanding, there is no changes.
The `Time` instance that is passed to a Model will be converted to a string with hard coded `Y-m-d H:i:s` in the Model.
See https://github.com/codeigniter4/CodeIgniter4/blob/3595944eac90df7e499403bad8767935c41488d9/system/BaseModel.php#L1316-L1320

src/Authentication/Authenticators/Session.php
```diff
@@ -922,17 +922,17 @@ protected function rememberUser(User $user): void
...
         $this->setRememberMeCookie($rawToken);
     }
 
-    private function calcExpires(): string
+    private function calcExpires(): Time
     {
         $timestamp = Time::now()->getTimestamp() + setting('Auth.sessionConfig')['rememberLength'];
 
-        return Time::createFromTimestamp($timestamp)->format('Y-m-d H:i:s');
+        return Time::createFromTimestamp($timestamp);
```
datamweb, line 935: @kenjis This issue is not related to this PR, but I realized it now.
It seems that method `createFromTimestamp()` is not affected by changing the value of `App::$appTimezone` .

For example, by changing the value to App::$appTimezone = 'Asia/Tehran', I would expect the `createFromTimestamp()` method to be affected, but **timezone** is still set to 'UCT', which doesn't seem right. look :
https://github.com/codeigniter4/CodeIgniter4/blob/7d50bd33056cd54aef7f1905c1304947409ef078/system/I18n/TimeTrait.php#L266-L272

To fix it, you must do the following:
```diff
+use Config/App;
-return Time::createFromTimestamp($timestamp);
+return Time::createFromTimestamp($timestamp, config(App::class)->appTimezone);
```
Or make changes in CI4 
https://github.com/codeigniter4/CodeIgniter4/blob/7d50bd33056cd54aef7f1905c1304947409ef078/system/I18n/TimeTrait.php#L266-L272

kenjis: The current behavior seems to be misleading.
It seems better to change `Time`.

Carbon sets the current timezone. 
> You can create instances from [unix timestamps](http://en.wikipedia.org/wiki/Unix_time). createFromTimestamp() create a Carbon instance equal to the given timestamp and will set the timezone as well or default it to the current timezone. 
https://carbon.nesbot.com/docs/#api-instantiation

kenjis: I sent a PR: https://github.com/codeigniter4/CodeIgniter4/pull/8544

kenjis: Fixed by v4.4.6.

datamweb (approved):

datamweb:
@kenjis thank you!

PR #1040, opened 2024-02-22
fix: `JWT::loggedIn()` does not remove `Bearer` prefix

kenjis:
**Description**
Supersedes #1039 , Close #1039
Fixes #1038 

**Checklist:**
- [x] Securely signed commits
- [x] Component(s) with PHPDoc blocks, only if necessary or adds value
- [ ] Unit testing, with >80% coverage
- [ ] User guide updated
- [x] Conforms to style guide

PR #1044, opened 2024-02-24
fix: LoginModel cannot use uuid for id column

MrFrost-Nv27:
**Description**
i want use uuid for user id, so let set the validation customable or change the behavior.

my opinion is for improve jwt token security auth at sub payload not integer id, but uuid

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide


kenjis:
Please remove the first commit. It will be fixed by #1040

PR #1046, opened 2024-02-24
fix: LoginModel cannot use uuid for id column

MrFrost-Nv27:
**Description**
Supersedes #1044

i want use uuid for user id, so let set the validation customable or change the behavior.

my opinion is for improve jwt token security auth at sub payload not integer id, but uuid

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/src/Models/LoginModel.php
+++ b/src/Models/LoginModel.php
@@ -39,7 +39,7 @@ class LoginModel extends BaseModel
         'id_type'    => 'required',
         'identifier' => 'permit_empty|string',
         'user_agent' => 'permit_empty|string',
-        'user_id'    => 'permit_empty|integer',
+        'user_id'    => 'permit_empty',
         'date'       => 'required|valid_date',
     ];
     protected $validationMessages = [];
--- a/src/Models/LoginModel.php
+++ b/src/Models/LoginModel.php
@@ -40,7 +40,7 @@ class LoginModel extends BaseModel
         'identifier' => 'permit_empty|string',
         'user_agent' => 'permit_empty|string',
         'user_id'    => 'permit_empty',
-        'date'       => 'required|valid_date',
+        'date'       => 'required',
     ];
     protected $validationMessages = [];
     protected $skipValidation     = false;


kenjis:
Thank you, but you missed GPG signing.

MrFrost-Nv27:
> Thank you, but you missed GPG signing. See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/workflow.md#gpg-signing-old-commits

Yes i forget, clear

datamweb:
@MrFrost-Nv27  Can you use uuid  instead of id only with this change? Have you done this practically?
As far as I remember, there were more changes needed to use uuid.

MrFrost-Nv27:
> @MrFrost-Nv27 Can you use uuid instead of id only with this change? Have you done this practically? As far as I remember, there were more changes needed to use uuid.

yes i do, LoginModel user_id get error only on jwt fail with user uuid, i delete integer and all clear, because the minimum scenario (i think) is only let validate the user id is not only integer, string so can be. But, The best way actually change the integer to uuid regex maybe, this the regex `regex_match[/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/]` see https://www.codeigniter.com/user_guide/incoming/routing.html#custom-placeholders and for implement it on validation see https://www.codeigniter.com/user_guide/libraries/validation.html#rules-for-general-use but I haven't tried it yet

ahh sorry, I just meant it. If your intention is to apply UUID to the entire project, including changing the ID from the user table to UUID, that only requires 2 additional steps, namely:
1. extending the migration file itself and changing all user id types to char(32) to prepare uuid
2. extending the UserModel and UserEntity

fyi i use the eloquent model hehe 😄 from beginning is start with eloquent service with CI4, this my code to add a user with credential and group
```php
UserModel::create([
            'username' => 'superadmin',
            'nama' => 'Super Admin',
        ])->setEmailIdentity([
                    'email' => 'superadmin@gmail.com',
                    'password' => "password",
                ])->addGroup('superadmin')->activate();
```

datamweb:
> But, The best way actually change the integer to uuid regex maybe

Honestly, I don't agree with the complete removal of the rule and I don't feel good. But it seems that using a regex is a good way. 

And your branch is currently conflicting, please rebase and fix conflicts.

src/Models/LoginModel.php
```diff
@@ -39,8 +39,8 @@ class LoginModel extends BaseModel
         'id_type'    => 'required',
         'identifier' => 'permit_empty|string',
         'user_agent' => 'permit_empty|string',
-        'user_id'    => 'permit_empty|integer',
-        'date'       => 'required',
+        'user_id'    => 'permit_empty',
+        'date'       => 'required|valid_date',
```
datamweb:
```suggestion
        'date'       => 'required',
```

src/Models/LoginModel.php
```diff
@@ -39,7 +39,7 @@ class LoginModel extends BaseModel
         'id_type'    => 'required',
         'identifier' => 'permit_empty|string',
         'user_agent' => 'permit_empty|string',
-        'user_id'    => 'permit_empty|integer',
+        'user_id'    => 'permit_empty',
```
kenjis, line 42: I think this is fine.
Since this model only logs login (attempt), it does not seem necessary to be very strict with validation.

Since user id can be letters instead of a number, `integer` is obviously wrong.

kenjis (approved):

src/Models/LoginModel.php
MrFrost-Nv27: Oukayyy, ahh i have a question bro, is JWT Auth Setting key has sync with CI Setting library ? maybe i want try the scenario of saving key on database

kenjis: No, at the moment. See the source code, it uses `config()` instead of `settings()`.

datamweb (approved):

PR #1068, opened 2024-03-17
Prep for 1.0.2 release

kenjis:
**Description**
Updates version references for 1.0.2.
Prev #1030

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/src/Auth.php
+++ b/src/Auth.php
@@ -41,7 +41,7 @@ class Auth
     /**
      * The current version of CodeIgniter Shield
      */
-    public const SHIELD_VERSION = '1.0.1';
+    public const SHIELD_VERSION = '1.0.2';
 
     protected AuthConfig $config;
     protected ?Authentication $authenticate = null;


src/Auth.php
```diff
@@ -41,7 +41,7 @@ class Auth
     /**
      * The current version of CodeIgniter Shield
      */
-    public const SHIELD_VERSION = '1.0.1';
+    public const SHIELD_VERSION = '1.0.2';
```
datamweb, line 44: Given a version number MAJOR.**MINOR**.PATCH, increment the:

MAJOR version when you make incompatible API changes
**MINOR version when you add functionality in a backward compatible manner**
PATCH version when you make backward compatible bug fixes
Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.

@Kenjis, isn't the addition of a new language translation(nl) one of the new features? I mean, shouldn't we increase the minor?

kenjis: I don't think so. Adding a new language translation does not provide an additional functionality.
No additional API.

Since it provides a new language translation, it can be considered a feature in the broadest sense, but there is no need to increase the minor version from a SemVer perspective.
Because after all, SemVer is about APIs. Adding new translations is just adding new data.

datamweb: Thanks for explaining it.

datamweb (approved):

PR #1106, opened 2024-04-24
lang: [de] Added missing translations in Auth.php

sba:

PR #1107, opened 2024-04-24
test: [de] Update GermanTranslationTest.php

sba:


datamweb:
@sba Can you do this commit in PR #1106 ?

PR #1113, opened 2024-05-03
feat: Change session manipulation methods from `private` to `protected`

michalsn:
**Description**
This PR changes the session manipulation methods from `private` to `protected`, so we can use them if we extend the `Authentication\Authenticators\Session` class.

Reference: https://github.com/codeigniter4/shield/discussions/1111

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -680,7 +680,7 @@ class Session implements AuthenticatorInterface
     /**
      * Gets User Info in Session
      */
-    private function getSessionUserInfo(): array
+    protected function getSessionUserInfo(): array
     {
         return session(setting('Auth.sessionConfig')['field']) ?? [];
     }
@@ -688,7 +688,7 @@ class Session implements AuthenticatorInterface
     /**
      * Removes User Info in Session
      */
-    private function removeSessionUserInfo(): void
+    protected function removeSessionUserInfo(): void
     {
         session()->remove(setting('Auth.sessionConfig')['field']);
     }
@@ -698,7 +698,7 @@ class Session implements AuthenticatorInterface
      *
      * @return int|string|null
      */
-    private function getSessionKey(string $key)
+    protected function getSessionKey(string $key)
     {
         $sessionUserInfo = $this->getSessionUserInfo();
 
@@ -710,7 +710,7 @@ class Session implements AuthenticatorInterface
      *
      * @param int|string|null $value
      */
-    private function setSessionKey(string $key, $value): void
+    protected function setSessionKey(string $key, $value): void
     {
         $sessionUserInfo       = $this->getSessionUserInfo();
         $sessionUserInfo[$key] = $value;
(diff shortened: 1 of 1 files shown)


datamweb (approved):

src/Authentication/Authenticators/Session.php
```diff
@@ -698,7 +698,7 @@ private function removeSessionUserInfo(): void
      *
      * @return int|string|null
      */
-    private function getSessionKey(string $key)
+    protected function getSessionKey(string $key)
```
kenjis: It seems the method name is not good.
This method gets the key value in Session User Info.
Better name?

src/Authentication/Authenticators/Session.php
```diff
@@ -710,7 +710,7 @@ private function getSessionKey(string $key)
      *
      * @param int|string|null $value
      */
-    private function setSessionKey(string $key, $value): void
+    protected function setSessionKey(string $key, $value): void
```
kenjis: It also seems the method name is not good.
This sets the key value in Session User Info.

src/Authentication/Authenticators/Session.php
```diff
@@ -720,7 +720,7 @@ private function setSessionKey(string $key, $value): void
     /**
      * Remove the key value in Session User Info
      */
-    private function removeSessionKey(string $key): void
+    protected function removeSessionKey(string $key): void
```
kenjis: It also seems the method name is not good.

kenjis:
This Session class is already very long, and would violate the SRP.
Don't we need a new class for these methods?

src/Authentication/Authenticators/Session.php
michalsn: Updated to `getSessionUserKey`.

src/Authentication/Authenticators/Session.php
michalsn: Updated to `setSessionUserKey`.

src/Authentication/Authenticators/Session.php
michalsn: Updated to `removeSessionUserKey`.

michalsn:
The methods related to the user session will give us about 50 lines of code. I don't see the point of moving them to a separate class - especially if they are to remain non-public.

kenjis (approved):

kenjis:
The session data manipulation is needed only for the Session authenticator.
So extracting them into a class does not seem to help much in the future.
If we extract classes, we should extract another functions, but so far there is no need.

By the way, is this an enhancement or a bug fix?
In other words, should the next release be minor version up?

michalsn:
I would say this is an enhancement.

kenjis:
Then it should be released as v1.1.0.

michalsn:
Okay. I'm not familiar with the Shield release cycle, so it's up to you when you'd like to merge this.

PR #1135, opened 2024-06-22
feat: add magic link force login config

warcooft:
**Description**

This feature provides config for developers so that the behavior when checking user sessions in the `startLogin` function can be easily controlled. So the user does not need to log out to assign a new user session. 

We also set the default value in that configuration to `false` to preserve the previous behavior

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/docs/references/magic_link_login.md
+++ b/docs/references/magic_link_login.md
@@ -23,6 +23,15 @@ in the **app/Config/Auth.php** file.
 public int $magicLinkLifetime = HOUR;
 ```
 
+### Magic Link Force Login
+
+By default, Magic Link will throw Logic exception if current user session is exist. If you want replace existing user session with new user session change it to `true`
+in the **app/Config/Auth.php** file.
+
+```php
+public bool $allowMagicLinkForceLogin = false;
+```
+
 ## Responding to Magic Link Logins
 
 !!! note
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -646,8 +646,8 @@ class Session implements AuthenticatorInterface
         /** @var int|string|null $userId */
         $userId = $this->getSessionUserKey('id');
 
-        // Check if already logged in.
-        if ($userId !== null) {
+        // Check if already logged in and magic link force login config.
+        if ($userId !== null && config('Auth')->allowMagicLinkForceLogin === false) {
             throw new LogicException(
                 'The user has User Info in Session, so already logged in or in pending login state.'
                     . ' If a logged in user logs in again with other account, the session data of the previous'
--- a/src/Config/Auth.php
+++ b/src/Config/Auth.php
@@ -180,6 +180,15 @@ class Auth extends BaseConfig
 
     /**
      * --------------------------------------------------------------------
+     * Force Magic Link Logins
+     * --------------------------------------------------------------------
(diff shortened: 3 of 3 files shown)


src/Authentication/Authenticators/Session.php
```diff
@@ -646,8 +646,8 @@ public function startLogin(User $user): void
         /** @var int|string|null $userId */
         $userId = $this->getSessionUserKey('id');
 
-        // Check if already logged in.
-        if ($userId !== null) {
+        // Check if already logged in and magic link force login config.
+        if ($userId !== null && config('Auth')->allowMagicLinkForceLogin === false) {
```
kenjis, line 650: This seems to invalidate the following security check.
So I think this is a security hole.

docs/references/magic_link_login.md
```diff
@@ -23,6 +23,15 @@ in the **app/Config/Auth.php** file.
 public int $magicLinkLifetime = HOUR;
 ```
 
+### Magic Link Force Login
+
+By default, Magic Link will throw Logic exception if current user session is exist. If you want replace existing user session with new user session change it to `true`
```
kenjis, line 28: LogicException means, "your code has a bug, so please fix it".
So I never accept a flag to disable LogicException.
This PR seems to be wrong implementation.

First of all, why a logged in user uses magic link login?

src/Authentication/Authenticators/Session.php
warcooft: should we delete the old session and then continue its state safely?

docs/references/magic_link_login.md
warcooft: I submit a magicLink request using the incognito browser, then the magicLink sends an email to my my gmail in main browser. when opened it will cause a LogicException because in my main browser I am already logged in.

warcooft: But this scenario rarely happens in real cases. this case often occurs when development involves switching accounts with specific roles.

kenjis:
> I submit a magicLink request using the incognito browser, then the magicLink sends an email to my my gmail in main browser. when opened it will cause a LogicException because in my main browser I am already logged in.

This could happen rarely.
In this case, the CodeIgniter\Shield\Exceptions\LogicException should not happen.
But I don't think you should be logged out automatically and logged in with the magic link.
Because this scenario is not a normal use case. 
Showing the message with "you are already logged in as ..." is better?

kenjis:
> this case often occurs when development involves switching accounts with specific roles.

Shield does not have a feature to switch accounts.
When you are logged in, if you navigate to `login` or `login/magic-link`, you will be redirected.

warcooft:
> > I submit a magicLink request using the incognito browser, then the magicLink sends an email to my my gmail in main browser. when opened it will cause a LogicException because in my main browser I am already logged in.
> 
> This could happen rarely. In this case, the CodeIgniter\Shield\Exceptions\LogicException should not happen. But I don't think you should be logged out automatically and logged in with the magic link. Because this scenario is not a normal use case. Showing the message with "you are already logged in as ..." is better?

I agree instead of displaying exception/whoops page.

kenjis (changes requested):
See https://github.com/codeigniter4/shield/pull/1135#issuecomment-2185750431

PR #1141, opened 2024-07-08
feat: new command shield:extend

warcooft:
**Description**
This PRs provide a new command `shield:extend` to easily extend any controller.

**The Command's Usage**
```
shield:extend
```

**The Command's Options**
```php
protected $options = [
        '-i' => 'The index of shield controllers to be extending in your app.',
        '-f' => 'Force overwrite ALL existing files in destination.',
    ];
```

Example:
```shell
this@PC playground % php spark shield:extend   

CodeIgniter v4.5.3 Command Line Tool - Server Time: 2024-07-09 05:32:31 UTC+07:00

List of the controller that will be extend:

  [1] LoginController
  [2] MagicLinkController
  [3] RegisterController

Please select one of these (1/2/3) : 3

  Created: APPPATH/Controllers/RegisterController.php
  After extending, don't forget to change the route. See https://shield.codeigniter.com/customization/route_config
``` 

Example when the controller already exists:

```shell
this@PC playground % php spark shield:extend        

CodeIgniter v4.5.3 Command Line Tool - Server Time: 2024-07-09 05:30:57 UTC+07:00

List of the controller that will be extend:

  [1] LoginController
  [2] MagicLinkController
  [3] RegisterController

Please select one of these (1/2/3) : 3
  File 'APPPATH/Controllers/RegisterController.php' already exists in destination. Overwrite? [n, y]: y

  Created: APPPATH/Controllers/RegisterController.php
  After extending, don't forget to change the route. See https://shield.codeigniter.com/customization/route_config
``` 

Or we can use the full option as below:

```shell
this@PC playground % php spark shield:extend -i 3 -f

CodeIgniter v4.5.3 Command Line Tool - Server Time: 2024-07-09 05:27:18 UTC+07:00


  Created: APPPATH/Controllers/RegisterController.php
  After extending, don't forget to change the route. See https://shield.codeigniter.com/customization/route_config
``` 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/src/Commands/Extend.php
+++ b/src/Commands/Extend.php
@@ -1,0 +1,198 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * This file is part of CodeIgniter Shield.
+ *
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view
+ * the LICENSE file that was distributed with this source code.
+ */
+
+namespace CodeIgniter\Shield\Commands;
+use CodeIgniter\CLI\CLI;
+use CodeIgniter\Shield\Commands\Setup\ContentReplacer;
+
+class Extend extends BaseCommand
+{
+    /**
+     * The Command's Name
+     *
+     * @var string
+     */
+    protected $name = 'shield:extend';
+
+    /**
+     * The Command's Description
+     *
+     * @var string
+     */
+    protected $description = 'Extending the Controllers.';
+
+    /**
+     * The Command's Usage
+     *
+     * @var string
+     */
+    protected $usage = 'shield:extend';
+
+    /**
+     * The Command's Arguments
+     *
+     * @var array
+     */
+    protected $arguments = [];
+
+    /**
+     * The Command's Options
+     *
+     * @var array
+     */
+    protected $options = [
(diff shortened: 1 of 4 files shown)


src/Commands/Extend.php
```diff
@@ -0,0 +1,198 @@
...
+
+    /**
+     * Actually execute a command.
+     *
+     * @param array $params
+     */
+    public function run(array $params)
+    {
+        $this->replacer = new ContentReplacer();
+        $this->sourcePath = __DIR__ . '/../../examples/';
```
kenjis, line 72: Why do you need to have examples?
I think we should avoid maintaining both the original controller and the controller in Examples.

kenjis:
Also, please fix errors in GitHub Action checks, and add test code.
Reference: https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md

src/Commands/Extend.php
warcooft: I created an example to provide a simple way to create an extended controller because in some cases we need to customize the controller. I did not copy all the methods from the source, only some parts that most likely need to be modified.

This is quite helpful for novice users like me, if you guys do not agree with this way, feel free to close this PRs.

PR #1146, opened 2024-07-17
feat: using a more precise return

warcooft:
**Description**

The previous behavior may look like there is no problem, but when the user submits the verification code it will automatically change the state of the URL to `http://localhost:8080/auth/a/verify`. 

When the codes don't match and they accidentally refresh the page it will cause a 404 error because the route `auth/a/verify` with the `GET` method does not exist.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/src/Authentication/Actions/EmailActivator.php
+++ b/src/Authentication/Actions/EmailActivator.php
@@ -118,9 +118,7 @@ class EmailActivator implements ActionInterface
 
         // No match - let them try again.
         if (! $authenticator->checkAction($identity, $postedToken)) {
-            session()->setFlashdata('error', lang('Auth.invalidActivateToken'));
-
-            return $this->view(setting('Auth.views')['action_email_activate_show']);
+            return redirect()->back()->with('error', lang('Auth.invalidActivateToken'));
         }
 
         $user = $authenticator->getUser();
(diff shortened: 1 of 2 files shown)


kenjis:
Please write test code to prove your code works fine in all PRs. 
See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#unit-testing

kenjis:
> [!IMPORTANT]
> We expect all code changes or bug-fixes to be accompanied by one or more tests
> added to our test suite to prove the code works.

This PR does not have the necessary test code. If you want this PR to be reviewed, please add the test code.
See https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/pull_request.md#unit-testing

If the test is not added, this PR will be closed.

kenjis:
The following test fails. Please fix your code. Thank you.
```
There was 1 failure:

1) Tests\Controllers\RegisterTest::testRegisterActionRedirectsIfTokenNotMatch
Do not see 'The code was incorrect.' in response.
Failed asserting that false is true.

/home/runner/work/shield/shield/vendor/codeigniter4/framework/system/Test/TestResponse.php:436
/home/runner/work/shield/shield/tests/Controllers/RegisterTest.php:360
phpvfscomposer:///home/runner/work/shield/shield/vendor/phpunit/phpunit/phpunit:97
```
https://github.com/codeigniter4/shield/actions/runs/9990325709/job/27721717998?pr=1146

PR #1153, opened 2024-08-07
test: fix incorrect condition in AbstractTranslationTestCase

kenjis:
**Description**
Fix the following error.
```
 ------ -----------------------------------------------------------------------
  Line   tests/Language/AbstractTranslationTestCase.php
 ------ -----------------------------------------------------------------------
  327    Strict comparison using === between array{list<non-empty-string>} and
         array{} will always evaluate to false.
         💡 Because the type is coming from a PHPDoc, you can turn off this
            check by setting treatPhpDocTypesAsCertain: false in your
            phpstan.neon.dist.
 ------ -----------------------------------------------------------------------
```
https://github.com/codeigniter4/shield/actions/runs/10269040515/job/28413552557?pr=1152

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/.github/workflows/phpunit-lang.yml
+++ b/.github/workflows/phpunit-lang.yml
@@ -9,6 +9,7 @@ on:
       - '!src/Language/en/**.php'
       - 'phpunit*'
       - '.github/workflows/phpunit-lang.yml'
+      - 'tests/Language/AbstractTranslationTestCase.php'
   push:
     branches:
       - develop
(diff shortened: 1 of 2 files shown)


paulbalandan (approved):

datamweb (changes requested):
It looks like we need to update the `paths` variable in phpunit-lang Github Actions by adding `tests/Language/AbstractTranslationTestCase.php` to see the changes applied.

datamweb (approved):

datamweb:
@kenjis thanks.

PR #1160, opened 2024-08-12
fix: `chain` filter does not update `last_active`

### LINKED ISSUE #1154: Bug: last_active not set when filter set to chain
### PHP Version

8.2.16

### CodeIgniter4 Version

4.5.4

### Shield Version

1.1.0

Windows

### Which server did you use?

apache

### Database

MariaDB 10.6.11

### Did you customize Shield?

Added Shield oAuth

### What happened?

When logging in, last_active is not set in the database when filter is configured `'chain' => ['except' => ['login*', 'auth/a/*', 'oauth/*', 'logout']],`. When configured `'session' => ['except' => ['login*', 'auth/a/*', 'oauth/*', 'logout']],` it works, but for my specific use case I need chain as I'm using both session and access token for different purposes.

### Steps to Reproduce

Configure **Filters.php** file with e.g.,

```php
public $globals = [
    'before' => [
        // ...
        'chain' => ['except' => ['login*', 'register', 'auth/a/*', 'logout']],
    ],
    // ...
];
```

### Expected Output

last_active to be set in the database.

### Anything else?

_No response_

kenjis:
**Description**
Fixes #1154

The `chain` filter also should update `last_active`.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/src/Config/Auth.php
+++ b/src/Config/Auth.php
@@ -160,7 +160,7 @@ class Auth extends BaseConfig
      * --------------------------------------------------------------------
      * If true, will always update the `last_active` datetime for the
      * logged-in user on every page request.
-     * This feature only works when session/tokens filter is active.
+     * This feature only works when session/tokens/hmac/chain/jwt filter is active.
      *
      * @see https://codeigniter4.github.io/shield/quick_start_guide/using_session_auth/#protecting-pages for set filters.
      */
--- a/src/Filters/ChainAuth.php
+++ b/src/Filters/ChainAuth.php
@@ -53,9 +53,17 @@ class ChainAuth implements FilterInterface
         $chain = config('Auth')->authenticationChain;
 
         foreach ($chain as $alias) {
-            if (auth($alias)->loggedIn()) {
+            $auth = auth($alias);
+
+            if ($auth->loggedIn()) {
                 // Make sure Auth uses this Authenticator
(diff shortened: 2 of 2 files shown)


warcooft (approved):

datamweb (changes requested):
**This feature only works when session/tokens filter is active.**

https://github.com/codeigniter4/shield/blob/0ead2a1b9113ebd462157a80a2500c4aed99cba4/src/Config/Auth.php#L157-L167

warcooft:
> @kenjis said: Yes,`chain` uses either `session` or `tokens` or `hmac` by default.
> 
> > chain
> > The filter will check authenticators in sequence to see if the user is logged in through either of authenticators, allowing a single API endpoint to work for both an SPA using session auth, and a mobile app using access tokens.
> > https://shield.codeigniter.com/references/controller_filters/?h=chain#provided-filters

kenjis:
@datamweb The option is used in HmacAuth, JWTAuth, SessionAuth, and TokenAuth filters.
Find `'Auth.recordActiveDate'` in the source  code.

Updated the comment for `$recordActiveDate`.

datamweb (approved):
@warcooft Thank you for your participation.
@kenjis Yes, the text should have been updated.

LGTM!

PR #1161, opened 2024-08-12
chore: update phpstan-baseline.php

kenjis:
**Description**
Run `vendor/bin/phpstan analyze --generate-baseline phpstan-baseline.php`.

**Checklist:**
- [x] Securely signed commits
- [] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/phpstan-baseline.php
+++ b/phpstan-baseline.php
@@ -2,366 +2,438 @@
 
 $ignoreErrors = [];
 $ignoreErrors[] = [
+	// identifier: function.deprecated
 	'message' => '#^Call to deprecated function random_string\\(\\)\\:
 The type \'basic\', \'md5\', and \'sha1\' are deprecated\\. They are not cryptographically secure\\.$#',
 	'count' => 1,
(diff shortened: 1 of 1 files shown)


datamweb (approved):
```
P:\MyGitHubWork\shield>php vendor/bin/phpstan analyze --generate-baseline phpstan-baseline.php
Note: Using configuration file P:\MyGitHubWork\shield\phpstan.neon.dist.
 196/196 [============================] 100%



 [OK] Baseline generated with 137 errors.
```

The fact is that because of the length of the `php vendor/bin/phpstan analyze --generate-baseline phpstan-baseline.php` I never use it.
I choose to use command `composer analyze`.

Composer may need to be updated.

kenjis:
The `--generate-baseline` option updates the baseline file.
We don't need to update it normally because we should not add lines in the baseline file.

In this time, I've got too many conflicts when git rebasing. So I've used it and found the new comment lines.

If you want a composer command to update the baseline, 
https://github.com/codeigniter4/CodeIgniter4/blob/f60d27702b6844826343619755340b66faff9383/composer.json#L124-L125

PR #1162, opened 2024-08-12
fix: `shield:user create` does not assign a default group

### LINKED ISSUE #998: Bug: spark cli does not assign a default group to newly created user

kenjis:
**Description**
Supersedes #999
Fixes #998

The user registration form sets the default group value for the user when a new user is registered.
But the command `shield:user create` does not set the default group. 
I think we should fix the inconsistency as a bug.

> What do you think about reading the available roles from the config file and presenting that as a required choice to the user? https://github.com/codeigniter4/shield/pull/999#issuecomment-1890845904

It would be nice, but nobody has implemented it. It can be implemented as enhancement later.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/src/Commands/User.php
+++ b/src/Commands/User.php
@@ -310,6 +310,12 @@ class User extends BaseCommand
             $userModel->save($user);
             $this->write('User "' . $username . '" created', 'green');
         }
+
+        // Add to default group
+        $user = $userModel->findById($userModel->getInsertID());
+        $userModel->addToDefaultGroup($user);
+
+        $this->write('The user is added to the default group.', 'green');
     }
 
     /**
--- a/tests/Commands/UserTest.php
+++ b/tests/Commands/UserTest.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace Tests\Commands;
 
 use CodeIgniter\Shield\Commands\User;
+use CodeIgniter\Shield\Config\AuthGroups;
 use CodeIgniter\Shield\Entities\User as UserEntity;
 use CodeIgniter\Shield\Models\UserModel;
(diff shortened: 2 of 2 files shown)


datamweb (approved):
lonnieezell say: **instead of being forced into the default and then having to go to the database to change it.**

In any case, I do not see any problem in the implementation.

kenjis:
Oh, I have not read the comment carefully. Thanks.

And @bgeneto suggested `-g` option instead of showing groups and the user interaction. https://github.com/codeigniter4/shield/pull/999#issuecomment-1895726347

datamweb:
> suggested `-g` option

I like this because of less user interactions.

kenjis:
I sent a PR to implement `-g` option:  #1164

PR #1165, opened 2024-08-13
lang: [nl] fix missing translation

christianberkman:
**Description**
Translated an untranslated key in `Language/NL`

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/src/Language/nl/Auth.php
+++ b/src/Language/nl/Auth.php
@@ -25,7 +25,7 @@ return [
     'badToken'              => 'Ongeldige toegangstoken.',
     'oldToken'              => 'Vervallen toegangstoken.',
     'noUserEntity'          => 'Gebruikersentiteit moet worden opgegeven voor wachtwoordvalidatie.',
-    'invalidEmail'          => '(To be translated) Unable to verify the email address matches the email "{0}".',
+    'invalidEmail'          => 'Het e-mailadres "{0}" kon niet geverifieerd worden.',
     'unableSendEmailToUser' => 'Sorry, er is een probleem opgetreden bij het verzenden van de e-mail. We konden geen e-mail versturen naar "{0}".',
     'throttled'             => 'Te veel inlogpogingen van dit IP adres. Probeer het over {0} seconden opnieuw.',
     'notEnoughPrivilege'    => 'U hebt niet de nodige rechten om de gewenste bewerking uit te voeren.',


src/Language/nl/Auth.php
```diff
@@ -25,7 +25,7 @@
     'badToken'              => 'Ongeldige toegangstoken.',
     'oldToken'              => 'Vervallen toegangstoken.',
     'noUserEntity'          => 'Gebruikersentiteit moet worden opgegeven voor wachtwoordvalidatie.',
-    'invalidEmail'          => '(To be translated) Unable to verify the email address matches the email "{0}".',
+    'invalidEmail'          => 'Het e-mailadres "{0}" kon niet geverifieerd worden.',
```
datamweb, line 28: @christianberkman How about this?
```suggestion
    'invalidEmail'          => 'Kan niet bevestigen dat het e-mailadres "{0}" overeenkomt met het opgegeven e-mailadres.',
```

christianberkman: To be honest I have not used this functionality and the message puzzles me a bit. What is the context for this message? Password recovery? Then your suggestion makes sense, altough I would then I would add 
"... het opgegeven e-mailadres voor deze gebruiker" which would be ".. matches the email for this user".

datamweb: This text is not used to recover the password, but to log the user in via the magic link.

The magic link allows the user to temporarily enter it and then set a new password.

In any case, I think that the translation of a subject should be as close as possible to the original subject. Nothing should be added or taken away from it. Of course, I understand that sometimes it is difficult to do such a thing in translation.

datamweb: @NotCoffee418 Can you help please?

NotCoffee418: I have a problem with the english version in the first place.

Literal translation:
Kan niet verifiëren dat het e-mailadres overeenkomt met het e-mailadres "{0}"


Preferably, I'd go with:
Kan niet verifiëren dat het e-mailadres overeenkomt met "{0}"

christianberkman: So are we checking if the email entered by the user to receive the magic link is the email stored in the table?

In that case would the English "E-mail {0} is not associated with this account" be more clear? I am still not sure when the user would see this message and how the error could be triggered.

NotCoffee418: I can't fire up the code right now to check it but i dont quite get why this error exists in the first place.  
The [code](https://github.com/codeigniter4/shield/blob/0aaa04b11edd086694fef8f81f0e4a7bc8437539/src/Authentication/Actions/Email2FA.php#L76) indicates it occurs when you're in pending login state, and it's about to email you with a code.
So after logging in, you'd still be prompted for your e-mail in order to get the 2FA code.
I don't quite see the point of it except a very minimal layer of security and a lot of users not being able to log in easily.
But again, I haven't fired up the code so, it's just what i gather from context.

If it must exist, i'd agree with the english version provided by @christianberkman and it would translate to:
Het e-mailadres "{0}" is niet gekoppeld aan deze account.

datamweb: @NotCoffee418 Thank you for your help. As you mentioned, this is a security issue and it should be there.

I also agree with message «E-mail {0} is not associated with this account.».

Let me see what other members decide.

kenjis: @christianberkman 
> In that case would the English "E-mail {0} is not associated with this account" be more clear? I am still not sure when the user would see this message and how the error could be triggered.

The meaning is, "the email address you passed is not stored in the database".
Screenshot 2024-08-16 11 35 15
Screenshot 2024-08-16 11 35 23

kenjis: "E-mail {0} is not associated with this account" seems to be a bit different.
What is the "this account"?

christianberkman: From your example:
* "The e-mail address" refers to the e-mail in the databse, but that is not revealed.
* {0} refers to "not.exist@example.jp"

I think most users would assume "The e-mail address" is what they have just entered in the form, so they could interpert the error message as "Unable to verify [not.exist@example.jp] matches the e-mail not.exist@example.jp". That is confusing. Hence I would suggest


"Unable to verify {0} is associated with any registered user"  (or user account, etc).

kenjis: The line was `Unable to verify the email address matches the email on record.` https://github.com/codeigniter4/shield/commit/50591b5dfb425b9cdb2120d2f58ebab94a57d7f1
So the following is correct?
- Unable to verify the email address "[not.exist@example.jp](mailto:not.exist@example.jp)" matches the email on record.

christianberkman: I would say it is correct. Could argue still for "any email" as we have more than one email on record.

kenjis: I sent PR #1175

PR #1172, opened 2024-08-18
feat: add `UserModel::getReturnType()` and RegisterController::getUserEntity() uses it 

christianberkman:
**Description**
See #1170
Refactor `RegisterController::getUserEntity` to use `UserModel::returnType` instead of the hard coded `new User()`. This allows for easy use of a custom User Entity without needing to extend the `RegisterController` and add the routes.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/docs/customization/user_provider.md
+++ b/docs/customization/user_provider.md
@@ -55,10 +55,13 @@ class UserModel extends ShieldUserModel
 }
 ```
 
-## Using a custom ReturnType
-If you have set a custom `ReturnType` in your custom `UserModel`, you may retrieve the `ReturnTypt` using `UserModel::getReturnType()` and easily create a new `UserEntity` using your custom `ReturnType`:
+## Using a Custom User Entity
+
+If you have set a custom `$returnType` in your custom `UserModel`, you may
+retrieve the return type using the `UserModel::getReturnType()` method and
+easily create a new User Entity using it:
 
 ```php
 $userEntityClass = $userModel->getReturnType();
-$newUser = new $userEntityClass();
+$newUser         = new $userEntityClass();
 ```
--- a/docs/customization/user_provider.md
+++ b/docs/customization/user_provider.md
@@ -54,3 +54,11 @@ class UserModel extends ShieldUserModel
     }
 }
 ```
+
+## Using a custom ReturnType
+If you have set a custom `ReturnType` in your custom `UserModel`, you may retrieve the `ReturnTypt` using `UserModel::getReturnType()` and easily create a new `UserEntity` using your custom `ReturnType`:
+
+```php
+$userEntityClass = $userModel->getReturnType();
+$newUser = new $userEntityClass();
+```
--- a/src/Controllers/RegisterController.php
+++ b/src/Controllers/RegisterController.php
@@ -19,7 +19,6 @@ use CodeIgniter\HTTP\RedirectResponse;
 use CodeIgniter\HTTP\RequestInterface;
 use CodeIgniter\HTTP\ResponseInterface;
 use CodeIgniter\Shield\Authentication\Authenticators\Session;
-use CodeIgniter\Shield\Entities\User;
 use CodeIgniter\Shield\Exceptions\ValidationException;
 use CodeIgniter\Shield\Models\UserModel;
 use CodeIgniter\Shield\Traits\Viewable;
@@ -163,7 +162,10 @@ class RegisterController extends BaseController
      */
     protected function getUserEntity(): User
     {
-        return new User();
+        $userProvider    = $this->getUserProvider();
+        $userEntityClass = $userProvider->getReturnType();
+
+        return new $userEntityClass();
     }
 
     /**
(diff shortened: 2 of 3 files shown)


src/Models/UserModel.php
```diff
@@ -395,4 +395,9 @@ private function checkReturnType(): void
             throw new LogicException('Return type must be a subclass of ' . User::class);
         }
     }
+
+    public function getUserEntity(): User
```
kenjis:
```suggestion
    /**
     * Returns User Entity Type
     *
     * @return class-string<User>
     */
    public function getReturnType(): string
```

src/Controllers/RegisterController.php
```diff
@@ -19,7 +19,6 @@
 use CodeIgniter\HTTP\RequestInterface;
 use CodeIgniter\HTTP\ResponseInterface;
 use CodeIgniter\Shield\Authentication\Authenticators\Session;
-use CodeIgniter\Shield\Entities\User;
```
kenjis: Revert this.

docs/customization/user_provider.md
```diff
@@ -54,3 +54,11 @@ class UserModel extends ShieldUserModel
     }
 }
 ```
+
+## Using a custom ReturnType
+If you have set a custom `ReturnType` in your custom `UserModel`, you may retrieve the `ReturnTypt` using `UserModel::getReturnType()` and easily create a new `UserEntity` using your custom `ReturnType`:
```
kenjis:
```suggestion
## Using a Custom User Entity

If you have set a custom `$returnType` in your custom `UserModel`, you may
retrieve the return type using the `UserModel::getReturnType()` method and
easily create a new User Entity using it:
```

docs/customization/user_provider.md
```diff
@@ -54,3 +54,11 @@ class UserModel extends ShieldUserModel
...
+
+```php
+$userEntityClass = $userModel->getReturnType();
+$newUser = new $userEntityClass();
```
kenjis:
```suggestion
$newUser         = new $userEntityClass();
```

src/Models/UserModel.php
kenjis: Is it better to add `createNewUser(array $data): User` or something instead of `getReturnType(): string`?

christianberkman: That would omit the UserEntity entirely so we would still not be able to use UserEntity during the RegistrationAction?

kenjis: If we add the following method to `UserModel`:
```php
    /**
     * Returns a new User Entity.
     *
     * @param array<string, mixed> $data The data passed to the constructor
     */
    public function createNewUser(array $data = []): User
    {
        return new $this->returnType($data);
    }
```

`RegisterController` would be like this:
```php
    protected function getUserEntity(): User
    {
        $userProvider = $this->getUserProvider();

        return $userProvider->createNewUser();
    }
```

kenjis: @christianberkman 
I think adding a method to create a  new user is better than a method that just returns the user type string.
Would you like to make such an update?

christianberkman: You you popose

```php
// UserModel
public function createNewUserEntity(): User
{
  return new $this->returnType;
}

// RegisterController
public function getUserEntity(): User
{
        $userProvider    = $this->getUserProvider();
        return $userProvider->createNewUserEntity();
}

Comment (kenjis) [reply src/Models/UserModel.php]: What do you mean by "popose"?

Comment (christianberkman) [reply src/Models/UserModel.php]: Sorry typo. I meant

So do you propose the following? (My last code snippet)

I also read your snippet with createNewUser wrong and actually we are on the same line indeed. 

Is it now better to open a fresh pull request since this differs from the original suggestion?

Comment (kenjis) [reply src/Models/UserModel.php]: Yes, it would be better to open a fresh PR.

And I propose to add the first parameter:
```php
    /**
     * Returns a new User Entity.
     *
     * @param array<string, mixed> $data (Optional) user data
     */
    public function createNewUser(array $data = []): User
    {
        return new $this->returnType($data);
    }
```
Then, a dev can create a complete new User entity.
If a dev want to use *Complete Constructor* to the custom User entity, s/he needs complete data to construct a new entity.
Also, the CI4 Entity permits an empty instance creation, so we can just call it without arguments.

datamweb: I personally don't like using `mixed`, and I prefer to explicitly define the possible types as much as we can. For any other unexpected values, we can wait for bug reports from users.

kenjis: If we don't use `mixed`, something like this:
`array<string, array|bool|int|float|null|object|string>`

But it contains `array`, and it is not acceptable. Therefore,
`array<string, array<array-key, array|bool|int|float|null|object|string>|bool|int|float|null|object|string>`

But still contains `array`... Give up and use `mixed`.
`array<string, array<array-key, mixed>|bool|int|float|null|object|string>`

Which is better?
- `@param array<string, mixed> $data (Optional) user data`
- `@param array<string, array<array-key, mixed>|bool|int|float|null|object|string> $data (Optional) user data`

or something else?

datamweb: My choice is the second option because it strikes a balance between overly vague type definitions and overwhelming complexity. While it's not ideal, it offers a reasonable level of flexibility and clarity.
```php
@param array<string, array<array-key, mixed>|bool|int|float|null|object|string> $data (Optional) user data
```

kenjis:
I sent #1196
Check, please.

PR #1178, opened 2024-08-22
refactor: [Authorizable Trait] move $matrix outside foreach loop

christianberkman:
**Description**
`$matrix = setting('AuthGroups.matrix');` can be moved out of the `foreach ($permissions as $permission)` loop since the value is the same for every iteration.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


datamweb (approved):
LGTM!

kenjis (approved):
LGTM!

Issue #1171, opened 2024-08-18
Bug: Error Occurred While Changing User Password: strtolower(): Argument #1 ($string) must be of type string, null given in NothingPersonalValidator


arashsaffari:
### PHP Version

8.3.10

### CodeIgniter4 Version

4.5.4

### Shield Version

1.1.0

### Which operating systems have you tested for this bug?

macOS

### Which server did you use?

apache

### Database

MySQL 8

### Did you customize Shield?

Yes, but I believe my changes have nothing to do with the error.

### What happened?

When changing a user's password, I encounter the following error:
```
strtolower(): Argument #1 ($string) must be of type string, null given
```

This error is caused by the NothingPersonalValidator.

### Steps to Reproduce

```php
if (! $this->validate($userValidationRules))
{
return redirect()->to('/users/update/' . $id)->withInput()->with('errors', $this->validator->getErrors());
}
```

### Expected Output

<img width="1239" alt="Bildschirmfoto 2024-08-18 um 03 22 27" src="https://github.com/user-attachments/assets/0de8bbcc-afee-43d6-acb9-14a9e7d702d8">

warcooft:
Can you show the complete code?

it looks like the email is missing or it could be that you create username only as your login credentials instead of using email.

datamweb:
Hey @arashsaffari ,

In Shield, all users are required to have a unique email address, which is mandatory during the registration process. However, it's possible to log in using other identifiers besides email, such as a username, employee ID, etc.

Considering this, it seems that the user you're trying to change the password for does not have an email address associated with their account.

Please verify that the user has an email by running the following command:

```console
php spark db:table auth_identities
```

Ensure that the user has a valid email (field **secret**).

If the user does have an email and the problem persists, please provide the minimal code necessary to reproduce the error.

arashsaffari:
I have an email address stored in the database. The problem is that when using `$user->email`, the `getEmail()` function in the User entity is not being called. However, when I call the `$user->getEmail()` function in the controller, `$user->email` is populated with the email address. Interestingly, when I work with `$user->password_hash`, the `getPasswordHash()` function in the entity is called, and I receive the hash. It's worth noting that I have the email stored in my `auth_identities` table, and my login credentials use instead of the email.

warcooft:
Can you show the code how you call the user data? if you call it correctly then the email displays well.

For example:
```php

$usersModel = auth()->getProvider();
//find user where id is "7"
$user = $usersModel->find(7);
[...]

PR #1183, opened 2024-08-26
fix: `strong_password` rule fails when the personal field contains an integer value.

warcooft:
**Description**

`strong_password` rule fails when the personal field contains an integer value. so, we must convert the personal field's data type to a string. See https://github.com/codeigniter4/shield/issues/1171#issuecomment-2309437552

Screenshot 2024-08-26 at 13 45 08

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/src/Authentication/Passwords/NothingPersonalValidator.php
+++ b/src/Authentication/Passwords/NothingPersonalValidator.php
@@ -95,7 +95,7 @@ class NothingPersonalValidator extends BaseValidator implements ValidatorInterfa
 
             foreach ($personalFields as $value) {
                 if (! empty($user->{$value})) {
-                    $needles[] = strtolower($user->{$value});
+                    $needles[] = strtolower((string) $user->{$value});
                 }
             }
 


datamweb:
@warcooft Thank you for submitting the PR. Could you please provide a step-by-step explanation of how to reproduce the issue?

I was unable to reproduce the issue for the value `employee_id`, which is an `number`.

```php
    public array $personalFields = [
        //'f_name',
        'employee_id'
    ];
```
```console
Data of Table "users":

+----+----------+--------+----------------+--------+-------------+--------------------+--------------------+------------+-------------+
| id | username | status | status_message | active | last_active | created_at         | updated_at         | deleted_at | employee_id |
+----+----------+--------+----------------+--------+-------------+--------------------+--------------------+------------+-------------+
| 6  | datamweb |        |                | 1      |             | 2024-08-26 20:4... | 2024-08-26 20:4... |            | 30303030    |
+----+----------+--------+----------------+--------+-------------+--------------------+--------------------+------------+-------------+
```

warcooft:
**Step to Reproduce**

make sure field `employee_id` set data type to `INT` or `TINYINT`
```php
$fields = [
    'employee_id' => ['type' => 'TINYINT', 'default' => 1],
];
``` 
Login first! then add this to your controller.
```php
  public function index()
  {
        //login first
        $data = [
            'password'      => '12345678Aa',
            'password_confirm' => '12345678Aa',
            'old_password' => 'P@ssw0rd' //change this with your password
        ];

        if (!$this->validateData($data, $this->getValidationRules())) {
            dd($this->validator->getErrors());
        }

        $result = auth()->check([
            'email'    => auth()->user()->email,
            'password' => $data['old_password'],
        ]);

        if (!$result->isOK()) {
            // wrong password
            dd('wrong password');
        }

        // Success!
        $users = auth()->getProvider();

        $user = auth()->user()->fill([
            'password' => $data['password']
        ]);

        $users->save($user);
}

 protected function getValidationRules(): array
 {
        return setting('Validation.changePassword') ?? [
            'password' => [
                'label' => 'Auth.password',
                'rules' => 'required|strong_password',
            ],
            'password_confirm' => [
                'label' => 'Auth.passwordConfirm',
                'rules' => 'required|matches[password]',
            ],
        ];
  }
``` 

the error seems to arise from the `strong_password` rule.

kenjis:
> [!IMPORTANT]
> We expect all code changes or bug-fixes to be accompanied by one or more tests added to our test suite to prove the code works.

First, please write test code that reproduces the error.
And add a commit to fix the error.
Then, the test should pass, and it proves the error is surely fixed.

PR #1184, opened 2024-08-26
fix: Update active date only for users activated and not banned.

warcooft:
**Description**
See https://github.com/codeigniter4/shield/discussions/1179#discussion-7084659

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


datamweb (approved):
LGTM! Thanks.

kenjis:
Because the existing PHPUnit tests do not fail, there seems no test code for `recordActiveDate`.
I would like you to add test code:
- does not update active date for a user that is not activated
- does not update active date for a user that is banned

PR #1186, opened 2024-08-27
refactor: remove unneeded Entities\Cast\IntBoolCast

grimpirate:
**Description**
While looking at the shield source code it was noted that the [IntBoolCast class is already provided by codeigniter](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Entity/Cast/IntBoolCast.php). Furthermore, the int_bool cast handler of the custom shield Entity [already provided as an int-bool in the framework's default Entity](https://github.com/codeigniter4/CodeIgniter4/blob/develop/system/Entity/Entity.php#L98). This commit cleans up redundant code.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


datamweb (commented):
I’m not entirely sure, but as far as I remember, these features were added to the framework after they were identified as necessary in Shield. Since Shield also supports older versions of the framework, we need to ensure that these changes do not break compatibility.

Could you please investigate which version of the framework these features were introduced in and make sure that the proposed changes will not cause issues with older versions?

kenjis:
Thank you for sending this PR.
Also, please fix all errors in GitHub Actions.

grimpirate:
You were correct regarding supporting older framework it appears datamweb: see [here](https://github.com/codeigniter4/CodeIgniter4/commit/076a5427b69060080f86d45e15b0f91ab152138e) and [here](https://github.com/codeigniter4/CodeIgniter4/commit/899187a79363596342c5710c761f980a35d373b6). Feature was implemented approximately two years ago.

kenjis:
@datamweb 

We have phpunit test "[phpunit / PHP 7.4 - MySQLi - lowest](https://github.com/codeigniter4/shield/actions/runs/10569805908/job/29283249126?pr=1186#logs)" that installs the lowest dependencies.
But it installs v4.4.7.

- Installing codeigniter4/framework (v4.4.7): Extracting archive

Ah, that is because v4.4.6 or below has one or more known vulnerabilities.
https://github.com/codeigniter4/CodeIgniter4/security/advisories

kenjis:
@grimpirate See the version tags.
Screenshot 2024-08-27 10 30 15
https://github.com/codeigniter4/CodeIgniter4/commit/899187a79363596342c5710c761f980a35d373b6

kenjis:
@grimpirate Why did you close this?
IntBoolCast has been introduced in v4.3.0, and Shield requires v4.3.5 or later.
See https://shield.codeigniter.com/getting_started/install/#requirements

grimpirate:
@kenjis sorry about that I thought based on the commit history that it was for backwards compatibility. I didn't think to compare the version numbers.

kenjis:
@grimpirate Okay, no problem. As you see, your refactoring is okay.
So if you send the PR again, we would review and merge.

PR #1192, opened 2024-08-31
feat: improve query performance for `find*()`

datamweb:
**Description**
This Pull Request introduces a database **migration** that adds an index to the **deleted_at** column in the **users table**. This change is intended to enhance the performance of queries that **filter based on the deleted_at** field, particularly for cases involving soft deletes.

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,5 +1,17 @@
 # Upgrade Guide
 
+## Version 1.1.0 to 1.1.1
+
+## Added Index to `deleted_at` Column
+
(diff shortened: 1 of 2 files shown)


kenjis:
This needs to be carefully considered.
See https://tech-torana-co-jp.translate.goog/entry/2023/11/27/123000?_x_tr_sl=ja&_x_tr_tl=en&_x_tr_hl=ja&_x_tr_pto=wapp

datamweb:
Hi Kenjis,
I did some research on this and it seems like it won’t have a positive impact. So, I’m going to close it. Thanks for the link you shared!"

PR #1196, opened 2024-09-08
feat: add ``UserModel::createNewUser()``  and `RegisterController` uses it

kenjis:
**Description**
Supersedes  #1172

- add `UserModel::createNewUser()`
- refactor

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/docs/customization/user_provider.md
+++ b/docs/customization/user_provider.md
@@ -55,13 +55,16 @@ class UserModel extends ShieldUserModel
 }
 ```
 
-## Using a Custom User Entity
+## Creating a Custom User Entity
 
-If you have set a custom `$returnType` in your custom `UserModel`, you may
-retrieve the return type using the `UserModel::getReturnType()` method and
-easily create a new User Entity using it:
+Starting from v1.2.0, `UserModel` in Shield has the `createNewUser()` method to
+create a new User Entity.
 
 ```php
-$userEntityClass = $userModel->getReturnType();
-$newUser         = new $userEntityClass();
+$user = $userModel->createNewUser($data);
 ```
+
+It takes an optional user data array as the first argument, and passes it to the
+constructor of the `$returnType` class.
+
+If your custom User entity cannot be instantiated in this way, override this method.
--- a/docs/customization/user_provider.md
+++ b/docs/customization/user_provider.md
@@ -55,10 +55,13 @@ class UserModel extends ShieldUserModel
 }
 ```
 
-## Using a custom ReturnType
-If you have set a custom `ReturnType` in your custom `UserModel`, you may retrieve the `ReturnTypt` using `UserModel::getReturnType()` and easily create a new `UserEntity` using your custom `ReturnType`:
+## Using a Custom User Entity
+
+If you have set a custom `$returnType` in your custom `UserModel`, you may
+retrieve the return type using the `UserModel::getReturnType()` method and
+easily create a new User Entity using it:
 
 ```php
 $userEntityClass = $userModel->getReturnType();
-$newUser = new $userEntityClass();
+$newUser         = new $userEntityClass();
 ```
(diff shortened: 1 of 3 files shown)


datamweb (approved):
LGTM!

najdanovicivan:
@kenjis will this work if returnType is 'array' ?

christianberkman (approved):

bgeneto:
Interesting the `auth()->user()` seems not to return a subclass of ` CodeIgniter\Shield\Entities\User` so we can't use a custom/inherited entity class (very useful when [Adding Attributes to Users](https://shield.codeigniter.com/customization/adding_attributes_to_users/)). 
But, interesting, something like `auth()->getProvider()->findById(\user_id())` can be used. Is it possible @kenjis  to normalize this kind of situation using returnType?

datamweb:
> Interesting the `auth()->user()` seems not to return a subclass of ` CodeIgniter\Shield\Entities\User` so we can't use a custom/inherited entity class (very useful when [Adding Attributes to Users](https://shield.codeigniter.com/customization/adding_attributes_to_users/)).

What do you mean exactly? Could you clarify?
The output for me is as follows:

image

bgeneto:
> 
> What do you mean exactly? Could you clarify? The output for me is as follows:
> 

Maybe I'm confused here or maybe I'm missing something... But I thought that:

`auth()->user()` was equivalent to `auth()->getProvider()->findById(user_id())`. 

But, in this case, when using the first approach, I'm getting a call to undefined method error: `auth()->user()->getName()`
I will provide a MWE below to make things clear and also to help other confused users like me :-) 

```php
<?php
// app/Config/Auth.php
public string $userProvider = \App\Models\UserModel::class;
```

My custom user model:

```php
<?php

declare(strict_types=1);

namespace App\Models;

use App\Entities\User;
use CodeIgniter\Shield\Models\UserModel as ShieldUserModel;

class UserModel extends ShieldUserModel
{
    protected function initialize(): void
    {
        parent::initialize();

        $this->allowedFields = [
            ...$this->allowedFields,
            'name', // Added
        ];
        $this->returnType = User::class;
    }
}
```

And finally our custom entity class:

```php
<?php

declare(strict_types=1);

namespace App\Entities;

use CodeIgniter\Shield\Entities\User as ShieldUser;

/**
 * @property string|null  $name
 */
class User extends ShieldUser
{
    private ?string $name = null;

    public function getName(): ?string {
        return $this->name;
    }

    public function setName(string $name): void {
        $this->name = $name;
    }
}
```
Now in any controller that extends the BaseController:

```php
<?php

// this works by calling method App\Entities\User::getName()
$name = auth()->getProvider()->findById(\user_id())->getName();

// this gives error: Call to undefined method CodeIgniter\Shield\Entities\User::getName()
$name = auth()->user()->getName();
```

My question/concern is whether the `getAuthenticator()->getUser()` method called in `auth()->user()` shouldn't return the derived entity `App\Entities\User` instead of `CodeIgniter\Shield\Entities\User` for the sake of library uniformity. Maybe I can use the new `createNewUser()` here also, but I don't think so...  

image

> PS1: Migration class for adding new fields to the user table omitted.
> PS2: I'm using CI 4.6.0 and Shield develop branch.

datamweb:
Please review the subject with the following changes:

```diff
<?php

declare(strict_types=1);

namespace App\Models;

use App\Entities\User;
use CodeIgniter\Shield\Models\UserModel as ShieldUserModel;

class UserModel extends ShieldUserModel
{
++protected $returnType = User::class;

    protected function initialize(): void
    {
        parent::initialize();

        $this->allowedFields = [
            ...$this->allowedFields,
            'name', // Added
        ];
--$this->returnType = User::class;
    }
}
```

bgeneto:
Wow... that was tricky! Thanks @datamweb for such a prompt reply. I'm always a bit confused when methods such as `initController()` and `initialize()` will be used and by whom :-(

PR #1217, opened 2024-11-02
lang: [tr] Language additions

YusufDurmaz:


src/Language/tr/Auth.php
```diff
@@ -25,7 +25,7 @@
     'badToken'              => 'Erişim anahtarı geçersiz.',
     'oldToken'              => 'Erişim anahtarının süresi doldu.',
     'noUserEntity'          => 'Parola doğrulaması için Kullanıcı Varlığı sağlanmalıdır.',
-    'invalidEmail'          => '(To be translated) Unable to verify the email address "{0}" matches the email on record.',
+    'invalidEmail'          => 'E-posta adresinin kayıtlı e-posta ile eşleştiği doğrulanamıyor.',
```
datamweb, line 28: Removing `"{0}"` has caused the test to fail. This placeholder is responsible for substituting the email value in the text, so removing it results in the email not being inserted, causing the test to fail.

It’s best to keep `"{0}"` to ensure the email value is properly replaced and the tests pass.

datamweb (changes requested):
It seems that you have created a folder named "shield," which is incorrect. Please apply your changes in path **src/Language/tr/Auth.php**, **NOT** in path **shield/src/Language/tr/Auth.php**.

PR #1219, opened 2024-11-03
feat: Add expiration date to access token & hmac keys

### LINKED ISSUE #926: Dev: set the expiration date of the token when creating the token
Currently, tokens expire based on the `Config/AuthToken::$unusedTokenLifetime` setting. It is better that users are able to set the expiration date of the token when creating the token(Access Token,HMAC SHA256 Token), like GitHub's token.

Sample:
Screenshot 2023-10-22 151933

CosDiabos:
**Description**
Implements custom date expiration for AccessTokens and Hmac Tokens leveraging the `expires` column within auth_identities table. 

Implements `setIdentityExpirationById()` model method, `updateAccessTokenExpiration()`, `removeAccessTokenExpiration()`, `isAccessTokenExpired()`, `hasAccessTokenExpiry()`, `updateHmacTokenExpiration()`, `removeHmacTokenExpiration()`, `isHmacTokenExpired()`, `hasHmacTokenExpiry()` trait methods and the CLI command `shield:hmac invalidateAll`.

Implements and closes #926 

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


src/Authentication/Traits/HasAccessTokens.php
```diff
@@ -34,15 +37,18 @@ trait HasAccessTokens
...
-     * @param string       $name   Token name
-     * @param list<string> $scopes Permissions the token grants
+     * @param string       $name      Token name
+     * @param list<string> $scopes    Permissions the token grants
+     * @param string       $expiresAt Sets token expiration date. Accepts DateTime string formatted as 'Y-m-d h:i:s' or DateTime relative formats (1 day, 2 weeks, 6 months, 1 year) to be added to DateTime 'now'
+     *
+     * @throws InvalidArgumentException
      */
-    public function generateAccessToken(string $name, array $scopes = ['*']): AccessToken
+    public function generateAccessToken(string $name, array $scopes = ['*'], ?string $expiresAt = null): AccessToken
```
michalsn: The type chosen for `$expiresAt` is quite strange to me - why `?string` and not `?Time`?
`Time` seems a lot more universal and we wouldn't need `checkExpiresAtFormat()` method at all.

src/Authentication/Traits/HasAccessTokens.php
```diff
@@ -165,4 +171,64 @@ public function setAccessToken(?AccessToken $accessToken): self
...
+     * Returns formatted date to expiration for provided AccessToken
+     *
+     * @param AccessToken $accessToken AccessToken
+     * @param string      $format      The return format - "date" or "human".  Date is 'Y-m-d h:i:s', human is 'in 2 weeks'
+     *
+     * @return string|null Returns a formatted expiration date or null if the expire field is not set.
+     *
+     * @throws InvalidArgumentException
+     */
+    public function getAccessTokenTimeToExpire(?AccessToken $accessToken, string $format = 'date'): string|null
```
michalsn: Relying directly on the `Time` object seems more universal to me. Let the developers decide what format they want to display the expiration date in. I would get rid of this method.

src/Authentication/Traits/HasAccessTokens.php
```diff
@@ -165,4 +171,64 @@ public function setAccessToken(?AccessToken $accessToken): self
 
         return $this;
     }
+
+    /**
+     * Checks if the provided Access Token has expired.
+     *
+     * @return bool|null Returns true if Access Token has expired, false if not, and null if the expire field is null
+     */
+    public function hasAccessTokenExpired(?AccessToken $accessToken): bool|null
```
michalsn: Can we return here only `true`/`false` - this seems more appropriate.

We can introduce another method, like `canAccessTokenExpire()` to determine whether the token has an expiration date.

michalsn (commented):
Sorry you had to wait so long for feedback.

The PR looks nice, although I have questions about the reasons for some of the "implementation" choices.

src/Authentication/Traits/HasAccessTokens.php
CosDiabos: You are right, a ```string``` doesn't make sense in this situation, a ```Time``` type is the correct answer, specially, as you mentioned gives people more flexibility. What I do think happen was, I got caught up with the date/time format and then, mistakenly went with strings without re-evaluating.

src/Authentication/Traits/HasAccessTokens.php
CosDiabos: Like I've mentioned in my previous comment, I agree with you and will change it accordingly. The idea with this method initially, was to return the "humanized" representation easily but then for the sake of completeness, I also brought in the date which now makes more sense getting both from the Time object directly.

src/Authentication/Traits/HasAccessTokens.php
CosDiabos: Yes, that makes more sense.

datamweb:
Is it possible to generate a token without a time limit?

CosDiabos:
> Is it possible to generate a token without a time limit?

Sure! Just skip that argument when generating a token. It will follow the defined token lifetime though.

docs/references/authentication/hmac.md
```diff
@@ -97,6 +97,59 @@ You can revoke all HMAC Keys with the `revokeAllHmacTokens()` method.
...
+// Or Expiration date = 1 month + 15 days into the future
+$expiresAt = Time::now();
+$expiresAt = $expiresAt->addMonths(1);
+$expiresAt = $expiresAt->addDays(15);
+$token = $user->setHmacTokenExpirationById($token->id, $expiresAt);
+```
+
+The following support methods are also available:
+
+`hasHmacTokenExpired(AccessToken $HmacToken)` - Checks if the HMAC key has expired. Returns `true` if the HMAC key has expired, `false` if not, and `null` if the expire date is null.
```
michalsn:
```suggestion
`hasHmacTokenExpired(AccessToken $HmacToken)` - Checks if the HMAC key has expired. Returns `true` if the HMAC key has expired, `false` if not.
```

docs/references/authentication/hmac.md
```diff
@@ -97,6 +97,59 @@ You can revoke all HMAC Keys with the `revokeAllHmacTokens()` method.
...
+By default, the HMAC keys don't expire unless they meet the HMAC Keys lifetime expiration after their last used date.
+
+HMAC keys can be set to expire through the `generateHmacToken()` method. This takes the expiration date as the $expiresAt argument. It's also possible to update an existing HMAC key using `setHmacTokenExpirationById($HmacTokenID, $expiresAt)`
+
+`$expiresAt` [Time](/libraries/time.html) object
+
+```php
+// Expiration date = 2024-11-03 12:00:00
+$expiresAt = Time::parse('2024-11-03 12:00:00');
+$token = $this->user->generateHmacToken('foo', ['foo:bar'], $expiresAt);
```
michalsn:
```suggestion
$token = $this->user->generateHmacToken('foo', ['foo.bar'], $expiresAt);
```
Scope names should avoid using a colon, according to the docs.

docs/references/authentication/hmac.md
```diff
@@ -97,6 +97,59 @@ You can revoke all HMAC Keys with the `revokeAllHmacTokens()` method.
...
+
+// Expiration date = 2024-11-15 00:00:00
+$expiresAt = Time::parse('2024-11-15 00:00:00');
+$token = $user->setHmacTokenExpirationById($token->id, $expiresAt);
+
+// Or Expiration date = 1 month + 15 days into the future
+$expiresAt = Time::now();
+$expiresAt = $expiresAt->addMonths(1);
+$expiresAt = $expiresAt->addDays(15);
```
michalsn:
```suggestion
$expiresAt = Time::now()->addMonths(1)->addDays(15);
```

docs/references/authentication/hmac.md
```diff
@@ -97,6 +97,59 @@ You can revoke all HMAC Keys with the `revokeAllHmacTokens()` method.
...
+$token = $user->setHmacTokenExpirationById($token->id, $expiresAt);
+```
+
+The following support methods are also available:
+
+`hasHmacTokenExpired(AccessToken $HmacToken)` - Checks if the HMAC key has expired. Returns `true` if the HMAC key has expired, `false` if not, and `null` if the expire date is null.
+
+```php
+$expiresAt = Time::parse('2024-11-03 12:00:00');
+$token = $this->user->generateHmacToken('foo', ['foo:bar'], $expiresAt);
```
michalsn:
```suggestion
$token = $this->user->generateHmacToken('foo', ['foo.bar'], $expiresAt);
```

docs/references/authentication/hmac.md
```diff
@@ -97,6 +97,59 @@ You can revoke all HMAC Keys with the `revokeAllHmacTokens()` method.
...
+
+$this->user->hasHmacTokenExpired($token); // Returns true
+```
+
+`canHmacTokenExpire(AccessToken $HmacToken)` - Checks if HMAC key has an expiration set. Returns `true` or `false` accordingly.
+
+```php
+$expiresAt = Time::parse('2024-11-03 12:00:00');
+
+$token = $this->user->generateHmacToken('foo', ['foo:bar'], $expiresAt);
```
michalsn:
```suggestion
$token = $this->user->generateHmacToken('foo', ['foo.bar'], $expiresAt);
```

docs/references/authentication/hmac.md
```diff
@@ -228,6 +281,9 @@ that CodeIgniter provides.
 public $unusedTokenLifetime = YEAR;
 ```
 
+### HMAC Keys Expiration vs Lifetime
+Expiration and Lifetime are different concepts. The lifetime is the maximum time allowed for the HMAC Key to exist since its last use. HMAC Key expiration, on the other hand, is a set date in which the HMAC Key will cease to function.
```
michalsn:
```suggestion
### HMAC Keys Expiration vs Lifetime

Expiration and Lifetime are different concepts. The lifetime is the maximum time allowed for the HMAC Key to exist since its last use. HMAC Key expiration, on the other hand, is a set date in which the HMAC Key will cease to function.
```
Just formatting - one line space between the heading and the content.

docs/references/authentication/tokens.md
```diff
@@ -137,6 +137,60 @@ that CodeIgniter provides.
...
+By default, the Access Tokens don't expire unless they meet the Access Token lifetime expiration after their last used date.
+
+Access Tokens can be set to expire through the `generateAccessToken()` method. This takes the expiration date as the $expiresAt argument. It's also possible to update an existing HMAC key using `setAccessTokenById($HmacTokenID, $expiresAt)`
+
+`$expiresAt` [Time](/libraries/time.html) object
+
+```php
+// Expiration date = 2024-11-03 12:00:00
+$expiresAt = Time::parse('2024-11-03 12:00:00');
+$token = $this->user->generateAccessToken('foo', ['foo:bar'], $expiresAt);
```
michalsn:
```suggestion
$token = $this->user->generateAccessToken('foo', ['foo.bar'], $expiresAt);
```

docs/references/authentication/tokens.md
```diff
@@ -137,6 +137,60 @@ that CodeIgniter provides.
...
+
+// Expiration date = 2024-11-15 00:00:00
+$expiresAt = Time::parse('2024-11-15 00:00:00');
+$user->setAccessTokenExpirationById($token->id, $expiresAt);
+
+// Or Expiration date = 1 month + 15 days into the future
+$expiresAt = Time::now();
+$expiresAt = $expiresAt->addMonths(1);
+$expiresAt = $expiresAt->addDays(15);
```
michalsn:
```suggestion
$expiresAt = Time::now()->addMonths(1)->addDays(15);
```

docs/references/authentication/tokens.md
```diff
@@ -137,6 +137,60 @@ that CodeIgniter provides.
...
+
+$user->setAccessTokenExpirationById($token->id, $expiresAt);
+```
+
+The following support methods are also available:
+
+`hasAccessTokenExpired(AccessToken $accessToken)` - Checks if Access Token has expired. Returns `true` if the Access Token has expired, `false` if not, and `null` if the expire date is not set.
```
michalsn:
```suggestion
`hasAccessTokenExpired(AccessToken $accessToken)` - Checks if Access Token has expired. Returns `true` if the Access Token has expired, `false` if not.
```

docs/references/authentication/tokens.md
```diff
@@ -137,6 +137,60 @@ that CodeIgniter provides.
...
+
+```php
+$expiresAt = Time::parse('2024-11-03 12:00:00');
+
+$token = $this->user->generateAccessToken('foo', ['foo:bar'], $expiresAt);
```
michalsn:
```suggestion
$token = $this->user->generateAccessToken('foo', ['foo.bar'], $expiresAt);
```

docs/references/authentication/tokens.md
```diff
@@ -137,6 +137,60 @@ that CodeIgniter provides.
...
+
+$this->user->hasAccessTokenExpired($token); // Returns true
+```
+
+`canAccessTokenExpire(AccessToken $HmacToken)` - Checks if Access Token has an expiration set. Returns `true` or `false` accordingly.
+
+```php
+$expiresAt = Time::parse('2024-11-03 12:00:00');
+
+$token = $this->user->generateAccessToken('foo', ['foo:bar'], $expiresAt);
```
michalsn:
```suggestion
$token = $this->user->generateAccessToken('foo', ['foo.bar'], $expiresAt);
```

src/Authentication/Traits/HasAccessTokens.php
```diff
@@ -165,4 +171,46 @@ public function setAccessToken(?AccessToken $accessToken): self
 
         return $this;
     }
+
+    /**
+     * Checks if the provided Access Token has expired.
+     *
+     * @return bool Returns true if Access Token has expired, false if not
+     */
+    public function hasAccessTokenExpired(?AccessToken $accessToken): bool
```
michalsn:
```suggestion
    public function hasAccessTokenExpired(AccessToken $accessToken): bool
```
I don't think we should allow null.

src/Authentication/Traits/HasAccessTokens.php
```diff
@@ -165,4 +171,46 @@ public function setAccessToken(?AccessToken $accessToken): self
...
+    }
+
+    /**
+     * Checks if the current Hmac token can expire
+     *
+     * @return bool Returns true if AccessToken can expire.
+     */
+    public function canAccessTokenExpire(AccessToken $accessToken): bool
+    {
+        return isset($accessToken->expires);
```
michalsn:
```suggestion
        return $accessToken->expires !== null;
```
Since the existence of the object is guaranteed, we can go with a strict check.

src/Authentication/Traits/HasHmacTokens.php
```diff
@@ -156,4 +161,46 @@ public function setHmacToken(?AccessToken $accessToken): self
 
         return $this;
     }
+
+    /**
+     * Checks if the provided Access Token has expired.
+     *
+     * @return bool Returns true if Access Token has expired, false if not
+     */
+    public function hasHmacTokenExpired(?AccessToken $accessToken): bool
```
michalsn:
```suggestion
    public function hasHmacTokenExpired(AccessToken $accessToken): bool
```

src/Authentication/Traits/HasAccessTokens.php
```diff
@@ -165,4 +171,46 @@ public function setAccessToken(?AccessToken $accessToken): self
...
+
+    /**
+     * Sets an expiration for Access Tokens by ID.
+     *
+     * @param int  $id        AccessTokens ID
+     * @param Time $expiresAt Expiration date
+     *
+     * @return bool Returns true if expiration date is set or updated.
+     */
+    public function setAccessTokenExpirationById(int $id, Time $expiresAt): bool
```
michalsn: I wonder... since we already allow the expiration date to be changed, should we also allow the expiration to be removed altogether?

src/Authentication/Traits/HasHmacTokens.php
```diff
@@ -156,4 +161,46 @@ public function setHmacToken(?AccessToken $accessToken): self
...
+
+    /**
+     * Sets an expiration for Hmac Key/Token by ID.
+     *
+     * @param int  $id        AccessToken ID
+     * @param Time $expiresAt Expiration date
+     *
+     * @return bool Returns true if expiration date is set or updated.
+     */
+    public function setHmacTokenExpirationById(int $id, Time $expiresAt): bool
```
michalsn: Same question as with `setAccessTokenExpirationById()` - should we allow `?Time`?

src/Authentication/Traits/HasHmacTokens.php
```diff
@@ -156,4 +161,46 @@ public function setHmacToken(?AccessToken $accessToken): self
...
+    }
+
+    /**
+     * Checks if the current Hmac token can expire
+     *
+     * @return bool Returns true if Hmac Token can expire.
+     */
+    public function canHmacTokenExpire(AccessToken $accessToken): bool
+    {
+        return isset($accessToken->expires);
```
michalsn:
```suggestion
        return $accessToken->expires !== null;
```

michalsn (changes requested):
@CosDiabos Thank you for the updates. I have a few other comments.

Of other things, I see that one action "Detect Merge Commits" reports an error. I'm not sure because of which commit, but you can try to rebase (there were some changes in the codebase since you started this PR). Or alternatively, if you identify the commit that is interfering, you can try to remove it from this PR - here are quite clear instructions that should help: https://graphite.dev/guides/how-to-delete-a-git-commit#deleting-old-or-specific-commits

@datamweb - is this failed action a deal breaker for us or not really?

datamweb:
> is this failed action a deal breaker for us or not really?

@michalsn , commit [e6c10c0d65c3360639fe39120d48e3949b740238](https://github.com/codeigniter4/shield/pull/1219/commits/e6c10c0d65c3360639fe39120d48e3949b740238) was correctly detected by the action. The user has likely used `pull` to update their branch, whereas we emphasize using `rebase` instead.  

We prefer **avoiding merge commits** to keep the *graph clean*, but it is *not mandatory*. However, based on your explanation, I believe the user should be able to handle the issue.  

If the user needs to update their branch, they should use `rebase` instead.
https://github.com/codeigniter4/CodeIgniter4/blob/develop/contributing/workflow.md#updating-your-branch

src/Authentication/Traits/HasAccessTokens.php
CosDiabos: I think it could be useful. Right now, if a expiration date is set, it can't be undone intuitively.

Would it be better to make its own method or clear it when it isn't present (`?Time` as you mentioned)?

src/Authentication/Traits/HasAccessTokens.php
CosDiabos: Yeah, that make sense.

I also noticed that  `setAccessToken()` and `setHmacToken()` are allowing a null value, and they probably shouldn't unless there is a reason to clear the access/hmac token.

CosDiabos:
Hey!

Thank you all for your help and support.

Yes I had merged because I had a couple of out of date files that I was working on. I've removed the merge commit with the rebase. I think now everything is in order with most change requests updated!

docs/references/authentication/tokens.md
```diff
@@ -125,7 +125,7 @@ Configure **app/Config/AuthToken.php** for your needs.
 
 ### Access Token Lifetime
 
-Tokens will expire after a specified amount of time has passed since they have been used.
+Tokens will expire after a specified amount of time has passed since they last have been used.
```
paulbalandan:
```suggestion
Tokens will expire after a specified amount of time has passed since they have last been used.
```

src/Authentication/Traits/HasAccessTokens.php
michalsn: `?Time` is enough - I don't think we need a separate method for this. But this will require changes also to the `setIdentityExpirationById()` method.

docs/references/authentication/tokens.md
```diff
@@ -137,6 +137,58 @@ that CodeIgniter provides.
...
+$token = $this->user->generateAccessToken('foo', ['foo.bar'], $expiresAt);
+$this->user->canAccessTokenExpire($token2); // Returns false
+
+$token2 = $this->user->generateAccessToken('bar');
+$this->user->canAccessTokenExpire($token); // Returns true
+```
+
+
+### Access Token Expiration vs Lifetime
+Expiration and Lifetime are different concepts. The lifetime is the maximum time allowed for the token to exist since its last use. Token expiration, on the other hand, is a set date in which the Access Token will cease to function.
```
michalsn: Please add a line break between the section header and the content.

michalsn (commented):
As for Psalm errors, this is out of the scope of this PR. Seems like the new release has changed some default settings. You can add `ensureOverrideAttribute="false"` to the `psalm.xml` file, right after `findUnusedCode="false"` to fix this.

src/Authentication/Traits/HasAccessTokens.php
michalsn: Well, the `$accessToken` in `setAccessToken()` and `setHmacToken()` can be `null`, because the user may simply have no token (e.g. see `AccessTokens::loginById()`). 

But this new method doesn't rely on the `$this->currentAccessToken` property, but is rather intended to be used to administer tokens when we use `accessTokens()`. So, my thinking was the `$accessToken` should not be optional.

src/Authentication/Traits/HasAccessTokens.php
CosDiabos: Great, sounds good!

CosDiabos:
> As for Psalm errors, this is out of the scope of this PR. Seems like the new release has changed some default settings. You can add `ensureOverrideAttribute="false"` to the `psalm.xml` file, right after `findUnusedCode="false"` to fix this.

Yeah, I skimmed the errors looking for a file that I could have edited but couldn't find it. Thanks for the tip.

michalsn (approved):
Thank you - it looks good. Let's see what others think.

docs/references/authentication/hmac.md
```diff
@@ -97,6 +97,60 @@ You can revoke all HMAC Keys with the `revokeAllHmacTokens()` method.
 $user->revokeAllHmacTokens();
 ```
 
+## Expiring HMAC Keys
+
+By default, the HMAC keys don't expire unless they meet the HMAC Keys lifetime expiration after their last used date.
+
+HMAC keys can be set to expire through the `generateHmacToken()` method. This takes the expiration date as the $expiresAt argument. To update or remove an existing HMAC key expiration date use `setHmacTokenExpirationById($HmacTokenID, $expiresAt)`
```
datamweb:
```suggestion
HMAC keys can be set to expire through the `generateHmacToken()` method. This takes the expiration date as the `$expiresAt` argument. To update or remove an existing HMAC key expiration date use `setHmacTokenExpirationById($HmacTokenID, $expiresAt)`
```

docs/references/authentication/hmac.md
```diff
@@ -97,6 +97,60 @@ You can revoke all HMAC Keys with the `revokeAllHmacTokens()` method.
...
+
+`$expiresAt` [Time](/libraries/time.html) object
```
datamweb:
```suggestion
`$expiresAt` [Time](https://codeigniter.com/user_guide/libraries/time.html) object
```

docs/references/authentication/hmac.md
```diff
@@ -97,6 +97,60 @@ You can revoke all HMAC Keys with the `revokeAllHmacTokens()` method.
...
+
+```php
+// Expiration date = 2024-11-03 12:00:00
+$expiresAt = Time::parse('2024-11-03 12:00:00');
+$token = $this->user->generateHmacToken('foo', ['foo:bar'], $expiresAt);
```
datamweb:
```suggestion
$token = $this->user->generateHmacToken('foo', ['foo.bar'], $expiresAt);
```

docs/references/authentication/hmac.md
datamweb: _To update or remove an existing HMAC key expiration date use `setHmacTokenExpirationById($HmacTokenID, $expiresAt)`_

Since these are two distinct responsibilities, I think it's better to create separate functions for updating and removing the HMAC token expiration. This improves code clarity and maintainability.  

Update expiration date:
   ```php
   public function updateHmacTokenExpiration($hmacTokenID,  $expiresAt): bool
   ```
Remove expiration date (set to null)
   ```php
   public function removeHmacTokenExpiration($hmacTokenID): bool
   ```

What do you think?

docs/references/authentication/hmac.md
```diff
@@ -97,6 +97,60 @@ You can revoke all HMAC Keys with the `revokeAllHmacTokens()` method.
...
+$expiresAt = Time::now()->addMonths(1)->addDays(15);
+$token = $user->setHmacTokenExpirationById($token->id, $expiresAt);
+
+// Remove the expiration date
+$token = $user->setHmacTokenExpirationById($token->id, null);
+```
+
+The following support methods are also available:
+
+`hasHmacTokenExpired(AccessToken $HmacToken)` - Checks if the HMAC key has expired. Returns `true` if the HMAC key has expired, `false` if not.
```
datamweb: Wouldn't it be better to use `isHmacTokenExpired` instead of `hasHmacTokenExpired`? 
This name is clearer and more naturally indicates that the function checks whether the token has expired.

docs/references/authentication/hmac.md
```diff
@@ -97,6 +97,60 @@ You can revoke all HMAC Keys with the `revokeAllHmacTokens()` method.
...
+
+```php
+$expiresAt = Time::parse('2024-11-03 12:00:00');
+$token = $this->user->generateHmacToken('foo', ['foo.bar'], $expiresAt);
+
+$this->user->hasHmacTokenExpired($token); // Returns true
+```
+
+`canHmacTokenExpire(AccessToken $HmacToken)` - Checks if HMAC key has an expiration set. Returns `true` or `false` accordingly.
```
datamweb: Wouldn't `hasHmacTokenExpiry` be a better choice instead of `canHmacTokenExpire`? 
The word **hasExpiry** more clearly indicates whether the token has an expiration set, avoiding potential confusion with can, which might imply permission or the ability to change.

docs/references/authentication/hmac.md
```diff
@@ -97,6 +97,60 @@ You can revoke all HMAC Keys with the `revokeAllHmacTokens()` method.
...
+
+$token = $this->user->generateHmacToken('foo', ['foo.bar'], $expiresAt);
+$this->user->canHmacTokenExpire($token); // Returns true
+
+$token2 = $this->user->generateHmacToken('bar');
+$this->user->canHmacTokenExpire($token2); // Returns false
+```
+
+You can also easily set all existing HMAC keys/tokens as expired with the `spark` command:
+```
```
datamweb:
```suggestion
```console
```

docs/references/authentication/hmac.md
```diff
@@ -97,6 +97,60 @@ You can revoke all HMAC Keys with the `revokeAllHmacTokens()` method.
...
+php spark shield:hmac invalidateAll
+```
+**Careful!** This command invalidates _all_ keys for _all_ users.
```
datamweb:
```suggestion
**Careful!** This command invalidates _all_ keys for _all_ users.

!!! warning

    This command invalidates _all_ keys for _all_ users.
```

docs/references/authentication/tokens.md
```diff
@@ -137,6 +137,61 @@ that CodeIgniter provides.
 public $unusedTokenLifetime = YEAR;
 ```
 
+
+## Expiring Access Tokens
+
+By default, the Access Tokens don't expire unless they meet the Access Token lifetime expiration after their last used date.
+
+Access Tokens can be set to expire through the `generateAccessToken()` method. This takes the expiration date as the $expiresAt argument. To update or remove an existing HMAC key expiration date use `setAccessTokenById($HmacTokenID, $expiresAt)`
```
datamweb:
```suggestion
Access Tokens can be set to expire through the `generateAccessToken()` method. This takes the expiration date as the `$expiresAt` argument. To update or remove an existing HMAC key expiration date use `setAccessTokenById($HmacTokenID, $expiresAt)`
```
I have the same question about the function name `setHmacTokenExpirationById` here as well. What do you think? 😃

docs/references/authentication/tokens.md
```diff
@@ -137,6 +137,61 @@ that CodeIgniter provides.
...
+
+`$expiresAt` [Time](/libraries/time.html) object
```
datamweb:
```suggestion
`$expiresAt` [Time](https://codeigniter.com/user_guide/libraries/time.html) object
```

docs/references/authentication/tokens.md
```diff
@@ -137,6 +137,61 @@ that CodeIgniter provides.
...
+
+```php
+$expiresAt = Time::parse('2024-11-03 12:00:00');
+
+$token = $this->user->generateAccessToken('foo', ['foo.bar'], $expiresAt);
+
+$this->user->hasAccessTokenExpired($token); // Returns true
+```
+
+`canAccessTokenExpire(AccessToken $HmacToken)` - Checks if Access Token has an expiration set. Returns `true` or `false` accordingly.
```
datamweb: What do you think? 😃
```suggestion
`hasAccessTokenExpiry(AccessToken $HmacToken)` - Checks if Access Token has an expiration set. Returns `true` or `false` accordingly.
```

docs/references/authentication/tokens.md
```diff
@@ -137,6 +137,61 @@ that CodeIgniter provides.
...
+$expiresAt = Time::now()->addMonths(1)->addDays(15);
+$user->setAccessTokenExpirationById($token->id, $expiresAt);
+
+// Remove the expiration date
+$token = $user->setAccessTokenExpirationById($token->id, null);
+```
+
+The following support methods are also available:
+
+`hasAccessTokenExpired(AccessToken $accessToken)` - Checks if Access Token has expired. Returns `true` if the Access Token has expired, `false` if not.
```
datamweb: What do you think? 😃
```suggestion
`isAccessTokenExpired(AccessToken $accessToken)` - Checks if Access Token has expired. Returns `true` if the Access Token has expired, `false` if not.
```

src/Authentication/Authenticators/AccessTokens.php
```diff
@@ -154,6 +154,19 @@ public function check(array $credentials): Result
 
         assert($token->last_used_at instanceof Time || $token->last_used_at === null);
 
+        // Is expired ?
+        if (
+            $token->expires !== null
+            && $token->expires->isBefore(
+                Time::now(),
```
datamweb, line 161: Did the tools add this, or is it extra?
```suggestion
                Time::now()
```

src/Models/UserIdentityModel.php
```diff
@@ -224,6 +229,24 @@ public function getAllAccessTokens(User $user): array
...
+     * @param mixed $id
+     *
+     * @return bool Returns true if expiration date was set or updated.
+     */
+    public function setIdentityExpirationById($id, User $user, ?Time $expiresAt = null, ?string $type_token = null): bool
+    {
+        $this->checkUserId($user);
+
+        return $this->where('user_id', $user->id)
+            ->where('type', $type_token)
```
datamweb: If the `$type_token` is `null`, what happens with the `where('type', $type_token) `condition?

tests/Authentication/HasAccessTokensTest.php
```diff
@@ -140,24 +141,111 @@ public function testTokenCanNoTokenSet(): void
...
+        $tokenExpiration = Time::parse('2024-11-03 12:00:00');
+        $token           = $this->user->generateAccessToken('foo', ['foo.bar'], $tokenExpiration);
         $this->user->setAccessToken($token);
 
-        $this->assertFalse($this->user->tokenCant('foo:bar'));
-        $this->assertTrue($this->user->tokenCant('foo:baz'));
+        $this->assertSame($tokenExpiration->format('Y-m-d h:i:s'), $this->user->currentAccessToken()->expires->format('Y-m-d h:i:s'));
+
+        $tokenExpiration = $tokenExpiration->addMonths(1);
+        $tokenExpiration = $tokenExpiration->addYears(1);
```
datamweb:
```suggestion
        $tokenExpiration = $tokenExpiration->addYears(1)->addMonths(1);
```

tests/Authentication/HasHmacTokensTest.php
```diff
@@ -149,10 +150,110 @@ public function testHmacTokenCantNoTokenSet(): void
...
+    {
+        $tokenExpiration = Time::parse('2024-11-03 12:00:00');
+
+        $token = $this->user->generateHmacToken('foo', ['foo.bar'], $tokenExpiration);
+        $this->user->setHmacToken($token);
+
+        $this->assertSame($tokenExpiration->format('Y-m-d h:i:s'), $this->user->currentHmacToken()->expires->format('Y-m-d h:i:s'));
+
+        $tokenExpiration = $tokenExpiration->addMonths(1);
+        $tokenExpiration = $tokenExpiration->addYears(1);
```
datamweb:
```suggestion
        $tokenExpiration = $tokenExpiration->addYears(1)->addMonths(1);
```

tests/Authentication/HasHmacTokensTest.php
```diff
@@ -149,10 +150,110 @@ public function testHmacTokenCantNoTokenSet(): void
...
+        $this->assertTrue($this->user->hasHmacTokenExpired($token));
+    }
+
+    /**
+     * See https://github.com/codeigniter4/shield/issues/926
+     */
+    public function testHmacTokenTimeToExpired(): void
+    {
+        $tokenExpiration = Time::now();
+        $tokenExpiration = $tokenExpiration->addYears(1);
```
datamweb, line 218:
```suggestion
        $tokenExpiration = Time::now()->addYears(1);
```

tests/Authentication/HasHmacTokensTest.php
```diff
@@ -149,10 +150,110 @@ public function testHmacTokenCantNoTokenSet(): void
...
+        $this->assertFalse($this->user->canHmacTokenExpire($token));
+    }
+
+    /**
+     * See https://github.com/codeigniter4/shield/issues/926
+     */
+    public function testHmacTokenRemoveExpiration(): void
+    {
+        $tokenExpiration = Time::now();
+        $tokenExpiration = $tokenExpiration->addYears(1);
```
datamweb:
```suggestion
        $tokenExpiration = Time::now()->addYears(1);
```

datamweb (changes requested):
@CosDiabos Thank you for submitting the PR, and I apologize for the long delay. I have provided some suggested changes.  

Regarding the method names, my understanding is based on the documentation, but I might have misinterpreted it. If I have misunderstood any part of the code, I would appreciate your clarification.

src/Authentication/Traits/HasHmacTokens.php
```diff
@@ -35,17 +38,19 @@ trait HasHmacTokens
     /**
      * Generates a new personal HMAC token for this user.
      *
-     * @param string       $name   Token name
-     * @param list<string> $scopes Permissions the token grants
+     * @param string       $name      Token name
+     * @param list<string> $scopes    Permissions the token grants
+     * @param Time         $expiresAt Expiration date
```
kenjis:
```suggestion
     * @param Time|null    $expiresAt Expiration date
```

src/Authentication/Traits/HasAccessTokens.php
```diff
@@ -34,15 +37,18 @@ trait HasAccessTokens
     /**
      * Generates a new personal access token for this user.
      *
-     * @param string       $name   Token name
-     * @param list<string> $scopes Permissions the token grants
+     * @param string       $name      Token name
+     * @param list<string> $scopes    Permissions the token grants
+     * @param Time         $expiresAt Expiration date
```
kenjis:
```suggestion
     * @param Time|null    $expiresAt Expiration date
```

docs/references/authentication/hmac.md
kenjis: Yes, I also think `update***()` and `remove***()` would be better.

CosDiabos: Yeah, michalsn and I talked about that same question (separation) earlier and kept everything on the same method.

However, I understand that the difference in functionality should warrant their own methods, also makes the removal of the expiration more conscious and less error prone action. I agree in splitting the actions.

docs/references/authentication/hmac.md
CosDiabos: I agree it's better semantically.

src/Authentication/Authenticators/AccessTokens.php
CosDiabos: Yes it's from the CS tool. Due to 'trailing_comma_in_multiline' rule, I guess.

src/Models/UserIdentityModel.php
CosDiabos: Actually, it shouldn't be accepting a null value for that argument because `$type_token` should always be defined (Access or Hmac).

But, what I'm realising is if the `$id` is already provided, the `$type_token` arg and ` ->where('type', $type_token)` on the update seems too much since the `$id` already points there.

docs/references/authentication/hmac.md
```diff
@@ -146,10 +146,12 @@ $this->user->canHmacTokenExpire($token2); // Returns false
...
 
 You can also easily set all existing HMAC keys/tokens as expired with the `spark` command:
-```
+```console
 php spark shield:hmac invalidateAll
 ```
-**Careful!** This command invalidates _all_ keys for _all_ users.
+
+!!! warning
+    This command invalidates _all_ keys for _all_ users.
```
datamweb, line 155:
```suggestion

    This command invalidates _all_ keys for _all_ users.
```

docs/references/authentication/hmac.md
```diff
@@ -97,6 +97,62 @@ You can revoke all HMAC Keys with the `revokeAllHmacTokens()` method.
...
+
+The following support methods are also available:
+
+`isHmacTokenExpired(AccessToken $HmacToken)` - Checks if the HMAC key is expired. Returns `true` if the HMAC key is expired, `false` if not.
+
+```php
+$expiresAt = Time::parse('2024-11-03 12:00:00');
+$token = $this->user->generateHmacToken('foo', ['foo.bar'], $expiresAt);
+
+$this->user->hasHmacTokenExpired($token); // Returns true
```
datamweb: Did you mean?
```suggestion
$this->user->isHmacTokenExpired($token); // Returns true
```

docs/references/authentication/hmac.md
```diff
@@ -97,6 +97,63 @@ You can revoke all HMAC Keys with the `revokeAllHmacTokens()` method.
...
+```
+
+`hasHmacTokenExpiry(AccessToken $HmacToken)` - Checks if HMAC key has an expiration set. Returns `true` or `false` accordingly.
+
+```php
+$expiresAt = Time::parse('2024-11-03 12:00:00');
+
+$token = $this->user->generateHmacToken('foo', ['foo.bar'], $expiresAt);
+$this->user->canHmacTokenExpire($token); // Returns true
```
datamweb, line 142: And ?

docs/references/authentication/hmac.md
```diff
@@ -97,6 +97,63 @@ You can revoke all HMAC Keys with the `revokeAllHmacTokens()` method.
...
+
+$token2 = $this->user->generateHmacToken('bar');
+$this->user->canHmacTokenExpire($token2); // Returns false
```
datamweb, line 145: And?

docs/references/authentication/tokens.md
```diff
@@ -137,6 +137,61 @@ that CodeIgniter provides.
...
+The following support methods are also available:
+
+`isAccessTokenExpired(AccessToken $accessToken)` - Checks if Access Token is expired. Returns `true` if the Access Token is expired, `false` if not.
+
+```php
+$expiresAt = Time::parse('2024-11-03 12:00:00');
+
+$token = $this->user->generateAccessToken('foo', ['foo.bar'], $expiresAt);
+
+$this->user->hasAccessTokenExpired($token); // Returns true
```
datamweb: And?
```suggestion
$this->user->isAccessTokenExpired($token); // Returns true
```

docs/references/authentication/tokens.md
```diff
@@ -137,6 +137,61 @@ that CodeIgniter provides.
...
+```
+
+`hasAccessTokenExpiry(AccessToken $HmacToken)` - Checks if Access Token has an expiration set. Returns `true` or `false` accordingly.
+
+```php
+$expiresAt = Time::parse('2024-11-03 12:00:00');
+
+$token = $this->user->generateAccessToken('foo', ['foo.bar'], $expiresAt);
+$this->user->canAccessTokenExpire($token2); // Returns false
```
datamweb, line 184: And?
```suggestion
$this->user->canAccessTokenExpire($token2); // Returns false
```

docs/references/authentication/tokens.md
```diff
@@ -137,6 +137,61 @@ that CodeIgniter provides.
...
+
+$token2 = $this->user->generateAccessToken('bar');
+$this->user->canAccessTokenExpire($token); // Returns true
```
datamweb, line 187: And?
```suggestion
$this->user->canAccessTokenExpire($token); // Returns true
```

src/Authentication/Traits/HasHmacTokens.php
```diff
@@ -156,4 +160,67 @@ public function setHmacToken(?AccessToken $accessToken): self
 
         return $this;
     }
+
+    /**
+     * Checks if the provided Access Token is expired.
+     *
+     * @return bool Returns true if Access Token is expired, false if not
```
paulbalandan: This comment is unnecessary as it can be inferred from the method description what it does.
```suggestion
```

src/Authentication/Traits/HasAccessTokens.php
```diff
@@ -165,4 +171,67 @@ public function setAccessToken(?AccessToken $accessToken): self
 
         return $this;
     }
+
+    /**
+     * Checks if the provided Access Token is expired.
+     *
+     * @return bool Returns true if Access Token is expired, false if not
```
paulbalandan:
```suggestion
```

datamweb (approved):
Please update the PR description. Also, be aware that, based on coordination with the team, others may provide suggestions for improvements here. I appreciate your patience in advance.

src/Authentication/Traits/HasAccessTokens.php
```diff
@@ -165,4 +171,67 @@ public function setAccessToken(?AccessToken $accessToken): self
...
+        if ($result) {
+            // refresh currentAccessToken with updated data
+            $this->currentAccessToken = $identityModel->getAccessTokenById($id, $this);
+        }
+
+        return $result;
+    }
+
+    /**
+     * Checks if the current Hmac token can expire
```
paulbalandan:
```suggestion
     * Checks if the access token has a set expiration date
```

src/Authentication/Traits/HasAccessTokens.php
```diff
@@ -165,4 +171,67 @@ public function setAccessToken(?AccessToken $accessToken): self
...
+     *
+     * @return bool Returns true if AccessToken can expire.
```
paulbalandan:
```suggestion
```

src/Authentication/Traits/HasHmacTokens.php
```diff
@@ -156,4 +160,67 @@ public function setHmacToken(?AccessToken $accessToken): self
...
+        if ($result) {
+            // refresh currentAccessToken with updated data
+            $this->currentHmacToken = $identityModel->getHmacTokenById($id, $this);
+        }
+
+        return $result;
+    }
+
+    /**
+     * Checks if the current Hmac token can expire
```
paulbalandan:
```suggestion
     * Checks if the current HMAC token has a set expiration date
```

src/Authentication/Traits/HasHmacTokens.php
```diff
@@ -156,4 +160,67 @@ public function setHmacToken(?AccessToken $accessToken): self
...
+     *
+     * @return bool Returns true if Hmac Token can expire.
```
paulbalandan:
```suggestion
```

src/Commands/Hmac.php
```diff
@@ -196,4 +202,33 @@ static function ($identity) use ($uIdModelSub, $encrypter, $that): void {
...
+
+    /**
+     * Invalidates all HMAC Keys/Tokens for every user.
+     */
+    public function invalidateAll(): void
+    {
+        $uIdModel    = new UserIdentityModel();
+        $uIdModelSub = new UserIdentityModel();
+
+        $that = $this;
```
paulbalandan: Why do you need to assign `$this` to a variable? You can just make the closure below as not static then just access `$this->write` normally.

src/Commands/Hmac.php
```diff
@@ -196,4 +202,33 @@ static function ($identity) use ($uIdModelSub, $encrypter, $that): void {
...
+
+        $uIdModel->where('type', 'hmac_sha256')->orderBy('id')->chunk(
+            100,
+            static function ($identity) use ($uIdModelSub, $that): void {
```
paulbalandan:
```suggestion
            function ($identity) use ($uIdModelSub): void {
```

src/Commands/Hmac.php
```diff
@@ -196,4 +202,33 @@ static function ($identity) use ($uIdModelSub, $encrypter, $that): void {
...
+                $timeNow = Time::now(); // Current date/time
```
paulbalandan:
```suggestion
                $timeNow = Time::now();
```

src/Commands/Hmac.php
```diff
@@ -196,4 +202,33 @@ static function ($identity) use ($uIdModelSub, $encrypter, $that): void {
...
+
+                if (null !== $identity->expires && $identity->expires->isBefore($timeNow)) {
+                    $that->write('Hmac Key/Token ID: ' . $identity->id . ', already expired, skipped.');
```
paulbalandan:
```suggestion
                    $this->write('Hmac Key/Token ID: ' . $identity->id . ', already expired, skipped.');
```

src/Commands/Hmac.php
```diff
@@ -196,4 +202,33 @@ static function ($identity) use ($uIdModelSub, $encrypter, $that): void {
...
+
+                    return;
+                }
+
+                $identity->expires = $timeNow;
+                $uIdModelSub->save($identity);
+
+                $that->write('Hmac Key/Token ID: ' . $identity->id . ', set as expired.');
```
paulbalandan:
```suggestion
                $this->write('Hmac Key/Token ID: ' . $identity->id . ', set as expired.');
```

CosDiabos:
> Please update the PR description. Also, be aware that, based on coordination with the team, others may provide suggestions for improvements here. I appreciate your patience in advance.

Updated! No problem!

src/Commands/Hmac.php
CosDiabos: Hey!

Thanks! I wasn't sure about how the Closure worked, and the docs sample for `chunk()` uses a static method, which, seemed the right way to do it.

docs/references/authentication/hmac.md
```diff
@@ -97,6 +97,63 @@ You can revoke all HMAC Keys with the `revokeAllHmacTokens()` method.
 $user->revokeAllHmacTokens();
 ```
 
+## Expiring HMAC Keys
+
+By default, the HMAC keys don't expire unless they meet the HMAC Keys lifetime expiration after their last used date.
```
paulbalandan:
```suggestion
By default, the HMAC keys don't expire unless they reach the HMAC keys' lifetime expiration after their last use date.
```

docs/references/authentication/hmac.md
```diff
@@ -97,6 +97,63 @@ You can revoke all HMAC Keys with the `revokeAllHmacTokens()` method.
...
+
+HMAC keys can be set to expire through the `generateHmacToken()` method. This takes the expiration date as the `$expiresAt` argument. To update an existing HMAC key expiration date use `updateHmacTokenExpiration($hmacTokenID, $expiresAt)` and to remove `removeHmacTokenExpiration($hmacTokenID)`.
```
paulbalandan:
```suggestion
HMAC keys can be set to expire through the `generateHmacToken()` method. This takes the expiration date as the `$expiresAt` argument. To update an existing HMAC key expiration date, use `updateHmacTokenExpiration($hmacTokenID, $expiresAt)`. To remove it, use `removeHmacTokenExpiration($hmacTokenID)`.
```

docs/references/authentication/hmac.md
```diff
@@ -97,6 +97,63 @@ You can revoke all HMAC Keys with the `revokeAllHmacTokens()` method.
...
+$expiresAt = Time::now()->addMonths(1)->addDays(15);
+$user->updateHmacTokenExpiration($token->id, $expiresAt);
+
+// Remove the expiration date
+$user->removeHmacTokenExpiration($token->id);
+```
+
+The following support methods are also available:
+
+`isHmacTokenExpired(AccessToken $HmacToken)` - Checks if the HMAC key is expired. Returns `true` if the HMAC key is expired, `false` if not.
```
paulbalandan:
```suggestion
`isHmacTokenExpired(AccessToken $hmacToken)` - Checks if the HMAC key is expired. Returns `true` if the HMAC key is expired; otherwise, returns `false`.
```

docs/references/authentication/hmac.md
```diff
@@ -97,6 +97,63 @@ You can revoke all HMAC Keys with the `revokeAllHmacTokens()` method.
...
+
+```php
+$expiresAt = Time::parse('2024-11-03 12:00:00');
+$token = $this->user->generateHmacToken('foo', ['foo.bar'], $expiresAt);
+
+$this->user->isHmacTokenExpired($token); // Returns true
+```
+
+`hasHmacTokenExpiry(AccessToken $HmacToken)` - Checks if HMAC key has an expiration set. Returns `true` or `false` accordingly.
```
paulbalandan:
```suggestion
`hasHmacTokenExpiry(AccessToken $hmacToken)` - Checks if HMAC key has an expiration set. Returns `true` if the HMAC key is expired; otherwise, returns `false`.
```

docs/references/authentication/hmac.md
```diff
@@ -228,6 +285,10 @@ that CodeIgniter provides.
 public $unusedTokenLifetime = YEAR;
 ```
 
+### HMAC Keys Expiration vs Lifetime
+
+Expiration and Lifetime are different concepts. The lifetime is the maximum time allowed for the HMAC Key to exist since its last use. HMAC Key expiration, on the other hand, is a set date in which the HMAC Key will cease to function.
```
paulbalandan:
```suggestion
Expiration and lifetime are two different concepts. The lifetime is the maximum time allowed for the HMAC Key to exist since its last use. HMAC Key expiration, on the other hand, is a set date in which the HMAC Key will cease to function.
```

docs/references/authentication/tokens.md
```diff
@@ -137,6 +137,61 @@ that CodeIgniter provides.
 public $unusedTokenLifetime = YEAR;
 ```
 
+
+## Expiring Access Tokens
+
+By default, the Access Tokens don't expire unless they meet the Access Token lifetime expiration after their last used date.
```
paulbalandan:
```suggestion
By default, the Access Tokens don't expire unless they reach the Access Token's lifetime expiration after their last use date.
```

docs/references/authentication/tokens.md
```diff
@@ -137,6 +137,61 @@ that CodeIgniter provides.
...
+
+Access Tokens can be set to expire through the `generateAccessToken()` method. This takes the expiration date as the `$expiresAt` argument. To update an existing HMAC key expiration date use `updateAcessTokenExpiration($hmacTokenID, $expiresAt)` and to remove `removeAccessTokenExpiration($hmacTokenID)`.
```
paulbalandan:
```suggestion
Access Tokens can be set to expire through the `generateAccessToken()` method. This takes the expiration date as the `$expiresAt` argument. To update an existing HMAC key expiration date, use `updateAcessTokenExpiration($accessTokenID, $expiresAt)`. To remove it, use `removeAccessTokenExpiration($accessTokenID)`.
```

docs/references/authentication/tokens.md
```diff
@@ -137,6 +137,61 @@ that CodeIgniter provides.
...
+$expiresAt = Time::now()->addMonths(1)->addDays(15);
+$user->updateAcessTokenExpiration($token->id, $expiresAt);
+
+// Remove the expiration date
+$user->removeAccessTokenExpiration($token->id);
+```
+
+The following support methods are also available:
+
+`isAccessTokenExpired(AccessToken $accessToken)` - Checks if Access Token is expired. Returns `true` if the Access Token is expired, `false` if not.
```
paulbalandan:
```suggestion
`isAccessTokenExpired(AccessToken $accessToken)` - Checks if Access Token is expired. Returns `true` if the Access Token is expired; otherwise, returns `false`.
```

docs/references/authentication/tokens.md
```diff
@@ -137,6 +137,61 @@ that CodeIgniter provides.
...
+
+```php
+$expiresAt = Time::parse('2024-11-03 12:00:00');
+
+$token = $this->user->generateAccessToken('foo', ['foo.bar'], $expiresAt);
+
+$this->user->isAccessTokenExpired($token); // Returns true
+```
+
+`hasAccessTokenExpiry(AccessToken $HmacToken)` - Checks if Access Token has an expiration set. Returns `true` or `false` accordingly.
```
paulbalandan:
```suggestion
`hasAccessTokenExpiry(AccessToken $accessToken)` - Returns `true` if the Access Token has a set expiration date; otherwise, returns `false`.
```

docs/references/authentication/tokens.md
```diff
@@ -137,6 +137,61 @@ that CodeIgniter provides.
...
+$this->user->hasAccessTokenExpiry($token2); // Returns false
+
+$token2 = $this->user->generateAccessToken('bar');
+$this->user->hasAccessTokenExpiry($token); // Returns true
+```
+
+
+### Access Token Expiration vs Lifetime
+
+Expiration and Lifetime are different concepts. The lifetime is the maximum time allowed for the token to exist since its last use. Token expiration, on the other hand, is a set date in which the Access Token will cease to function.
```
paulbalandan:
```suggestion
Expiration and lifetime are two different concepts. The lifetime is the maximum time allowed for the token to exist since its last use. Token expiration, on the other hand, is a set date in which the Access Token will cease to function.
```

src/Authentication/Traits/HasHmacTokens.php
```diff
@@ -156,4 +160,63 @@ public function setHmacToken(?AccessToken $accessToken): self
...
         return $this;
     }
+
+    /**
+     * Checks if the provided Access Token is expired.
+     */
+    public function isHmacTokenExpired(AccessToken $accessToken): bool
+    {
+        return $accessToken->expires !== null && $accessToken->expires->isBefore(Time::now());
+    }
```
paulbalandan:
```suggestion
     * Checks if the provided HMAC Token is expired.
     */
    public function isHmacTokenExpired(AccessToken $hmacToken): bool
    {
        return $hmacToken->expires !== null && $hmacToken->expires->isBefore(Time::now());
    }
```

src/Authentication/Traits/HasHmacTokens.php
```diff
@@ -156,4 +160,63 @@ public function setHmacToken(?AccessToken $accessToken): self
...
+
+    /**
+     * Sets an expiration for HMAC token by ID.
+     *
+     * @param int  $id        AccessToken ID
```
paulbalandan:
```suggestion
     * @param int  $id        HMAC Token ID
```

src/Authentication/Traits/HasHmacTokens.php
```diff
@@ -156,4 +160,63 @@ public function setHmacToken(?AccessToken $accessToken): self
...
+            $this->currentAccessToken = $identityModel->getHmacTokenById($id, $this);
+        }
+
+        return $result;
+    }
+
+    /**
+     * Removes the expiration date for HMAC token by ID.
+     *
+     * @param int $id AccessToken ID
```
paulbalandan:
```suggestion
     * @param int $id HMAC Token ID
```

src/Authentication/Traits/HasHmacTokens.php
```diff
@@ -156,4 +160,63 @@ public function setHmacToken(?AccessToken $accessToken): self
...
+        return $result;
+    }
+
+    /**
+     * Checks if the current HMAC token has a set expiration date
+     */
+    public function hasHmacTokenExpiry(AccessToken $accessToken): bool
+    {
+        return $accessToken->expires !== null;
+    }
```
paulbalandan:
```suggestion
    public function hasHmacTokenExpiry(AccessToken $hmacToken): bool
    {
        return $hmacToken->expires !== null;
    }
```

paulbalandan (commented):
Some grammar fixes

paulbalandan (approved):

src/Authentication/Traits/HasAccessTokens.php
```diff
@@ -165,4 +171,63 @@ public function setAccessToken(?AccessToken $accessToken): self
...
+            $this->currentAccessToken = $identityModel->getAccessTokenById($id, $this);
+        }
+
+        return $result;
+    }
+
+    /**
+     * Checks if the access token has a set expiration date
+     */
+    public function hasAccessTokenExpiry(AccessToken $accessToken): bool
```
michalsn: I'm not sure if this has been changed or if it was here from the beginning, but reading the code, I thought: "Wait a minute, don't we have two methods with the same functionality?".

If someone reads the method name without checking its implementation, they might assume it returns true when the token is already expired, not just when an expiry date is set.

IMO "Has expiry" could be interpreted as "Has the token expired?" instead of "Does it have an expiry date?".

Maybe we could go with `canAccessTokenExpire()` or `isAccessTokenExpirable()`?

Same thing for Hmac.

CosDiabos: The method names have been a challenge so far, haha.

> 
> Maybe we could go with `canAccessTokenExpire()` or `isAccessTokenExpirable()`?
> 

I don't think `hasAccessTokenExpiry()` is necessarily confusing, but `canAccessTokenExpire()` seems to be more straightforward. `isAccessTokenExpirable()` is confusing, imo

michalsn: > The method names have been a challenge so far, haha.

Yeah, apparently 😆 

Let's wait a day - if there are no more opinions on the subject, choose what you think is best.

michalsn (approved):

src/Authentication/Authenticators/AccessTokens.php
```diff
@@ -154,6 +154,19 @@ public function check(array $credentials): Result
 
         assert($token->last_used_at instanceof Time || $token->last_used_at === null);
 
+        // Is expired ?
+        if (
+            $token->expires !== null
```
samsonasik: can this use instanceof, eg : 

```suggestion
            $token->expires instanceof Time
```

michalsn: Sounds good.

michalsn: @CosDiabos Would you find time for these final changes? I'm looking forward to merging this PR. Thanks!

src/Authentication/Traits/HasAccessTokens.php
```diff
@@ -165,4 +171,63 @@ public function setAccessToken(?AccessToken $accessToken): self
...
+            $this->currentAccessToken = $identityModel->getAccessTokenById($id, $this);
+        }
+
+        return $result;
+    }
+
+    /**
+     * Checks if the access token has a set expiration date
+     */
+    public function CanAccessTokenExpire(AccessToken $accessToken): bool
```
michalsn:
```suggestion
    public function canAccessTokenExpire(AccessToken $accessToken): bool
```
On more issue. The same applies to HMAC + all the mentions in the docs.

michalsn:
@CosDiabos Thank you for your updates and patience, and congratulations on your first contribution to the project!

CosDiabos:
@michalsn You're welcome! Thank you all for your help and reviews throughout.

PR #1220, opened 2024-11-04
feat: Settings - Decouple from shield

najdanovicivan:
**Description**
Decouple settings library so it is not required.
Related to #1215

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


src/Helpers/auth_helper.php
```diff
@@ -43,3 +44,33 @@ function user_id()
...
+            $parts = explode('.', $key);
+            if (count($parts) === 1) {
+                throw new InvalidArgumentException('$key must contain both the class and field name, i.e. Foo.bar');
+            }
+            return config($parts[0])?->{$parts[1]} ?? null;
+        }
+
+        throw new InvalidArgumentException('Settings library is not being used for shield');
+    }
+}
```
najdanovicivan, line 76: This is the important part. With the proposed change call to setting() method will only occur here.

TwistedAndy:
Yes, it would be great to decouple Shield and Settings packages. Personally, I don't see a lot of sense of having the Settings dependency here, because CodeIgniter already has BaseConfig to handle settings.

najdanovicivan:
@datamweb What else is needed for this one to be merged ?

datamweb:
@najdanovicivan I'm currently on mobile and need to review this PR on my PC. I'll get back to it soon.

docs/addons/jwt.md
```diff
@@ -190,7 +190,7 @@ class LoginController extends BaseController
         }
 
         // Get the credentials for login
-        $credentials             = $this->request->getJsonVar(setting('Auth.validFields'));
+        $credentials             = $this->request->getJsonVar(shieldSetting('Auth.validFields'));
```
paulbalandan, line 193: I believe this should be in snake case per our coding standards: `shield_setting`

src/Helpers/auth_helper.php
```diff
@@ -43,3 +44,33 @@ function user_id()
...
 }
+
+if (! function_exists('shieldSetting')) {
+    /**
+     * Provides a wrapper function for settings module.
+     *
+     * @param mixed $value
+     *
+     * @return array|bool|float|int|object|string|void|null
+     * @phpstan-return ($value is null ? array|bool|float|int|object|string|null : void)
```
paulbalandan, line 55: In what instance will this return void?

```suggestion
     * @return mixed
```

src/Helpers/auth_helper.php
```diff
@@ -43,3 +44,33 @@ function user_id()
...
+     */
+    function shieldSetting(?string $key = null, $value = null)
```
paulbalandan, line 57:
```suggestion
    function shield_setting(?string $key = null, mixed $value = null): mixed
```

src/Helpers/auth_helper.php
```diff
@@ -43,3 +44,33 @@ function user_id()
...
+    {
+        /** @var AuthConfig $config */
+        $config = config('Auth');
+        if($config->useSettings) {
```
paulbalandan, line 61: Please add a function_exists check

src/Helpers/auth_helper.php
```diff
@@ -43,3 +44,33 @@ function user_id()
...
+            return setting($key, $value);
+        }
+
+        // Getting the value?
+        if (!empty($key) && count(func_get_args()) === 1) {
```
paulbalandan, line 66: What's the use case for having null key? Would it be better to just require a string key?

src/Helpers/auth_helper.php
```diff
@@ -43,3 +44,33 @@ function user_id()
...
+            $parts = explode('.', $key);
+            if (count($parts) === 1) {
+                throw new InvalidArgumentException('$key must contain both the class and field name, i.e. Foo.bar');
+            }
+            return config($parts[0])?->{$parts[1]} ?? null;
```
paulbalandan, line 71: The `?->` would return null if `config($parts[0])` is not an object so the `?? null` is useless.

src/Helpers/auth_helper.php
```diff
@@ -43,3 +44,33 @@ function user_id()
...
+        }
+
+        throw new InvalidArgumentException('Settings library is not being used for shield');
```
paulbalandan, line 74: This capturing exception feels off. This would be reached if:

1. `Auth::$useSettings` is true but `settings` is not installed.
2. `$useSettings` is false but `$key` is null or empty string.
3. `$useSettings` is false, `$key` is non-empty-string, and $value is provided

These different scenarios should be tackled separately within each conditional above.

composer.json
```diff
@@ -42,7 +41,8 @@
     },
     "suggest": {
         "ext-curl": "Required to use the password validation rule via PwnedValidator class.",
-        "ext-openssl": "Required to use the JWT Authenticator."
+        "ext-openssl": "Required to use the JWT Authenticator.",
+        "codeigniter4/settings": "Required to store groups and permissions in database"
```
paulbalandan, line 45: Add this also to `require-dev`

src/Helpers/auth_helper.php
```diff
@@ -43,3 +44,33 @@ function user_id()
         return $auth->id();
     }
 }
+
+if (! function_exists('shieldSetting')) {
```
paulbalandan, line 48: Please add a test for this new function.

composer.json
```diff
@@ -25,8 +25,7 @@
         "slack": "https://codeigniterchat.slack.com"
     },
     "require": {
-        "php": "^8.1",
-        "codeigniter4/settings": "^2.1"
```
datamweb, line 29: The current software relies on the **codeigniter4/settings** package for data storage.
After installing or updating to the new version of **Shield**, this package(`codeigniter4/settings`) will be automatically removed, which could lead to a **breaking change**, unless it remains required by another package.

I’m deeply concerned about this, and at the very least, this change should be clearly, explicitly, and transparently documented in the docs and [UPGRADING](https://github.com/codeigniter4/shield/blob/develop/UPGRADING.md).

src/Config/Auth.php
```diff
@@ -40,6 +40,13 @@ class Auth extends BaseConfig
     public const RECORD_LOGIN_ATTEMPT_FAILURE = 1; // Record only failures
     public const RECORD_LOGIN_ATTEMPT_ALL     = 2; // Record all login attempts
 
+    /**
+     * --------------------------------------------------------------------
+     * Use Settings
+     * --------------------------------------------------------------------
+     */
+    public bool $useSettings = true;
```
datamweb, line 48: You previously removed `"codeigniter4/settings": "^2.1"` from the require section:

If the `settings` package isn’t installed, doesn’t this make it essentially useless?
In any case, it will cause failures for any software that relies on the `settings` package.

src/Helpers/auth_helper.php
datamweb: Good question!

src/Helpers/auth_helper.php
```diff
@@ -43,3 +44,33 @@ function user_id()
...
+        /** @var AuthConfig $config */
+        $config = config('Auth');
+        if($config->useSettings) {
+            return setting($key, $value);
+        }
+
+        // Getting the value?
+        if (!empty($key) && count(func_get_args()) === 1) {
+            $parts = explode('.', $key);
+            if (count($parts) === 1) {
```
datamweb, line 68:
```suggestion
            if (count($parts) !== 2) {
```
How about keeping it this way?

datamweb (commented):
We reference the `settings` package directly in several parts of the documentation.
If this PR is going to be merged, these sections of the docs should be **updated and clarified** accordingly.
https://shield.codeigniter.com/getting_started/concepts/?h=settings#settings
https://shield.codeigniter.com/getting_started/concepts/?h=settings#repository-state
https://shield.codeigniter.com/?h=settings#important-features

Please pay attention to @paulbalandan’s review.
Also, part of the GitHub Action failures is related to the removal of the `settings` package.
Please try to ensure that the actions pass successfully.

PR #1248, opened 2025-02-10
fix: rector

michalsn:
**Description**
This PR ignores two Rector rules which were added in the latest versions.

I believe that the application of these rules will negatively affect the readability of our tests.

Here are the proposed changes in the codebase by Rector: https://github.com/codeigniter4/shield/actions/runs/13240274064/job/36953902890

I'm not super convinced about it - this was just my first impression.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


samsonasik (approved):

datamweb (approved):

datamweb:
I see this change as fair and reasonable.

paulbalandan (approved):

michalsn:
Thanks for the review guys!

PR #1251, opened 2025-02-14
docs: require `activator` in custom actions class name for `register`

### LINKED ISSUE #1249: Bug: User `isActivated()` always returns true, even when custom activator is configured
### PHP Version

8.3.9

### CodeIgniter4 Version

4.5.7

### Shield Version

1.1.0

Windows

### Which server did you use?

cli-server (PHP built-in webserver)

### Database

Mysql 8

### Did you customize Shield?

Yes.  I created a php class that uses TwilioVerify.  It sends a code to the user via SMS, then the user enters the code and the verify method checks the code. 

I am using the same class for both login and register.  the user must receive a code to log in, and also when registering. 

### What happened?

The functionality in general works fine.  The user must enter the code (and it must be valid) or they are not able to log in.  The same with registration. 
One thing I noticed is that I forgot about executing `$user->activate()` after registering and entering a code.  Then I realized that I was able to log in regardless of this.  

I added a check in my Login controller so that if `$user->isActivated()` returns false, they must not have entered a code upon registration, so they should not be permitted to log in.  However, this always returns true. 

I read in the documentation that this will always return true, unless an activator is specified in Auth.php.  My Auth.php DOES have actions configured, as can be seen below, so it seems like it should not always be returning true. 

```

    /**
     * Authentication Actions
     * Specifies the class that represents an action to take after
     * the user logs in or registers a new account at the site.
     *
     * You must register actions in the order of the actions to be performed.
     *
     * Available actions with Shield:
     * - register: \CodeIgniter\Shield\Authentication\Actions\EmailActivator::class
     * - login:    \CodeIgniter\Shield\Authentication\Actions\Email2FA::class
     *
     * @var array<string, class-string<ActionInterface>|null>
     */
    public array $actions = [
        'register' => TwilioVerify::class,
        'login'    => TwilioVerify::class,
        //'login'    => null,
    ];
```

### Steps to Reproduce

Create a custom class and use it as an authentication action on both register and login.  Leave the user's active flag = 0, and you will still always receive a true in response to `$user->isActivated()` 

### Expected Output

I expect to receive false when using `$user->isActivated()`, and the user's active flag = false, since I have the action specified in Auth.php

### Anything else?

_No response_

datamweb:
**Description**
fix: #1249

In this implementation, since we cannot determine whether the user provides just the class name or the fully qualified namespace of the class, we are unable to directly verify the `ActionInterface` implementation. Therefore, we rely on the class name suffix ('Activator') as a consistent identifier for validation.

see:
https://github.com/codeigniter4/shield/blob/58d6d6f8820aa17016da4001d5da544ed13e8e75/src/Traits/Activatable.php#L63

**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/docs/references/authentication/auth_actions.md
+++ b/docs/references/authentication/auth_actions.md
@@ -57,10 +57,12 @@ public $views = [
 ## Defining New Actions
 
 While the provided email-based activation and 2FA will work for many sites, others will have different
-needs, like using SMS to verify or something completely different. Actions have only one requirement:
-they must implement `CodeIgniter\Shield\Authentication\Actions\ActionInterface`.
+needs, like using SMS to verify or something completely different. Custom actions must adhere to the following requirements:
 
-The interface defines three methods for `ActionController`:
+1. The class name for a "register" action must end with the suffix `Activator` (e.g., `SMSActivator`) to ensure consistency.
+2. All custom actions must implement the `CodeIgniter\Shield\Authentication\Actions\ActionInterface`.
+
+The `ActionInterface` defines three required methods that must be implemented to ensure the action integrates properly with the `ActionController`.
 
 **show()** should display the initial page the user lands on immediately after the authentication task,
 like login. It will typically display instructions to the user and provide an action to take, like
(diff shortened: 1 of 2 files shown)


michalsn (approved):

paulbalandan (approved):

datamweb:
Thanks!

PR #1252, opened 2025-02-16
refactor: `AuthToken`/`AuthJWT` Config Loading by Initializing in Constructor

datamweb:
**Description**
 `AuthToken` and `AuthJWT` Config Loading by Initializing in Constructor

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/src/Authentication/Authenticators/AccessTokens.php
+++ b/src/Authentication/Authenticators/AccessTokens.php
@@ -18,6 +18,7 @@ use CodeIgniter\I18n\Time;
 use CodeIgniter\Shield\Authentication\AuthenticationException;
 use CodeIgniter\Shield\Authentication\AuthenticatorInterface;
 use CodeIgniter\Shield\Config\Auth;
+use CodeIgniter\Shield\Config\AuthToken;
 use CodeIgniter\Shield\Entities\User;
 use CodeIgniter\Shield\Exceptions\InvalidArgumentException;
 use CodeIgniter\Shield\Models\TokenLoginModel;
@@ -29,6 +30,7 @@ class AccessTokens implements AuthenticatorInterface
 {
     public const ID_TYPE_ACCESS_TOKEN = 'access_token';
 
+    protected AuthToken $authTokenConfig;
     protected ?User $user = null;
     protected TokenLoginModel $loginModel;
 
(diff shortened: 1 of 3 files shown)


paulbalandan (commented):
Do you have a benchmark for this change?

datamweb:
> Do you have a benchmark for this change?

The presence of **Optimize** probably led to this question. It seems I made a mistake in choosing it, as my main goal in this configuration was to prevent the repeated use of `config('AuthToken')` and `config('AuthJWT')`, **nothing more**.

But now I’m curious—does this change actually impact performance or not? Your benchmark results could be really helpful! @paulbalandan If possible, please share them so we can analyze them together.

paulbalandan:
Yes, the word Optimize in your PR title made me curious about the impact to performance since we will now be loading the configs everytime a class is instantiated. I just remembered one time when I did a similar change to the Seeder class instantiating Faker's Generator everytime in the constructor caused a massive slowdown.

michalsn:
Are we actually calling an `attempt` method multiple times per request?

datamweb:
> Are we actually calling an `attempt` method multiple times per request?

I didn't quite understand your question, and I'm not sure how it relates to this PR. Could you please clarify?

michalsn:
Sorry, your reason for this change was described as:

> my main goal in this configuration was to prevent the repeated use of config('AuthToken') and config('AuthJWT')

So my question is. is this really the case? Are we actually calling these configs multiple times, or only once per request?

datamweb:
> is this really the case? Are we actually calling these configs multiple times, or only once per request?

In fact, the configuration is only called once per request. Therefore, it seems this change wouldn’t have a significant impact on performance. 100%

Which one do you find more suitable and why?"

**First method:**

```php
<?php

declare(strict_types=1);

class MyClass
{

    public function __construct() {
        $this->config = config('Xconfig');
    }

    public function Y()
    {
        $this->config;
    }

    public function X()
    {

        $this->config;

    }
}
```

**Second method:**
```php
<?php

declare(strict_types=1);

class MyClass
{

    public function __construct() {}

    public function Y()
    {
        $config = config('Xconfig);
    }

    public function X()
    {

        $config = config('Xconfig);

    }
}
```

michalsn:
Ok, sorry - that was my mistake. I thought we call this config only in one method. But now I see we call it several times.

michalsn (approved):

PR #1263, opened 2025-03-31
refactor: Ensure `findById` returns `null` instead of an array when ID is null

datamweb:
**Description**
Although the phpdocs states that the ID ($id) should not be `null`, in some cases, data may be retrieved from the database where the ID is set to `null`.

When `null` is passed to the `find($id)` method, instead of returning null or a User object, it returns an empty array ([]).

This leads to a type mismatch with the method signature, which expects a return type of `?User`, causing the error:

CodeIgniter\Shield\Models\UserModel::findById(): Return value must be of type ?CodeIgniter\Shield\Entities\User, array returned 

This change ensures that the method always adheres to the expected return type (?User), preventing unexpected runtime errors.

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/src/Models/UserModel.php
+++ b/src/Models/UserModel.php
@@ -182,7 +182,9 @@ class UserModel extends BaseModel
      */
     public function findById($id): ?User
     {
-        return $this->find($id);
+        $result = $this->find($id);
+        
+        return $result instanceof User ? $result : null;
     }
 
     /**


michalsn (commented):
Can we add a test for this?

PR #1283, opened 2025-08-07
fix: Bug in fetchIdentities when using first()

b1tc0re:
**Description**
This PR fixes a critical bug when using `withIdentities()->first()` and improves test reliability.

 **Fixed identity loading bug:**
   - Added missing ID population for `first()` method in `fetchIdentities`
   - Resolved "undefined array key id" error when calling `withIdentities()->first()`
   - Now correctly handles identity relationships for single results

**Checklist:**
- [ ] Securely signed commits
- [X] Component(s) with PHPDoc blocks, only if necessary or adds value
- [X] Unit testing, with >80% coverage
- [ ] User guide updated
- [X] Conforms to style guide

### DIFF
--- a/src/Models/UserModel.php
+++ b/src/Models/UserModel.php
@@ -142,6 +142,10 @@ class UserModel extends BaseModel
 
         $mappedUsers = $this->assignIdentities($data, $identities);
 
+        if ($data['method'] === 'first') {
+            $data['id'] = $data['data']->id;
+        }
+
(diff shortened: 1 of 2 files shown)


michalsn:
Thanks! However, there's no bug per se - methods like `withIdentities()`, `withGroups()`, and `withPermissions()` are intended to be used with the `find*` methods.

That said, we're always open to improvements. If you're interested, feel free to update all the relevant methods and documentation accordingly.

PR #1286, opened 2025-10-01
fix: check for empty actionClass so that actions can be disabled via .env params

najdanovicivan:
**Description**

With current setup setting login or register action to empty string will trigger error.  Empty values should be allowed so that it is possible to disable the action from env file.

Example in env to disable TOTP in development environment

auth.actions.login = ''

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/src/Authentication/Authenticators/Session.php
+++ b/src/Authentication/Authenticators/Session.php
@@ -195,7 +195,7 @@ class Session implements AuthenticatorInterface
     {
         $actionClass = setting('Auth.actions')[$type] ?? null;
 
-        if ($actionClass === null) {
+        if ($actionClass === null || $actionClass === '') {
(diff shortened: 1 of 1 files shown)


michalsn:
I believe we're stepping back from using `empty`, as it's too loose.

You can replace it with `$actionClass === null || $actionClass === ''`

najdanovicivan:
@michalsn  I've updated the code to avoid using empty() function. Can we merge this one now ?

michalsn (approved):
Thank you. Rector errors are not relevant to the PR.

datamweb (approved):
LGTM, Thanks @najdanovicivan ,@michalsn .

PR #1257, opened 2025-02-28
feat: Add user relations loading methods for groups and permissions

michalsn:
**Description**
This PR enhances user queries by enabling the retrieval of related user data more efficiently.

It adds two new methods: `withGroups()` and `withPermissions()`, which loads user groups and permissions directly when fetching users. The existing `withIdentities()` method (previously undocumented) has also been included in the documentation.

With these methods, all necessary user information can be loaded in a single query (one per used method), reducing the number of additional queries and improving performance.

The problem this PR solves was described in the Bonfire2 repo: https://github.com/lonnieezell/Bonfire2/issues/558

**Checklist:**
- [x] Securely signed commits
- [x] Component(s) with PHPDoc blocks, only if necessary or adds value
- [x] Unit testing, with >80% coverage
- [x] User guide updated
- [x] Conforms to style guide

PR #1290, opened 2025-10-05
fix: `withPermissions()` for users without permissions

michalsn:
**Description**
This PR fixes a bug affecting `UserModel::withPermissions()`, which caused users without permissions to trigger additional, unnecessary database queries. The same behavior was implemented for the `UserModel::withGroups()` for unification.

Bug reported here: https://github.com/codeigniter4/shield/pull/1257#issuecomment-3368990849

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


dgvirtual:
I can confirm that upon implementing these changes in the code Bonfire2 user list does not make lazy loading database queries for permissions. Query number got down from 25 to 10.

datamweb (approved):
@michalsn Thank you!

Issue #1293, opened 2025-10-30
Dev: Add bot detection for magic links


lonnieezell:
In one of my projects we ran into issues with Android mobile devices and SMS where the OS was trying to be too helpful and would visit the links to ensure they were safe, or prep for read aloud, etc, and it would invalidate the magic link since it had already been visited. 

While this was specifically for SMS, I can see email clients attempting to be helpful breaking the magic link system, also. Additionally, there is potential use by developers using that link for SMS messages. 

There are 2 solutions and we might want to consider both long-term. 

1) Analyze the user-agents and/or IP address of known bots that might interfere and block their attempt at that URL.
2) A better long-term alternative is to make it something that requires user input, like sending 4-6 numbers in the email and require user input so we don't have to play cat and mouse with bots. 

Here's the list of user agent strings that we've used and is working so far: 

- Googlebot
- Bingbot
- Slurp
- DuckDuckBot
- Baiduspider
- Sogou
- Exabot
- ia_archiver
- Google-PageRenderer
- Google-Read-Aloud
- Google-Safety
- bot
- crawler
- spider
- Yandex
- BingPreview

michalsn:
I believe someone mentioned a similar issue on Slack some time ago. I prefer option 2, as it appears to be the most widely adopted. We would still use the magic link, but require a 6-digit code for verification.

We could even automatically redirect users to the magic link and wait for them to enter the code - similar to how 2FA typically works. This approach would work for both email and SMS messages.

I'm not sure how many changes this would require, but it would likely "break" the current magic link implementation, since the logic would need to be updated and a new view added.

That said, option 1 seems less invasive, as we could provide a config to specify which user agents should be blocked.

lonnieezell:
I agree with those sentiments. I think maybe option 1 is a good short term fix, and then we either create a new path for option 2. or deprecate the existing magic links and replace it in the next major version?

michalsn:
Sounds good to me. Though I wonder - wouldn't it make sense to just update the `UserAgents::robots` array and use that to block access? That way, it wouldn't be configurable at the Shield level, but rather handled within CodeIgniter itself.

lonnieezell:
Yes, that sounds perfect.

datamweb:
> 2\. A better long-term alternative is to make it something that requires user input, like sending 4-6 numbers in the email and require user input so we don't have to play cat and mouse with bots.

The solution I prefer is the one I’ve previously implemented for activation links using SMS in several projects.

PR #1294, opened 2025-10-31
feat: ignore robots when accessing magic link

michalsn:
**Description**
This PR updates the magic link controller to ignore requests originating from bots or crawlers.

See #1293 for more details.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/docs/references/magic_link_login.md
+++ b/docs/references/magic_link_login.md
@@ -23,6 +23,10 @@ in the **app/Config/Auth.php** file.
 public int $magicLinkLifetime = HOUR;
 ```
 
+### Bot Detection
+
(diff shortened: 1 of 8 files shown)


datamweb (commented):
@michalsn, thank you. I know the main focus of the discussion has been on the magic link, but I remember someone mentioning a similar issue regarding the activation link as well.
I’d like to know what your plan is regarding that issue?

lonnieezell (approved):
Looks great, thanks!

datamweb:
Since the list of bots was recently updated in CodeIgniter version 4.7, but Shield remains compatible with earlier versions, I think it would be helpful to mention in Shield’s documentation that the bot list can be updated through the **app/Config/UserAgents.php** file.

michalsn:
@datamweb I have added the docs.

We can add the same functionality to all actions in a separate PR. I think the right place for that would be https://github.com/codeigniter4/shield/blob/develop/src/Controllers/ActionController.php#L83, right?

datamweb:
> https://github.com/codeigniter4/shield/blob/develop/src/Controllers/ActionController.php#L83, right?

Yes.

datamweb (approved):
LGTM, Thanks.

michalsn:
Thank you for your review!

PR #1300, opened 2025-12-05
feat: add Magic Login Modes (link or code with multiple formats)

### LINKED ISSUE #1293: Dev: Add bot detection for magic links

datamweb:
**Description**
This update makes the Magic-Login system configurable **without breaking backward** compatibility. Now developers can choose between:

1. A activation link(**clickable**)
2. A verification code — in **6-numeric**, **6-alpha**, **6-alnum** or **6-oneof** format.

This flexibility lets you adapt authentication to your app’s needs, compliance rules, or user-experience preferences.

For the code-based flow, a new dedicated view is introduced to keep things clean and avoid conflicts with existing or customized views, ensuring backwards compatibility.

close : #1293 
see : #1261
**Checklist:**
- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide


sanchawebo (commented):
One minor thing: You have multiple calls to `resolveMode()` in the `loginAction()` method, which generates a new token everytime. This could lead to confusion about its usage in the future.
Otherwise looks good to me and would fix my issue #1261

sanchawebo (approved):

datamweb:
@sanchawebo Thank you for your helpful review!

lonnieezell (approved):
What a simple, yet powerful solution. I like it. Looks great. Thanks!

michalsn (commented):
A 20-character alphanumeric token is safe to use as a standalone login identifier, because it has high entropy. The chance of guessing any user's token, even with many users in the system, is close to zero.

In contrast, a 6-digit or 6-character code is far too weak to use as the only credential that grants access. Even if it expires quickly, it is still unsafe if it is accepted without any additional user identifier. As the number of users grows, the system becomes more vulnerable because many short codes are active at once, which increases the probability that an attacker's random guess will match someones code.

The short lifetime does not fix the fundamental problem: if a code alone identifies and authenticates the user, then anyone who guesses an active code gains access to whichever account it belongs to.

IMO, to use short codes safely, we must ties them to a user identifier (user ID) and protect with rate limiting (Throttler).

PR #1322, opened 2026-04-20
feat: add `prefix` option for authentication routes

dimtrovich:
**Description**

This pull request adds the ability to prefix all authentication route paths using a new `prefix` option in the `routes()` method.

- **New `prefix` option**: Allows you to define a common prefix for all authentication routes (default `/`)
- **Refactoring**: Replaces the hard-coded route group `‘/’` with a configurable `$prefix` variable

#### Benefits of this change

- **Increased flexibility** - Allows you to group auth routes under a single prefix (e.g., `auth/login`, `auth/register`)  
- **Better organization** - Facilitates the organization of routes in complex applications  
- **Consistency** - Complements the existing `namespace` and `except` options  
- **Backward compatible** - The default value `/` preserves the existing behavior  

#### Usage example

```php
// All routes will be prefixed with ‘auth’
service(‘auth’)->routes($routes, [‘prefix’ => ‘auth’]);

// Generates: auth/login, auth/register, auth/logout, etc.
```

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/docs/customization/route_config.md
+++ b/docs/customization/route_config.md
@@ -31,10 +31,10 @@ This will generate the routes with the specified namespace instead of the defaul
 
 ## Change Prefix
 
-If you wish, you can prefix all defined authentication routes using the `prefix` option. This is particularly useful if you want all your routes to be under the same root (for example, `auth/login`, `auth/register`).
+If you wish, you can prefix all defined authentication routes using the `group` option. This is particularly useful if you want all your routes to be under the same root (for example, `auth/login`, `auth/register`).
 
 ```php
-service('auth')->routes($routes, ['prefix' => 'auth']);
+service('auth')->routes($routes, ['group' => 'auth']);
 ```
 
 This generates routes whose paths are all prefixed with `auth`.
--- a/docs/customization/route_config.md
+++ b/docs/customization/route_config.md
@@ -29,6 +29,16 @@ service('auth')->routes($routes, ['namespace' => '\App\Controllers\Auth']);
 
 This will generate the routes with the specified namespace instead of the default Shield namespace. This can be combined with any other options, like `except`.
 
+## Change Prefix
+
+If you wish, you can prefix all defined authentication routes using the `prefix` option. This is particularly useful if you want all your routes to be under the same root (for example, `auth/login`, `auth/register`).
+
+```php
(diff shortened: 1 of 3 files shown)


michalsn:
This makes sense. However, could we rename `prefix` to `group`? It would better align with the underlying method we use.

dimtrovich:
> This makes sense. However, could we rename `prefix` to `group`? It would better align with the underlying method we use.

done

michalsn:
Please rebase your branch onto `upstream/develop`. If you don't have an `upstream` remote configured, add it first and then run:
```
git rebase upstream/develop
```

michalsn:
It looks like your branch was updated by merging/adding commits from `develop`, so this PR now includes unrelated commits:

* `4f28d51` chore: fix code styles (#1324)
* `22bf063` chore: fix phpstan (#1325)
* `a1384fa` chore: remove deprecated rector class (#1326)

Those should not be part of this PR. Please rebase your branch on top of the latest `develop` instead, and sign your commits while rewriting them.

If you do not already have the main Shield repository added as `upstream`, add it first:

```bash
git remote add upstream https://github.com/codeigniter4/shield.git
git fetch upstream
```

Then rebase your branch:

```bash
git checkout feat/prefix-options-to-routes

# optional safety backup
git branch backup/feat-prefix-options-to-routes

git rebase -i --rebase-merges --exec "git commit --amend --no-edit -S" upstream/develop
```

In the interactive rebase, drop the unrelated commits listed above if they appear. Keep only the commits that belong to this PR.

Then force-push the cleaned branch:

```bash
git push --force-with-lease origin feat/prefix-options-to-routes
```

Please use rebase rather than merging `develop` into the PR branch, so the PR only contains your changes.

datamweb:
@dimtrovich Thanks for the PR! 

Previously, we encouraged users to handle this via standard route groups like so:
```php
$routes->group('accounts', static function($routes) {
    service('auth')->routes($routes);
});
```
However, this new `group` option is a great shorthand. 

Could you please add a quick reminder to the docs regarding filters? Something like:
> **Note:** If you have grouped or changed the default route format, ensure your paths match the new format in the `app/Config/Filters.php` file.

PR #1323, opened 2026-04-20
feat: add `only` option for authentication routes configuration

dimtrovich:
**Description**
This pull request adds the ability to specify only the authentication routes to include via a new `only` option in the `routes()` method.
In fact, if we only want the login, it would be easier to use ['only' => ['login']] than ['except' => ['register', 'logout', 'magic-link']]

#### Changes in Auth.php
- **New `only` option**: Allows specifying a list of routes to include (opposite of `except`)
- **Mutual validation**: Added a check to prevent the simultaneous use of `only` and `except`

#### Benefits of this change

- **Improved flexibility** - Complements the `except` option by offering a reverse approach (whitelist vs. blacklist)  
- **Intuitive API** - Allows for a more concise configuration when few routes are needed  

#### Usage examples

```php
// Include only the login and logout routes
service('auth')->routes($routes, ['only' => ['login', 'logout']]);

// Exclude magic-link routes (opposite behavior)
service('auth')->routes($routes, ['except' => ['magic-link']]);

// Error - these two options

**Checklist:**
- [ ] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### CODE REVIEW

**Comment by michalsn** [comment]
This looks useful, thanks.

**Comment by michalsn** [comment]
Please rebase your branch onto `upstream/develop`. If you don't have an `upstream` remote configured, add it first and then run:
```
git rebase upstream/develop
```

michalsn:
Please follow the similar path as described here: https://github.com/codeigniter4/shield/pull/1322#issuecomment-4405450038

dimtrovich:
> Please follow the similar path as described here: [#1322 (comment)](https://github.com/codeigniter4/shield/pull/1322#issuecomment-4405450038)

done

datamweb (approved):
LGTM! Thanks.

michalsn (approved):

datamweb:
Thanks!

PR #1329, opened 2026-05-30
fix: require login actions for magic links

memleakd:
**Description**

This fixes a gap in the magic-link login flow.

Password login already respects configured login auth actions, such as Email-based 2FA. Magic-link login, however, could finish immediately after the token was verified. So an app could have 2FA enabled for normal login, but magic-link login would not go through that same action flow.

With this PR, magic-link login now starts the configured login action when one applies to the user. If an action is required, the user is sent to the normal auth-action screen first, and the magic-link login is only treated as complete after that action passes.

The existing `magicLogin` session value and `magicLogin` event still work. They are just delayed until the login action is completed, so they continue to mean "the user has finished logging in with a magic link".

I also added a short note to the docs, plus tests for normal and conditional login actions.

**Checklist:**

- [x] Securely signed commits
- [x] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/docs/references/magic_link_login.md
+++ b/docs/references/magic_link_login.md
@@ -35,6 +35,8 @@ Some apps or devices may try to be "too helpful" by automatically visiting links
 
 Magic Link logins allow a user that has forgotten their password to have an email sent with a unique, one-time login link. Once they've logged in you can decide how to respond. In some cases, you might want to redirect them to a special page where they must choose a new password. In other cases, you might simply want to display a one-time message prompting them to go to their account page and choose a new password.
 
+If a login auth action is configured, such as Email-based Two Factor Authentication, the user must complete that action before the magic link login is finished.
+
 ### Session Notification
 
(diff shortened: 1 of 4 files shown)


michalsn (dismissed):
The fix makes sense. I only wonder... if we add another passwordless login method later, it may need the same handling. A generic pending login-method marker would keep this reusable... but maybe we should keep things simple and refactor it later if it will be needed.

michalsn:
@memleakd Please rebase and fix PHPUnit issues.

memleakd:
> @memleakd Please rebase and fix PHPUnit issues.

Done

memleakd:
> I only wonder... if we add another passwordless login method later, it may need the same handling. A generic pending login-method marker would keep this reusable...

Good point, thanks for calling that out. I refactored it to use a generic marker instead of magic-link-specific.

datamweb (approved):
LGTM! Thanks!

michalsn (approved):

PR #1331, opened 2026-05-30
docs: explain gateway auth action pattern

memleakd:
**Description**

This follows up on #1328 and @datamweb's idea about using a conditional action as a small gateway for more complex login flows.

I tried the pattern locally, and it works: an app can register one login action, let `appliesTo()` decide whether it should run for the user, and then handle the actual method choice inside the action's normal `show()` / `handle()` / `verify()` flow.

This PR adds that pattern to the Authentication Actions docs.

The important detail is that the gateway should still behave like one Shield action. It should return one type from `getType()` and use that as the pending action identity. The available methods, like email or SMS, should be handled inside the gateway instead of becoming separate action identities.

This documents a useful app-level pattern for people who need one login action to offer more than one verification method.

**Checklist:**

- [x] Securely signed commits
- [x] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/docs/references/authentication/auth_actions.md
+++ b/docs/references/authentication/auth_actions.md
@@ -135,3 +135,55 @@ the condition cannot be determined. It is not a replacement for authorization.
 
 Once an action is already pending in the session, Shield continues that pending
 action instead of rechecking the condition.
+
+## Gateway Actions
+
+Shield allows one configured action for each authentication event, such as
+`login` or `register`. If your application needs to choose between multiple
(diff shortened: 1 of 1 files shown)


michalsn:
Please rebase.

memleakd:
> Please rebase.

Done

datamweb (approved):
LGTM! Thanks!

michalsn (approved):

datamweb:
@memleakd @michalsn thanks!

PR #1330, opened 2026-05-30
chore(phpstan): use unsealed array shapes for credentials

datamweb:
**Description**
This PR fixes PHPStan errors related to extra keys being passed to the `$credentials` array in Authenticators.

```console
Error: Parameter #1 $credentials of method CodeIgniter\Shield\Authentication\Authenticators\Session::attempt() expects array{email?: string, username?: string, password?: string}, non-empty-array given.
Error: Parameter #1 $credentials of method CodeIgniter\Shield\Authentication\Authenticators\Session::attempt() expects array{email?: string, username?: string, password?: string}, array{status: 'abcde', password: 'secret123'} given.
Error: Parameter #1 $credentials of method CodeIgniter\Shield\Authentication\Authenticators\Session::attempt() expects array{email?: string, username?: string, password?: string}, array{status: '12345', password: 'secret123'} given.
 ------ ---------------------------------------------------------------------- 
  Line   src/Controllers/LoginController.php                                   
 ------ ---------------------------------------------------------------------- 
  71     Parameter #1 $credentials of method                                   
         CodeIgniter\Shield\Authentication\Authenticators\Session::attempt()   
         expects array{email?: string, username?: string, password?: string},  
         non-empty-array given.                                                
         🪪  argument.type                                                     
         💡  Sealed array shape can only accept a constant array. Extra keys   
         are not allowed.                                                      
 ------ ---------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------------- 
  Line   tests/Authentication/Authenticators/SessionAuthenticatorTest.php       
 ------ ----------------------------------------------------------------------- 
  478    Parameter #1 $credentials of method                                    
         CodeIgniter\Shield\Authentication\Authenticators\Session::attempt()    
         expects array{email?: string, username?: string, password?: string},   
         array{status: 'abcde', password: 'secret123'} given.                   
         🪪  argument.type                                                      
         💡  Sealed array shape does not accept array with extra key 'status'.  
  482    Parameter #1 $credentials of method                                    
         CodeIgniter\Shield\Authentication\Authenticators\Session::attempt()    
         expects array{email?: string, username?: string, password?: string},   
         array{status: '12345', password: 'secret123'} given.                   
         🪪  argument.type                                                      
         💡  Sealed array shape does not accept array with extra key 'status'.  
 ------ ----------------------------------------------------------------------- 

Error:  [ERROR] Found 3 errors                                                         
```

[...]

PR #1332, opened 2026-05-30
chore(phpstan): restrict unsealed credential keys to string keys

datamweb:
**Description**
This PR follows up on the previous change(PR #1330 ) that introduced PHPStan unsealed array shapes for authenticator credentials.

In the earlier PR, bare `...` was used in array shapes. However, in PHPStan this expands to `...<array-key, mixed>`, which allows both `string` and `int` keys. That is broader than intended for Shield credentials, because custom login fields are expected to be `string-keyed` fields.

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide

### DIFF
--- a/phpstan-baseline.php
+++ b/phpstan-baseline.php
@@ -68,13 +68,13 @@ $ignoreErrors[] = [
 	'path' => __DIR__ . '/src/Authentication/Authenticators/JWT.php',
 ];
 $ignoreErrors[] = [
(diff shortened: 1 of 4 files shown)


paulbalandan (approved):

michalsn (approved):

michalsn:
Thank you @datamweb

PR #1335, opened 2026-06-04
chore: update action pinning hash commit in github workflows

ddevsr:
**Description**
This PR pins GitHub Actions from mutable version tags (e.g. `action/checkout@v5`) to full commit SHAs, preventing silent supply chain attacks from compromised action repositories.

Recommended by [GitHub's security hardening guide](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions)

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -10,11 +10,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10
         with:
           fetch-depth: 0
       - name: Setup Python
-        uses: actions/setup-python@v6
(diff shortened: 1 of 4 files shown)


.github/workflows/docs.yml
```diff
@@ -10,11 +10,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10
```
paulbalandan: It is recommended to add as comment the equivalent version tags, since dependabot can update those as well. See the main repo for examples.

paulbalandan (approved):

samsonasik:
Thank you @ddevsr

PR #1336, opened 2026-06-04
chore: migrate from deptrac to structarmed

samsonasik:
**Description**

This PR try to migrate deptrac to structarmed, the vendor parts seems no need to be included in layer definition, structarmed treats dependencies on classes that don't match any defined layer as external/unregistered and automatically allows them, and `src/Views/` contains only PHP template files with no class or namespace declarations, so no PHP class would ever match a View layer pattern regardless

Let me know if there is a catch so I can improve in structarmed first itself if possible. 

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [x] Conforms to style guide


samsonasik:
Ready for review/merge 👍

.github/workflows/structarmed.yml
```diff
@@ -0,0 +1,78 @@
...
+
+jobs:
+  build:
+    name: Dependency Tracing
+    runs-on: ubuntu-latest
+    if: (! contains(github.event.head_commit.message, '[ci skip]'))
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
```
datamweb:
```suggestion
        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3 
```

.github/workflows/structarmed.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Set up PHP
+        uses: shivammathur/setup-php@v2
```
datamweb:
```suggestion
        uses: shivammathur/setup-php@7c071dfe9dc99bdf297fa79cb49ea005b9fcadbc # v2.37.1
```

.github/workflows/structarmed.yml
```diff
@@ -0,0 +1,78 @@
...
+          extensions: intl, json, mbstring, xml
+          coverage: none
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Get composer cache directory
+        run: echo "COMPOSER_CACHE_FILES_DIR=$(composer config cache-files-dir)" >> $GITHUB_ENV
+
+      - name: Cache composer dependencies
+        uses: actions/cache@v4
```
datamweb:
```suggestion
        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae #v5.0.5
```

samsonasik:
@datamweb review applied 👍

datamweb:
> structarmed treats dependencies on classes that don't match any defined layer as external/unregistered and automatically allows them

Since structarmed allows all unregistered classes by default, it seems the only way to forbid vendor usage in a specific layer is to explicitly register those vendor namespaces via `layerPattern()`, and then omit them from the ruleset.
Is registering vendors as layers the recommended approach in structarmed to achieve strict boundaries? Or do you plan to add a feature like a strict mode (deny unregistered by default for specific layers)?

datamweb (approved):
LGTM!

samsonasik:
@datamweb I think I need a real violation example first for that, eg use of:

```php
new \CodeIgniter\Controller();
```

on specific part, which not allowed. That will may use `MayNotDependOnRule`, probably something like:

```php
    ->layer('CI_Vendor_Controller', ....)
    ->rule(
        'entity.must_not_use_ci_controller',
        new MayNotDependOnRule(from: 'Model', to: 'CI_Vendor_Controller')
    )
```

michalsn (approved):

samsonasik:
Thank you @datamweb @michalsn , let's merge 👍

PR #1341, opened 2026-06-15
test: fix FirebaseAdapter encode test for unsupported algorithm

michalsn:
**Description**
This PR fixes the `FirebaseAdapter` encode test, which broke because `firebase/php-jwt` now recognizes PS256 as supported (it just needs phpseclib).

**Checklist:**
- [x] Securely signed commits
- [ ] User guide updated
- [ ] Conforms to style guide


datamweb (approved):
Thanks!

